PR #13, opened 2023-04-07
New Player Design and Listview

dobby5:
Added reworked Player (New design, Fullscreen mode, More functions, r…
Improved Listview
Sync localization
Improvements and fixes


marcelveldt (approved):
As discussed on PM the linting will be fixed in follow-up PR(s).
Nice work, its already looking very good !

PR #20, opened 2023-04-26
Redesign 2.1

dobby5:
Extensive revision of the display processes (ItemsListings, ListviewItem and PanelviewItem) and of MediaItemThumb as well as PlayerTrackDetails:

- Complete revision of the styles
- Improvement of the stability of the items
- Addition of a new (visual) Selection (Panel)
- Addition (Panel) and completion (List) of rating symbols
- Significant improvement of responsibility of cover images
- Modification of the content type button to improve the dark/light mode and hover actions

Relocation and renewal of the AppBar

Added experimental Button Icon component
Improved linter
Restoration of old cover art display in fullscreen until new error is fixed.
Many design improvements or adjustments for future appearances.


marcelveldt:
CI is failing on the linting, can we fix the issues its complaining about ?

marcelveldt:
can you please try to create smaller PR's with only a single change or a few related changes combined ? Now the changes are everywhere and I'm having a hard time reviewing it, especially because there seem to be side effects. 

I have extensively tested the PR and I have some feedback, I'll share in separate comments.

marcelveldt:
1) The header seems to be a bit broken. The breadcrumb navigation is now missing and the contextmenu dissapeared too.
Also, I'm not so sure if I'm a fan of the very large header font and space. It seems a bit out of place but that's maybe just me.

marcelveldt:
2) I'd drop the resizable menubar. It works a bit wonky and feels sluggish. Also it does not add much so for now I'd say drop it as it does not bring anything

.editorconfig
```diff
@@ -1,5 +1,12 @@
+root = true
```
marcelveldt, line 1: can we leave out all the linting and editor changes ? It makes the PR so hard to follow. I'm trying to debug the issues I found but its hard because you cant exactly pinpoint what changed due to the large amount of unrelated changes due to the linter preferences.

My proposal would be to create one PR for each functional change (or combine some small ones ofc) and one or more PR's for linter changes only otherwise. Also some feedback to why the linter changes are needed are also welcome as I see there are now some rules disabled which were previously enabled and we're further drifting apart from the baseline vue3/vuetify + vite config. My personal preference is to follow defaults as much as we can that comes with the language or tool.

dobby5:
> 2. I'd drop the resizable menubar. It works a bit wonky and feels sluggish. Also it does not add much so for now I'd say drop it as it does not bring anything

As discussed, the removal will follow.

dobby5:
> 1. breadcrumb

The context menu has been fixed.

This is not an error but the implementation according to the concept. The concept does not include a breadcrumb, and the text has been adapted according to the concept. If you want to adjust the font, you can easily do so in global.css.
Please note that these are "global" changes.

.editorconfig
dobby5: We had already talked about the linter several times and how useful it is, at that time I introduced it into the project.
The configuration is based on today's standards and recommendations. Not many rules were deactivated, the plug-ins add their own rules and not all of them are desired. There are more rules, which is not in proportion to the deactivations.

An undo at this point is not possible given the number of changes. Tell me what you want to quake and I'll do it. In addition, my code will not be overwritten and the work will not be disturbed. After all, you have given me the main responsibility for the frontend, so I can fix the errors or make the desired changes more quickly and cleanly.

marcelveldt:
> > 1. breadcrumb
> 
> The context menu has been fixed.
> 
> This is not an error but the implementation according to the concept. The concept does not include a breadcrumb, and the text has been adapted according to the concept. If you want to adjust the font, you can easily do so in global.css. Please note that these are "global" changes.

Ah ok, I see.
In that case I have some sort of challenge. I used the breadcrumb as "back navigation" to jump a list back into position.
For instance go to Artists, scroll down a lot, click an artist to view details. Then if you do browser back, you get back into the artists listing but not at the scrolling position which is annoying as you need to scroll down again. So there was a bit of a workaround in the code to remember the position in the router and then when you navigate back, you scroll back to that position again where you left off.

If we can somehow reintroduce that behavior again, I don't care about the breadcrumb :-)

dobby5:
> > > 1. breadcrumb
> > 
> > 
> > The context menu has been fixed.
> > This is not an error but the implementation according to the concept. The concept does not include a breadcrumb, and the text has been adapted according to the concept. If you want to adjust the font, you can easily do so in global.css. Please note that these are "global" changes.
> 
> Ah ok, I see. In that case I have some sort of challenge. I used the breadcrumb as "back navigation" to jump a list back into position. For instance go to Artists, scroll down a lot, click an artist to view details. Then if you do browser back, you get back into the artists listing but not at the scrolling position which is annoying as you need to scroll down again. So there was a bit of a workaround in the code to remember the position in the router and then when you navigate back, you scroll back to that position again where you left off.
> 
> If we can somehow reintroduce that behavior again, I don't care about the breadcrumb :-)

I looked at your implementation, unfortunately it only works moderately, but this is logically explainable. Because as soon as you scroll down for a long time and the elements load, the scroll position is recorded.
If you click on an album and then go back, the scroll bar moves, but not to the desired position. The reason for this is that the elements that are loaded dynamically when scrolling are not yet present at this position, and therefore the scrollbar is not yet long enough to be able to scroll to the old position.
You have made the best of it.

I can implement the function you want and can also improve it, but the limits are linked to the loading time.
Perhaps it could be improved by first passing on the number of elements when each page is called up, thus reducing the size of the requested data. Then the information of the elements would be loaded one by one. However, the placeholders are already there and so the full length of the scrollbar is also available.

marcelveldt:
Yeah, it was by no means perfect and more ideally we would just want that behavior on browser back. If its complicated we can just add a ticket to the backlog for now and look at it later.

So, to get back to this PR.
Let's fix the conflicts asap and make sure that CI passes the lint/tests. As long as the result matches the new UX/UI design you indeed have sort of a green pass to work on the frontend ;-) But still I'd like to see smaller PR's just so we can learn from eachother and review. Some of the code is maybe not so good but exists of history reasons.

dobby5:
> Yeah, it was by no means perfect and more ideally we would just want that behavior on browser back. If its complicated we can just add a ticket to the backlog for now and look at it later.
> 
> So, to get back to this PR. Let's fix the conflicts asap and make sure that CI passes the lint/tests. As long as the result matches the new UX/UI design you indeed have sort of a green pass to work on the frontend ;-) But still I'd like to see smaller PR's just so we can learn from eachother and review. Some of the code is maybe not so good but exists of history reasons.

I will add a function that tries to show the same behaviour as you want and as just described. It will not take long.

The PRs will become smaller in the future.
Code is never perfect, even mine, efficiency and effectiveness...
The most important thing is to keep the balance. I think we are not on the worst path.

marcelveldt:
I fixed the merge conflicts but CI is failing on the lint changes as it seems.
Also, still seeing weird (css?) issues on my side while testing like the provider icons and listitems alignments which I showed you in PM. I'll have a look at that tomorrow.

marcelveldt:
I spotted a few leftover issues but lets address them in follow up PR(s)

marcelveldt (approved):
Nice!

PR #45, opened 2023-05-14
browsefolder display image if available

jozefKruszynski:
Added extra condition to display an image on for BrowseFolder instead of default icon


src/components/ListviewItem.vue
```diff
@@ -39,6 +39,15 @@
...
         </div>
+        <div
+          v-else-if="item.media_type == MediaType.FOLDER && item.metadata.images"
+          class="listitem-thumb"
+        >
+          <MediaItemThumb
+            :item="item"
+            :size="50"
+          />
+        </div>
```
marcelveldt:
```suggestion
```

redundant code

src/components/ListviewItem.vue
```diff
@@ -25,7 +25,7 @@
           />
         </div>
         <div
-          v-else-if="item.media_type == MediaType.FOLDER"
+          v-else-if="item.media_type == MediaType.FOLDER && !item.metadata.images"
```
marcelveldt: this change alone should be enough

marcelveldt (dismissed):
thanks!

PR #213, opened 2023-08-18
Add dark mode theme-color for top bar on mobile browsers (and Safari desktop)

krisnoble:
On the [most common mobile browsers](https://caniuse.com/?search=theme-color) it's possible to theme the top bar colour to match the content. Currently in MA the `theme-color` is set to `#ffffff` (white) at all times, which doesn't adapt to dark mode like the rest of the UI.

This PR fixes that by adding an extra `meta` tag and giving them both media queries, so the browser will choose the appropriate colour based on the OS theme preference. This is currently [standard practice](https://web.dev/add-manifest/#theme-color) as multiple theme colours are not yet supported in manifests.


marcelveldt (approved):
LGTM, thanks!

PR #300, opened 2023-12-15
Add heart icon to grid view

JeffersonBledsoe:
Screenshot 2023-12-15 at 12 36 49 am

## Todo:

- [ ] Testing with variuos other metadata about the item also visible (such as disc number, explicit album, etc.)
  - I only have Spotify setup so am unable to test this myself
- [ ] i18n for the aria-label
  - How do you create a new translation?
  - What label would we want for this, `Is favorite`?


OzGav:
This is album view so disc number not required. We don’t show explicit anywhere else as far as I know so adding that might require a wider assessment of the frontend. 

Do you mean adding a translation as a new language or do you mean a string to be translated?

OzGav:
@JeffersonBledsoe i think mark this as ready as it fixes an existing issue and other enhancements can be considered at a later date

JeffersonBledsoe:
@OzGav > Do you mean adding a translation as a new language or do you mean a string to be translated?

Adding a new string to be translated. We have "mark/ unmark favourite" but don't have one that is just "Is favourite". I haven't figured out how the i18n machinery works in vue yet

Aside from that, if you're happy this needs no further testing then I'm happy :)

OzGav:
Yes I think click the ready for review button and see what Marcel thinks.

marcelveldt:
Do we only want to show the heart icon or make it a button ?
In the list view the heart is actionable

OzGav:
Good question. That would result in a hollow heart against all entries but I think it would be ok? Would a cut and paste of the list view code work?

marcelveldt:
> Good question. That would result in a hollow heart against all entries but I think it would be ok? Would a cut and paste of the list view code work?

If we want to be consistent then yes.

OzGav:
@JeffersonBledsoe Can you make additional changes to this PR? This is the list view code I believe:
```
        <!-- favorite (heart) icon -->
        <div v-if="getBreakpointValue('bp3') && 'favorite' in item && showFavorite && !$vuetify.display.mobile">
          <Button
            v-bind="props"
            variant="list"
            :icon="item.favorite ? 'mdi-heart' : 'mdi-heart-outline'"
            :title="$t('tooltip.favorite')"
            @click="api.toggleFavorite(item)"
            @click.prevent
            @click.stop
          />
        </div>
```

JeffersonBledsoe:
I'll get this updated later this week! 
Personally I don't see much of a benefit of having it actionable and adds alot of noise to the page, but we can always discuss that later. Let's get it consistent for now and we can always make tweaks later 🙂

marcelveldt:
@JeffersonBledsoe what's the status on this one ?

JeffersonBledsoe:
@marcelveldt Sorry about the delay in this one. Screen recording of the new behaviour of having the heart icon appear on a grid item and having it as a useable button for favouriting/ unfavouriting the selected item. I made the button a component so we can reuse it elsewhere.

https://github.com/music-assistant/frontend/assets/30210785/d18f42dd-5fbf-4f25-8652-652d3b269e48

There's an accessibility issue with having the `onClick` for the item and then having a button inside of it (you can't have nested interactive components as screen readers don't announce nested controls), but that's already an issue with the list view too so is a separate issue.

marcelveldt:
Woohoo, nice work!

marcelveldt (approved):
looks good! Thanks!

marcelveldt:
Yeah I think we have some more issues with the frontend, next to even the accessibility issues. But we're slowly working through them. Thanks for the fix!

PR #347, opened 2024-02-01
Add wheel event to PlayerVolume component

githrdw:
A little quality of life improvement; I missed the UI pattern of wheeling over the volume slider to adjust the volume.
Dragging the little thumb is a bit cumbersome sometimes 😅

PR #372, opened 2024-02-14
Added wheel event to PlayerVolume (along with existing native events)

githrdw:
Fix for reverted https://github.com/music-assistant/frontend/pull/347

In my previous PR, the wheel event had to emit it's value using `ctx.emit`.
To satisfy TypeScript I added `'update:model-value'` to the component's emits, not knowing that would stop passing the update event to the slider itself (in `v-bind="playerVolumeProps"`).

So I manually added the update event now.


marcelveldt:
Thanks, another chance to test it!

marcelveldt (approved):

PR #373, opened 2024-02-15
Remove duplicate [default]

OzGav:
In the frontend we currently see [default][default] against a number of options


marcelveldt (approved):
Thanks!

PR #374, opened 2024-02-15
Changes to when and how the player row cards are shown

jozefKruszynski:


src/views/HomeView.vue
```diff
@@ -254,11 +255,22 @@ onMounted(async () => {
   });
 });
 
+const playerStateOrder = {
+  [PlayerState.PLAYING]: 1,
+  [PlayerState.PAUSED]: 2,
+  [PlayerState.IDLE]: 3,
+} as const;
+
```
marcelveldt: I thought we agreed on treating paused and idle the same ?

src/views/HomeView.vue
```diff
@@ -268,12 +280,22 @@ const updateCurrentlyPlayingQueueWidgetRow = function () {
           )
         ) {
           widgetRows.value.queue.players.push(player);
+          widgetRows.value.queue.players.sort((a, b) =>
+            a.name.localeCompare(b.name),
```
marcelveldt:
```suggestion
            a.display_name.localeCompare(b.display_name),
```

marcelveldt: always use display_name as that takes into account any custom name the user gave to a player

jozefKruszynski: Will do

marcelveldt (approved):
Nice!

PR #376, opened 2024-02-15
Fix the weird behaviour with the search field and badge

jozefKruszynski:


src/components/ItemsListing.vue
```diff
@@ -692,6 +693,20 @@ const onClick = function (mediaItem: MediaItemType) {
   }
 };
 
+const onClear = function () {
+  params.value.search = '';
+  showSearch.value = false;
+  loadData(true);
+};
+
+const isSearchActive = function () {
```
marcelveldt: don't you want to make this a computed property instead of a function ?

jozefKruszynski: You're absolutely right. I'll make a change later or in the morning

marcelveldt (approved):

PR #388, opened 2024-02-24
Reworked Home View Page

dobby5:
Adaptation to the style regulations
Improvement of the HomeView Page


package.json
marcelveldt, line 1: please try to not update packages in a feature PR, we do package bumps in seperate PR's unless a specific new package is needed for a new feature.

marcelveldt:
I wanted to try out this PR but it completely locks up my browser, the behavior we've seen before with the infinite loading update. The new homepage loads but all other sections, such as the artists etc lock up completely.

I see you've updated the packages in this same PR, including the "inifinite-loading" package which is known to cause issues. I suggest to remove the unrelated package bumps and leave that to dependabot.

Generic feedback is that the changes to the homepages are an improvement but not sure if the thumbsize is maybe a bit too large. Also the title headings are a bit misaligned.

dobby5:
> I wanted to try out this PR but it completely locks up my browser, the behavior we've seen before with the infinite loading update. The new homepage loads but all other sections, such as the artists etc lock up completely.
> 
> I see you've updated the packages in this same PR, including the "inifinite-loading" package which is known to cause issues. I suggest to remove the unrelated package bumps and leave that to dependabot.
> 
> Generic feedback is that the changes to the homepages are an improvement but not sure if the thumbsize is maybe a bit too large. Also the title headings are a bit misaligned.

The headings were not touched during editing.
The thumbnail size is the regular size and can be adjusted in the "Breakpoint" plugin under "breakpoints" for all.
The package has been reset and should also fix an error for you.

marcelveldt:
Yes, now its good and no longer crashes. We should really look into that "v3-infinite-loading" package why it induces the crashing, its a bit scary.

marcelveldt (approved):
nice!

PR #402, opened 2024-02-27
WIP: Automated testing

JeffersonBledsoe:
Starting this draft PR so we can start a conversation incase anybody has opinions. This doesn't really do much right now other than pre-configure the frontend to specify the MA server to be at `localst:8095`, visit `localhost:8095` and check the pages SPA wrapper has been rendered.

## Todo

- [ ] Setup a backend server to connect and talk to (docker?)
- [ ] Mock data in the backend (how?)
- [ ] Write a test to use the homepage


marcelveldt:
I have no experience with automated testing of a frontend but I wouldn't run a full backend server.
Instead we should be able to start a mock server with some predefined output so that testing becomes easier. 

Nevertheless great that you started working on this!

JeffersonBledsoe:
> I have no experience with automated testing of a frontend but I wouldn't run a full backend server.
Instead we should be able to start a mock server with some predefined output so that testing becomes easier.

Great! We can use [mock service worker](https://mswjs.io/) to mock network responses. I was starting down the 'real server' route mostly due to unfamiliarity with the backend. The largest issue is just keeping the mocks up-to-date with how the server actually responds, but we can work through that problem as we move forward. Expect questions while I get this going!

marcelveldt:
Yeah, what I meant is that we should adjust the server so it can be run in mock mode so it doesn't make any network calls but can serve data as it would normally. This is the idea of a bigger plan to get testing on the rails but that will have to wait until we released 2.0 stable. So the endgoal would be to have a full test framework for the server, python client and vue frontend.

stvncode:
Hey ! Just wanted to say hello and to see if there is a following on this ? I wanted also to have a debate on cypress/playwright since I think, for MA, playwright have more sense. We can debate on this and maybe close this one for now and open a new one when the scope of this will be clear ?

PR #462, opened 2024-04-19
Feature: search for radio stream title from full screen player

sprocket-9:
This adds the feature to click the radio stream title in the full screen player and perform a global search for it.
 
I think this will be useful to have despite the varying quality and junk sometimes appearing in the stream title metadata!  Despite that, search makes a pretty good job and user can modify the search manually if needed.

Routing to /search when it's already the current route does nothing, so I used the eventbus to notify the seach component - couldn't think of a better way.

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
+++ b/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
@@ -121,7 +121,18 @@
                 store.curQueueItem?.streamdetails?.stream_title
               "
             >
-              <h4 class="fullscreen-track-info-subtitle">
+              <!-- radio live metadata -->
+              <h4
+                v-if="
+                  store.curQueueItem.streamdetails.stream_title &&
+                  store.curQueueItem.streamdetails.stream_title.includes(' - ')
+                "
+                class="fullscreen-track-info-subtitle"
+                style="cursor: pointer"
+                @click="
+                  radioTitleClick(store.curQueueItem.streamdetails.stream_title)
+                "
+              >
                 {{ store.curQueueItem.streamdetails.stream_title }}
               </h4>
             </div>
@@ -486,6 +497,13 @@ const radioNameClick = function (item: Radio | ItemMapping) {
       provider: item.provider,
     },
   });
+  store.showFullscreenPlayer = false;
+};
+
+const radioTitleClick = function (streamTitle: string) {
+  // radio station title clicked
+  store.globalSearchTerm = streamTitle;
+  router.push({ name: 'search' });
   store.showFullscreenPlayer = false;
 };
 
--- a/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
+++ b/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
@@ -124,7 +124,6 @@
               <!-- radio live metadata -->
               <h4
                 v-if="
-                  store.curQueueItem.streamdetails.stream_title &&
                   store.curQueueItem.streamdetails.stream_title.includes(' - ')
                 "
                 class="fullscreen-track-info-subtitle"
@@ -133,6 +132,9 @@
                   radioTitleClick(store.curQueueItem.streamdetails.stream_title)
                 "
               >
+                {{ store.curQueueItem.streamdetails.stream_title }}
+              </h4>
+              <h4 v-else class="fullscreen-track-info-subtitle">
                 {{ store.curQueueItem.streamdetails.stream_title }}
               </h4>
             </div>


sprocket-9:
I'm assuming we're protected from searching that oddly formatted stream metadata coming from littlebobbytables FM, right? 😅 https://xkcd.com/327/

marcelveldt:
I'm not sure this is good;

a) there is no common format for streamtitle, radio stations just mess around.
b) like you already discovered some radio stations have jiberish in this stream title, some even put xml or json in there.
c) I'm not sure of the search is well protected

Overall the outcome will not be predictable and cause confusion. 
Maybe we can do a safeguard and only start search if there's a hyphen in the title ?
Its still no guarantee as some radio stations do TITLE -ARTIST and some others do ARTIST-TITLE and some others do just the title or a variation but at least when there's an hyphen we know almost certain its possibly a artist/title combination.
Still its hacky because some radio stations just have the show title in their or a commercial or whatever.

Filtering out the jibberish items is something to add to the backend, there is already an issue report open for that:
https://github.com/music-assistant/hass-music-assistant/issues/2136

marcelveldt:
FYI, this is where the server parses this streamtitle from the icy chunk:
https://github.com/music-assistant/server/blob/main/music_assistant/server/helpers/audio.py#L538

and some info about ICY metadata:
https://cast.readme.io/docs/icy

sprocket-9:
I know, you're correct, the search results will be unpredictable.  I totally get the desire to present a nice clean uniform interface like the rest of MA - the insistence on properly tagged files etc is a good design decision to achieve this.  

But MA is not in control of the metadata on these radio streams and trying to massage the data to present uniformity is going to be very difficult given the abuse of fields out there.  Not saying it's useless to try for some of the more structured examples, but for every title the parsing rules fix, there will be another they will butcher, ending up in bug reports and a game of regex whack-a-mole.

There may have to be an acceptance on the user's part that radio stream titles could be very messy in the player.  Post 2.0, could make showing the radio stream title a config opt-in setting with a ? Help explanation about the reasons why and a section in the docs.

For this PR's feature (assuming you want to have this feature in without the config setting above ) there are a couple of other options available to make it less surprising in the UI.  Remove the clickable title and either:
1. Add a magnifying glass search button next to the title
2. Add a button to copy title to clipboard
3. Change the CSS rules to make the title manually-selectable

> c) I'm not sure of the search is well protected
Which ever option it is, this needs confirmed of course.

Now I have radio browser available I'm discovering quite a lot of new songs on stations that I want to add to my library, so having an easy method of searching for it within MA would be nice!  Up until now I've been resorting to shazam->spotify.

sprocket-9:
I had a quick look at the station in https://github.com/music-assistant/hass-music-assistant/issues/2136 looks like it's one of these hls stations - didn't you say somewhere there's a new method for getting metadata from these type of streams?

Looks like that station is blocking certain user-agent strings - got a 403 forbidden in MA when trying to retrieve the playlist:
wget --user-agent="Python/3.11 aiohttp/3.9.3" https://wz2liw.scahw.com.au/live/22000s_128.stream/playlist.m3u8 # 403 Forbidden using MA's UA string
wget https://wz2liw.scahw.com.au/live/22000s_128.stream/playlist.m3u8 # 200 using default wget UA

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
```diff
@@ -121,7 +121,15 @@
                 store.curQueueItem?.streamdetails?.stream_title
               "
             >
-              <h4 class="fullscreen-track-info-subtitle">
+              <!-- radio live metadata -->
+              <h4
+                v-if="store.curQueueItem.streamdetails.stream_title"
```
marcelveldt:
```suggestion
                v-if="store.curQueueItem.streamdetails.stream_title && store.curQueueItem.streamdetails.stream_title.includes(' - ')"
```

marcelveldt: For now just do a check if the hyphen is present as that is in 90% of the cases true when artist+title is in streamtitle. Not waterproof but at least as bit of guard .

src/views/Search.vue
```diff
@@ -152,6 +152,9 @@ const filteredItems = function (itemType: string) {
 };
 
 onMounted(() => {
+  eventbus.on('globalsearchterm', (evt: GlobalSearchTermEvent) => {
```
marcelveldt: I'm thinking it would be better to have the searchterm as path property - Let's see if we can fix that

marcelveldt:
> I had a quick look at the station in [music-assistant/hass-music-assistant#2136](https://github.com/music-assistant/hass-music-assistant/issues/2136) looks like it's one of these hls stations - didn't you say somewhere there's a new method for getting metadata from these type of streams?
> 
> Looks like that station is blocking certain user-agent strings - got a 403 forbidden in MA when trying to retrieve the playlist: wget --user-agent="Python/3.11 aiohttp/3.9.3" https://wz2liw.scahw.com.au/live/22000s_128.stream/playlist.m3u8 # 403 Forbidden using MA's UA string wget https://wz2liw.scahw.com.au/live/22000s_128.stream/playlist.m3u8 # 200 using default wget UA

I don't think so. HLS streams have no ICY metadata but their own way of setting metadata (not really supported by many providers).

User agent of MA when streaming is not aiohttp but ffmpeg useragent.

src/views/Search.vue
sprocket-9: You mean you want a route to handle /search?searchterm="blah"?

Can you router.push to this route twice? Thinking of the scenario if you click stream title to search, then mistakenly press the mini player to go full screen, then press the stream title again to search - will this route?

sprocket-9:
> User agent of MA when streaming is not aiohttp but ffmpeg useragent.

The 403 is returned when the playlist is retrieved:
https://github.com/music-assistant/server/blob/main/music_assistant/server/helpers/playlists.py#L134

Anyway just something I noticed - it's not causing an issue with playing the stream or getting the metadata.  The url in the playlist seems to be the same as the stream url but with a unique identifier on the end, for tracking I guess.

src/views/Search.vue
marcelveldt: yeah, I'm not sure about that either tbh :-)
I'll have to do some tests. Its kind of cool that you can have the searchterm in the url as path query param

sprocket-9: Don't waste any time on this -  I'll have a play around with the router see what I can come up with and get back to you if I run into a wall.

sprocket-9: I've gave this a bit more thought and I can't really think of any benefits of using a path query param in a SPA.

Currently performing a manual search does not alter the URL to contain the search term.
If this is implemented for searching stream_title, the URL will have the original search term, but performing further searches will not update the URL, which is a bit confusing for the user.
On my phone, in Chrome I've done the "Add to Homescreen" option which hides the address bar completely and makes it behave like an app, which works very well indeed and gives a nice experience - so the search URL would not even be seen.

So I'm thinking using the eventbus is the cleaner and less confusing option for the user - what do you think?

marcelveldt: Yeah in that case it doesn't make any sense but still the eventbus is strange to communicate parameters to a page. I would just use the properties for that. You can call the router and tell the properties to send. (ab)using the eventbus or a shared stored is more for dialogs.

This is for example already use din ItemsListing.vue where we redirect an in-library search to the global search.

Scherm­afbeelding 2024-04-27 om 11 37 36

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
sprocket-9: Ok I've added this guard.

src/views/Search.vue
sprocket-9: Yes I tried this, but remember you can't router.push to the existing route.  If /search is already the active route behind the full screen player, pressing the stream title to search for it will not work with router.push again.  In this case we need a way to communicate between two active components which have no direct ancestry.

https://vuejs.org/guide/components/events.html mentions resorting to an external event bus is sometimes needed:

>Unlike native DOM events, component emitted events do not bubble. You can only listen to the events emitted by a direct child component. If there is a need to communicate between sibling or deeply nested components, use an external event bus or a [global state management solution](https://vuejs.org/guide/scaling-up/state-management).

marcelveldt: hmmm, I still don't like it that as now there are multiple ways to do the same.

So we should either find a way to accept a router push/replace with updated params or switch to eventbus from the other components as well.

I have an idea... what about we replace the private searchterm in the search component with a global one on the store ? That way you can change the global property from anywhere and contain all logic inside search. Just add a watcher in there to do the actual search and store it to the localstorage etc.

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
```diff
@@ -489,6 +501,20 @@ const radioNameClick = function (item: Radio | ItemMapping) {
...
+      search_term: stream_title,
+    });
+  } else {
+    localStorage.setItem('globalsearch', stream_title);
+    router.push({
+      name: 'search',
+    });
+  }
+  store.showFullscreenPlayer = false;
+};
```
marcelveldt:
```suggestion
  store.globalSearchTerm = stream_title;
  if (router.currentRoute.value.name !== 'search') {
    router.push({
      name: 'search',
    });
  }
  store.showFullscreenPlayer = false;
};
```

src/views/Search.vue
marcelveldt: Update: I have updated the current implementation to the global variable in the store so if you rebase this branch you should be able to build on top of that.

sprocket-9: Ah ok so this is using the other suggestion in the vue docs for communicating between components by using a global state object any component can access.  Good idea, this should make it nice and consistent then!

I've rebased and updated to use the new solution.

sprocket-9:
Just noticed that as it is now, if the metadata has no dash in it, it doesn't only make it non-clickable, it won't even show it.  I'll add a fix for that.

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
```diff
@@ -489,6 +502,13 @@ const radioNameClick = function (item: Radio | ItemMapping) {
   store.showFullscreenPlayer = false;
 };
 
+const radioTitleClick = function (streamTitle: string) {
+  // radio station title clicked
+  store.globalSearchTerm = streamTitle;
+  router.push({ name: 'search' });
```
marcelveldt, line 508: doesn't this issue a warning if the route is already active ? Keep an eye on it, if you see an error reported by the vue router then this needs to be guarded.

marcelveldt (approved):
Nice work again!

PR #488, opened 2024-05-07
Fix: Search ignores last press sometimes

jonbng:
Sometimes when searching you have to press space for it to trigger the searching update. I think its better if the keypress event listens to `keyup` instead of `keydown` to allow it to always have the latest value when searching. Correct me if I am wrong.


marcelveldt (approved):
Not sure if this is the right fix for the issue you describe but cant harm anyways

PR #502, opened 2024-05-13
Fix sort name and tidy

MelHarbour:
Currently the UI shows the text 'Alphabetically' for both the line items to sort by 'name' and 'sort_name'. This PR makes the latter consistent with the text label for 'sort_name_desc' and also tidies a few other sort order translations that are unused.


marcelveldt (approved):
Thanks!

PR #504, opened 2024-05-14
Match menu to tooltip

OzGav:
I noticed that the tooltip says Toggle view mode list/thumbs but the menu was different.


marcelveldt (approved):
I'm still not sure if it needs to be called panel view or thumbs but at least this makes it consistent

PR #506, opened 2024-05-15
Add support for child path ./home/player

Eraldometris:
Add support for child path ./home/player this will direct to homepage with the player in full screen.

### DIFF
--- a/src/plugins/router.ts
+++ b/src/plugins/router.ts
@@ -18,6 +18,17 @@ const routes = [
         component: () =>
           import(/* webpackChunkName: "home" */ '@/views/HomeView.vue'),
         props: true,
+        children: [
+          {
+            path: 'player',
+            name: 'player',
+            component: () => import('@/views/HomeView.vue'),
+            beforeEnter: (to, from, next) => {
+              store.showFullscreenPlayer = true;
+              next();
+            },
+          },
+        ],
       },
       {
         path: '/search',
--- a/src/plugins/router.ts
+++ b/src/plugins/router.ts
@@ -23,9 +23,8 @@ const routes = [
             path: 'player',
             name: 'player',
             component: () => import('@/views/HomeView.vue'),
-            beforeEnter: (to, from, next) => {
+            beforeEnter: () => {
               store.showFullscreenPlayer = true;
-              next();
             },
           },
         ],


marcelveldt:
Can you tell me a bit about the usecase for this ?

There is one problem with this approach which is that there is no player selected at this point.

I think it would be better to make the fullscreen player a resuseable component that takes any player so it can be launched from a page like /home/player/playerid. 

The fullscreen dialog is tied to the activeplayer (and activequeue) but it can also be used standalone.

Eraldometris:
Hi @marcelveldt,

`Can you tell me a bit about the usecase for this ?`
The use case that motivated me for this feature stems from home assistant. I have a smaller media card on my dashboard and have set up a tap action such that it navigates to the Music Assistant addon, to the home view. Though I would like it to navigate to the full screen player such that it feels more of an extension of the media card. (I.e. blow up the media card to a full screen player in music assistant).

`There is one problem with this approach which is that there is no player selected at this point. I think it would be better to make the fullscreen player a resuseable component that takes any player so it can be launched from a page like /home/player/playerid.`

That is true there is no default player selected, though that is fine as the purpose of this route is to to navigate to the current player. If the current player does not exist the default 'No player selected' is shown and seems recoverable (straight forward to rectify). I can't think of a use case to have a player pop up for a specific playerid that has not been selected. I can not comment on this use case. I do believe there is merit though to have basic functionality to navigate to the selected player. 

image

Eraldometris:
> There is one problem with this approach which is that there is no player selected at this point.
 
Hi @marcelveldt, 

I can route it back to home if there is no player selected? Does this work for you?

marcelveldt:
Sorry, no. I cant agree with this PR because this is calling a dialog from a route path.
That is just hacky/anti-pattern. I understand your usecase (and it has been requested a bit more often) but let's do this the right way by creating a separate view for the fullscreen player.

This solution would also allow you to open the fullscreen player for the actual player you clicked in the HA interface instead of whatever is set in MA.

Let me know if you want to take stab at implementing it yourself or you want me to implement it.

OzGav:
@Eraldometris comments/thoughts?

Eraldometris:
I have been thinking about this and how to implement @OzGav . The requirements are not very clear. I can detach the player to be a standalone view but it would still need to be able to interact with the home view, otherwise we won't be able to add music to queue etc.

To add music to queue we will need to assign the playerid as the selectedplayer and the view will need be able to navigate back to home view. This makes me question what's the reason of making it a separate view.

Happy for you guys to develop this if there is a clear vision I'm missing here.

marcelveldt:
Those are 2 different cases. Lets start by asking about your usecase.
Why and how are you opening the MA interface and for what player ?

Some options:

1) You click a specific player button/card in HA and you want to open the fullscreen player in MA for THAT player, for example in an iframe or whatever.

2) You click some button somewhere in HA that just launches MA interface on the fullscreen player dialog of the default.last active player.

3) Same as 2 but you want to select a player in one go.

If it's (2) you're looking for, I suggest we introduce some query param to define an action to select the player and open the full screen player. If it's 1

Eraldometris:
Hi @marcelveldt, 

I think option two is what I was trying to achieve.

If I implement this, I can make playerID a query parameter for /home route which will navigate to home view and set active player and full screen in store.

What do you think?

marcelveldt:
> Hi @marcelveldt,
> 
> I think option two is what I was trying to achieve.
> 
> If I implement this, I can make playerID a query parameter for /home route which will navigate to home view and set active player and full screen in store.
> 
> What do you think?

Yes that makes sense to me - you then just give a query param to the home view which is an action.
That is cleaner (imo) than abusing a route view to open a dialog.

The option to open a specific player fullscreen view (+ queue) by path is then something that we leave as feature request if there is real demand but maybe we should instead focus on some cards or standalone components that can be used for that so folks can include it in their HA dashboards.

So for this PR/request:

add prop to home view "showFullscreenPlayer" (boolean) and "setDefaultPlayerId" (or something alike that) and use a watch in the home view to handle it.

marcelveldt:
superseded by https://github.com/music-assistant/frontend/pull/585

PR #517, opened 2024-05-30
Add base shortcuts

max246:
This is part of the request #1645   , it adds shortcuts to the UI

Adding the media control for:
Mute
Volume Up/Down
Play/Pause
Stop
Prev/Next track

Unfortunately  you need to have the window focus to make it work since mediaSession doesnt seems to work with my integration


src/App.vue
```diff
@@ -55,6 +55,26 @@ onMounted(() => {
   window
     .matchMedia('(prefers-color-scheme: dark)')
     .addEventListener('change', setTheme);
+
+  const handleKeyDown = (e) => {
+      const key = e.key;
+      let keyDetected = true;
+
```
marcelveldt, line 66: this code needs a guard to handle the following cases:
- api connection not (yet) initialized or lost
- no active player 
- no queue actions available

src/layouts/default/HelperShortcuts.vue
marcelveldt, line 1: Please remove this - keep this PR simple to only provide media controls
In a follow up PR you can do more controls and add config options (to the settings view!)

src/App.vue
```diff
@@ -55,6 +55,26 @@ onMounted(() => {
...
+      else if (key === 'MediaStop') api.queueCommandStop(store.activePlayerQueue!.queue_id);
+      else if (key === 'MediaTrackPrevious') api.queueCommandPrevious(store.activePlayerQueue!.queue_id);
+      else if (key === 'MediaTrackNext')  api.queueCommandNext(store.activePlayerQueue!.queue_id);
+      else if (key === 'AudioVolumeUp') api.playerCommandVolumeUp(store.activePlayer?.player_id || '');
+      else if (key === 'AudioVolumeDown')  api.playerCommandVolumeDown(store.activePlayer?.player_id || '');
+      else if (key === 'AudioVolumeMute') api.playerCommandVolumeMute(store.activePlayer?.player_id || '', !store.activePlayer?.volume_muted  || false );
+      else keyDetected = false;
+    
+      if (keyDetected)
+        e.preventDefault();
```
marcelveldt: Please carefully test all cases where we also listen for keypresses, such as the library listings.
If you are watching a list like Artists or Albums, you should be able to just start typing to interactively search

src/layouts/default/HelperShortcuts.vue
max246: ok cool will do

marcelveldt (changes requested):
see in-line feedback - keep the PR small and simple and focus only on the media controls. Also, you are only accounting for the happy flow now. If the api connection is not there yet or no player is active this will crash

src/App.vue
max246: I did try and it wasnt affecting it

marcelveldt:
I've just tested this PR and it was not working, especially not in the subpages. 
So back to the drawing board, we might want to look into mediaSession again

OzGav:
@max246 Do you want to progress this?

max246:
> @max246 Do you want to progress this?

Yeah sorry I got bit busy at work. 
I need to dig up more how media access works since when I tried didn't seem to do anything. 

If you have any tips or you want to collaborate let me know

marcelveldt:
superseded by #769

PR #518, opened 2024-05-31
Centre the media controls (and simplify some CSS along the way)

JeffersonBledsoe:
This PR centres the media controls (play, next, shuffle, etc.) as it is currently slightly off-centre. I simplified the CSS a little along the way


src/layouts/default/PlayerOSD/Player.vue
```diff
@@ -1,8 +1,5 @@
 <template>
-  <div
-    :class="`${useFloatingPlayer ? 'mediacontrols-bg-2' : 'mediacontrols-bg-1'}`"
-    :style="`background: ${backgroundColor};`"
-  ></div>
+  <div class="mediacontrols-bg" :data-floating="useFloatingPlayer"></div>
```
marcelveldt, line 2: data-floating is a bit of a weird property name

src/layouts/default/PlayerOSD/Player.vue
```diff
@@ -14,25 +11,16 @@
...
   />
 
-  <div
-    class="mediacontrols"
-    :style="`padding: ${
-      getBreakpointValue({ breakpoint: 'phone' }) ? 3 : 10
-    }px ${getBreakpointValue({ breakpoint: 'phone' }) ? 10 : 10}px;`"
-  >
-    <div :class="`mediacontrols-left-${getBreakpointValue('bp3') ? '1' : '2'}`">
+  <div class="mediacontrols" :data-mobile="mobile">
```
marcelveldt, line 14: see my previous comment: whats up with the "data-" prefixed value ?

src/layouts/default/PlayerOSD/Player.vue
```diff
@@ -134,6 +123,8 @@ defineProps<Props>();
 
 // global refs
 const theme = useTheme();
+// Custom breakpoint for compatibility with `getBreakpointValue`. Can replace once we switch to using built-in Vuetify breakpoints
```
marcelveldt, line 126: Oh yeah, I looked into this a couple of weeks ago and its a complete mystery to me why we aren't just using default vuetify for this. My best guess is that this was not yet available back then in the vuetify beta's.

To me, we should now go (back) to using as much native vuetify as possible and just follow along

src/layouts/default/PlayerOSD/Player.vue
JeffersonBledsoe: I'm using the `data-*` to reflect the state in the HTML ([Cube CSS](https://cube.fyi/exception.html#why-data-attributes) does the same thing). We could use a class here and do something like `class="mediacontrols-bg mediacontrols-bg__floating"`, but I find the data attribute a little clearer.

src/layouts/default/PlayerOSD/Player.vue
JeffersonBledsoe: Replied in the previous comment 🙂

src/layouts/default/PlayerOSD/Player.vue
JeffersonBledsoe: Agreed. I did try going straight to using the vuetify config with this change, but the breakpoints are out-of-sync for some things, causing some buttons to be display and others not at certain widths. We should do one single change to all of them at the same time to move to the Vuetify config to better catch any bugs that might crop up. I've slightly started this work in https://github.com/JeffersonBledsoe/music-assistant-frontend/tree/get-breakpoint-tidy-1

marcelveldt (approved):
Nice!

PR #529, opened 2024-06-11
Fix playbar small size

JeffersonBledsoe:
Only briefly tested, but this should fix the progress bar being really small and the now playing controls jumping about after my last PR


marcelveldt (approved):
lgtm, thanks!

PR #548, opened 2024-06-21
Various minor improvements

OzGav:


src/translations/en.json
```diff
@@ -300,7 +300,7 @@
         },
         "announce_volume_strategy": {
             "label": "Volume strategy for Announcements",
-            "description": "When an announcement is being broadcast to this player, how should the volume be adjusted temporary (use in combination with the volume value below).",
+            "description": "When an announcement is being broadcast to this player, how should the volume be adjusted temporarily (use in combination with the volume value below).",
```
MelHarbour: If you're going to fix the grammar here, you probably want a question mark on the end of the line at the same time.

OzGav: I said grammar not punctuation! But yeah I will fix that and some other things :-) Thanks!

src/translations/en.json
```diff
@@ -397,13 +397,8 @@
     "sort": {
         "duration": "Duration",
         "position": "Position",
-        "position DESC": "Position (descending)",
```
marcelveldt, line 400: How is it possible that this reverted ?
this should be position_desc as key in the language file.

are we 100% sure that lokalize is not overwriting our base file ?

MelHarbour: Some corrections I made to the en base got reverted by Lokalise automatically I think. ISTR that there's some setting in Lokalise to allow changing/deleting existing entries.

OzGav: Not sure. I wonder if it happened because I accidentally changed the English file…although it should have had the latest..? I will keep a close eye on things once this goes through.

marcelveldt (approved):
Lets keep an eye how this pans out with Lokalize

PR #559, opened 2024-06-26
Lokalise-cron-download

jozefKruszynski:
This should download the translations every Monday at 2:00 am

### DIFF
--- a/.github/workflows/lokalise-download.yml
+++ b/.github/workflows/lokalise-download.yml
@@ -1,0 +1,40 @@
+name: Download Lokalise translations
+on: 
+  schedule:
+  - cron: "0 2 * * 1"
+jobs:
+  build:
+    if: ${{ github.event.workflow_run.conclusion == 'success' || github.event_name == 'workflow_dispatch' }}
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


OzGav (dismissed):
Fingers crossed

jozefKruszynski:
I tested the download bit on my machine and it works as expected, the merge request bit of the workflow might need some tweaking

OzGav:
I was just looking to see where it says to download all languages. That may be a default?

jozefKruszynski:
It is

OzGav:
I didn't change the others here but just in case this is important these other options are set

image

image

I see these options also in the API

jozefKruszynski:
Added those options

OzGav:
I know you said you tested this but I see some options have a = and some don't? Is that correct?

OzGav (dismissed):
Lets see what happens!

OzGav (approved):
Looks good.

jozefKruszynski:
> I know you said you tested this but I see some options have a = and some don't? Is that correct?

Supposedly booleans require = and strings don't.

PR #563, opened 2024-06-30
Remove languages not mostly translated

OzGav:


marcelveldt (approved):
This still allows the language to be auto selected on browse preference but OK

PR #569, opened 2024-07-02
Fix favicon in index.html

ulflulfl:
I've wondered, why the favicon (the small icon in the tab) is not shown in Firefox

I've removed the line with favicon.svg, as that file doesn't exist.
In line 13 is already an entry for favicon.ico which should be correct (at least it exists in public/favicon.ico and is shown correctly in Firefox)

Just realized, that Chrome displays the favicon correctly (it seems to somehow ignore the first wrong "link rel=icon" entry or maybe always uses the last entry).

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,7 +4,6 @@
     <meta charset="utf-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width,initial-scale=1.0" />
-    <link rel="icon" href="favicon.svg" type="image/svg+xml">
     <link rel="alternate icon" href="favicon.ico" type="image/png" sizes="16x16">
     <link rel="apple-touch-icon" href="apple-touch-icon.png" sizes="180x180">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4,7 +4,6 @@
     <meta charset="utf-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width,initial-scale=1.0" />
-    <link rel="icon" href="favicon.svg" type="image/svg+xml">
```
marcelveldt, line 7: let's keep the icon and fix that instead of removing it entirely

ulflulfl: This removed line is a duplicate "link rel="icon"" entry .

The correct icon entry is already in line 13, which you don't see in the small snippet.

marcelveldt: Ah my mistake - I didnt spot that indeed!

marcelveldt (approved):
Thanks!

ulflulfl:
I've checked with 2.1.0b14 and the favicon shows up correctly in Firefox now.

Thanks!

PR #585, opened 2024-07-21
Open fullscreen player from URL

marcelveldt:
Allow displaying the fullscreen player for a specific player (or the last known) by adding "player=<name or id>" to the home URL. You can use a player id, player name or simply true to open the last known.

So for example 

http://192.168.1.1:8095/#/home?player=true

or 

http://192.168.1.1:8095/#?player=true

or 

http://192.168.1.1:8095/#/home?player=aabbccddeeffgg

or 

http://192.168.1.1:8095/#/home?player=Livingroom


jonasjeeliasson:
Is it possible to use this when you use Music Assistant within Home assistant, i.e when the URL looks like this:
`https://my-ha/d5369777_music_assistant/ingress`? Tried adding `player=true` to it but nothing happens.

marcelveldt:
> Is it possible to use this when you use Music Assistant within Home assistant, i.e when the URL looks like this: `https://my-ha/d5369777_music_assistant/ingress`? Tried adding `player=true` to it but nothing happens.

yes, but you most probably also need the #home part

EDIT: No it wont work. You will need to pass it to the internal (iframe) url of the ingress panel.

Eraldometris:
> > Is it possible to use this when you use Music Assistant within Home assistant, i.e when the URL looks like this: `https://my-ha/d5369777_music_assistant/ingress`? Tried adding `player=true` to it but nothing happens.
> 
> 
> 
> yes, but you most probably also need the #home part
> 
> 
> 
> EDIT: No it wont work. You will need to pass it to the internal (iframe) url of the ingress panel.

Hi @marcelveldt thanks for implementing this feature! I've been trying to find a way to pass the url to the iframe url. It doesn't seem to be straightforward, have you managed to think of or implement this?

PR #674, opened 2024-10-03
feat: update key structure and add descriptions

whc2001:
Already did it so gonna put it here, feel free to improve or use the @SantiagoSotoC's version


marcelveldt (approved):
This is the right way to do it so I'll merge this over the other open PR

PR #680, opened 2024-10-07
Add UI changes for Don't stop the music.

MarvinSchenkel:


src/translations/en_GB.json
marcelveldt, line 1: dont edit translations from a PR, instead use lokalize. Only en.json may be edited

marcelveldt (dismissed):

PR #694, opened 2024-10-21
add horizontal scrolling in carousel

nikscha:
When scrolling horizontally on a laptop touchpad or on a second scroll wheel (like on the Logitech MX Master), the carousel on the Home page now scrolls horizontally.


marcelveldt (approved):
LGTM
Thanks!

PR #745, opened 2024-11-22
feat: Add a frameless param

felipetoffolo1:
I'm adding a param called frameless

So that different view of MASS can be used in cards ([original suggestion](https://github.com/orgs/music-assistant/discussions/954#discussioncomment-11348170))

- Added 2 new params
  - frameless: Load just the content without the footer and navigation. It won't be saved in the store so that it goes away everytime we reload
  - targetPlayer: using that to select the activePlayer. I know we have a `player` param already. Put that param open the fullscreen player. And it might not be what we want.

- Moved code that handled `player` to the defaultView so that everything is in the same place

https://github.com/user-attachments/assets/9091d7b5-3901-453c-a4c6-ac9c18896079

### DIFF
--- a/src/layouts/default/Default.vue
+++ b/src/layouts/default/Default.vue
@@ -1,12 +1,12 @@
-<template>  
-  <v-app  v-if="store.connected">
+<template>
+  <v-app v-if="store.connected">
     <MainView v-if="framelessState" />
     <template v-else>
       <PlayerSelect />
       <MainView />
       <Footer />
     </template>
-    </v-app>
+  </v-app>
   <v-overlay
     v-else
     class="centeredoverlay"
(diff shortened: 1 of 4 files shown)


marcelveldt:
maybe keep "player" param to select the default player (on any view) and introduce a new param like "openview" or something like that ?

src/layouts/default/Default.vue
```diff
@@ -23,6 +26,48 @@ import Footer from "./Footer.vue";
...
+    store.showFullscreenPlayer = !!player;
+  },
+  { immediate: true },
+);
+
+watch(frameless, (frameless)=>{
+  if(frameless){
+    framelessState.value = true
+  }
+}, {immediate: true})
```
marcelveldt:
```suggestion
}, { immediate: true })
```

src/layouts/default/Default.vue
```diff
@@ -23,6 +26,48 @@ import Footer from "./Footer.vue";
...
+      const newPlayerId = Object.values(api.players).find(p =>{
+        return p.player_id === newActivePlayer || p.display_name.toLowerCase() === newActivePlayer.toLowerCase()
+      })?.player_id
+
+      if(newPlayerId){
+        store.activePlayerId = newPlayerId;
+
+      }
+
+    //showFullscreenPlayer if the param player is defined at all
```
marcelveldt:
```suggestion
    // showFullscreenPlayer if the param player is defined at all
```

felipetoffolo1:
> maybe keep "player" param to select the default player (on any view) and introduce a new param like "openview" or something like that ?

@marcelveldt I thought of that first. But I was not sure about adding a breaking change. If we are fine with that I'm happy to change

marcelveldt:
> @marcelveldt I thought of that first. But I was not sure about adding a breaking change. If we are fine with that I'm happy to change

I think its fine as we are still early on in the cycle and this feature is not very known.
Let's just make it work great instead of half baken. We should just be aware to mention it in the docs and release notes once this gets released.

felipetoffolo1:
> > @marcelveldt I thought of that first. But I was not sure about adding a breaking change. If we are fine with that I'm happy to change
> 
> I think its fine as we are still early on in the cycle and this feature is not very known. Let's just make it work great instead of half baken. We should just be aware to mention it in the docs and release notes once this gets released.

@marcelveldt sounds good, let know what you think of it. Happy to rename anything if you want it

marcelveldt (approved):
Looks good, nice work!

OzGav:
This will be a great update! Thanks @felipetoffolo1

skavan:
wait - so https://myurl:xxxx/#/playlists?frameless=true&targetPlayer=MyBasementPlayer will give me the playlist window, without the media player and the menu bar?

THIS-IS-HUGE. 
When will it be released into the beta?
If I could beg for one more addendum:
`?OnlyShowFavorites=true&sort=playcount&view=list&toggleAlbumArtist=true`
This isn't to replace the icons on the view, but to preselect the defaults that the user wants to see, without additional clicks.

At this point, you have effectively made MA totally modular. 
The 'views' will still be your `opinionated` views (which are responsive and great) -- but we will be able to do almost anything within HA without having dedicated MA-HA cards. Awesome.

s.

PR #757, opened 2024-11-29
chore: Make it minimally accessible by tab

felipetoffolo1:
Add minimal usability and accessibility using tabs

What did I do:
- Removed some button elements that were buttons but were doing nothing useful but were acessible using tab
- Change bottom navigation from tabs to buttons
- Re-ordered DOM so that we would tab into the menu first
- Stopped rendering the players select list when it was not opened (it was accessible using tab)

Next step:
- Many things but we would have to take component by component I think

https://github.com/user-attachments/assets/78c86dd4-100e-4127-9e10-37767a45151b


src/components/Toolbar.vue
```diff
@@ -1,11 +1,11 @@
 <template>
   <v-toolbar :color="color" class="header">
     <template v-if="icon" #prepend>
-      <v-btn :icon="icon" size="large" @click="emit('iconClicked')" />
```
marcelveldt, line 4: isn't this used for the back button ?

felipetoffolo1: Do you know in which page it's actually used?
I can add some kind of check if needed, but I could not find a place where it was used

marcelveldt: details page of an item (such as artist details)

felipetoffolo1: shold be good now

marcelveldt (approved):
Thanks!

PR #767, opened 2024-12-02
Add Marquee to Make Long Text Readable

maximmaxim345:
# Current Situation

Currently, Music Assistant on mobile devices cuts off long titles, albums, and artist names, making it impossible to read them completely.
This is especially problematic with Classical Music, where titles often include the composer, composition name, and movement.

# Overview

To address this issue, this PR adds a `MarqueeText` component, which automatically scrolls text that doesn't fit on screen.

To improve readability, `MarqueeText` includes the following features for use in Music Assistant:
- Constant Scrolling Speed: Text scrolls at a consistent speed, regardless of length
- Synchronization: Multiple instances of `MarqueeText` can be synchronized with `MarqueeTextSync` for coordinated scrolling

# MarqueeTextSync

When a `MarqueeTextSync` object is passed through the `sync` prop of `MarqueeText`, the reset animation will be synchronized across all instances using the same `MarqueeTextSync` object.
Additional delay is added at the start of the animation, so waiting widgets will pause with the text at the beginning.

This provides the following advantages:
- Reset animation is shared across all widgets to prevent chaotic jumping
- The constant scrolling speed makes reading with multiple marquees on screen considerably easier
- When waiting is required (for sync), the screen time for the beginning of the text is maximized, as it's often the most important part

Only visible `MarqueeText` widgets will be animated, ensuring animations won't be slowed down by off-screen `MarqueeText` components with very long text.

# Use of MarqueeText in this PR

In this PR, MarqueeText is added in the following places:
- Fullscreen Player Queue (For currently playing track, and on mouse hover)
- Fullscreen Player (Title, Artist, and Album Name, with different `MarqueeTextSync` than the Queue)
- Minimized Player (Artist and Title)
- InfoHeader (when viewing an Album/Artist/Playlist/...)

Scrolling is done using CSS, while CPU usage seems to be minimal (tested using Firefox Profiler).

# Screen Recordings of this PR in Action

https://github.com/user-attachments/assets/fb28e671-b439-4394-be90-115d49cc409e

https://github.com/user-attachments/assets/4ec67339-57c1-48bc-877d-80a00231582c

https://github.com/user-attachments/assets/35e06376-0952-477f-a5a9-ced10f073cc9

### DIFF
--- a/src/components/InfoHeader.vue
+++ b/src/components/InfoHeader.vue
@@ -83,7 +83,9 @@
             style="padding-left: 10px"
           />
           <v-card-title v-else>
-            {{ item.name }}
+            <MarqueeText :sync="marqueeSync">
+              {{ item.name }}
+            </MarqueeText>
           </v-card-title>
 
           <!-- other details -->
@@ -119,20 +121,22 @@
                 color="primary"
                 icon="mdi-account-music"
               />
-              <span
-                v-for="(artist, artistindex) in item.artists"
-                :key="artist.item_id"
-              >
-                <a style="color: accent" @click="artistClick(artist)">{{
-                  artist.name
-                }}</a>
+              <MarqueeText :sync="marqueeSync">
                 <span
-                  v-if="artistindex + 1 < item.artists.length"
-                  :key="artistindex"
-                  style="color: accent"
-                  >{{ " / " }}</span
+                  v-for="(artist, artistindex) in item.artists"
+                  :key="artist.item_id"
                 >
-              </span>
+                  <a style="color: accent" @click="artistClick(artist)">{{
+                    artist.name
+                  }}</a>
+                  <span
+                    v-if="artistindex + 1 < item.artists.length"
+                    :key="artistindex"
+                    style="color: accent"
+                    >{{ " / " }}</span
+                  >
+                </span>
+              </MarqueeText>
             </v-card-subtitle>
 
             <!-- playlist owner -->
@@ -146,7 +150,9 @@
                 small
                 icon="mdi-account-music"
               />
-              <a style="color: primary">{{ item.owner }}</a>
+              <MarqueeText :sync="marqueeSync">
+                <a style="color: primary">{{ item.owner }}</a>
+              </MarqueeText>
             </v-card-subtitle>
 
             <v-card-subtitle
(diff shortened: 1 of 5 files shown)


marcelveldt:
I appreciate the PR and I think this addition will be liked by many.
However I do not like the fact that multiple elements are scrolling at the same time, which makes the screen super busy and distracting even. Also I think that the marquee could be a bit slower. Or maybe pause shortly between scrolls, like the Sonos app is doing.

I think for the full list of queue tracks we should only scroll the currently active item, and not the entire list.

maximmaxim345:
Thanks for the feedback.

I will make the scrolling slower (maybe half of the current speed). I will look into how SONOS handles it later, but this PR already pauses for 2 seconds at the end of the text before resetting. How about extending that to 4 seconds?

As for the Queue, I absolutely agree with you about the busyness (I mainly added the marquee to absolutely everything to see what works). As an alternative solution, I can propose: adjusting the 3-dot menu on a queue item to show the title/artist name in full with wrapping.

Keeping scrolling on for the currently active queue item sounds good. Then we can synchronize scrolling between the left and right side of the full-screen player (first video; currently, it was syncing separately due to different text lengths). This should further reduce busyness a bit.

Let me know what you think of this. I will try to adjust this PR in the coming week.

marcelveldt:
Yeah sounds good on all points. 
As for the queue items I would indeed not let them all scroll (as that makes up for a super busy screen) but the suggestion to  "reveal" the full title with a click of the contextmenu sounds good. Also, on desktop OS we can consider to scroll on mouse over.

maximmaxim345:
Just updated this PR as discussed.
Also added your idea of scrolling when hovering with a mouse.

You can see the updated screen recordings above.

I also looked into how the SONOS app handles the marquee:
In my opinion, the main weakness of SONOS's approach is in cases where the album title is relatively short while the track title is long. In that case, for the majority of time, the album name is entirely hidden due to the scroll. This PR would show the beginning of the album name instead, which is considerably more useful than looking at a blank space.

As for revealing the full title in the context menu, I can add that in a separate PR later.

marcelveldt:
Perfect!

marcelveldt (approved):
Nice work @maximmaxim345 !

PR #769, opened 2024-12-06
Browser Media Controls and Metadata

DerDeathraven:
# Overview
this PR adds the PlayerBrowserMediaControls component which is managing the comunication with the browser concerning the media buttons and the metadata for the current song

# The issue before 
this is project is awesome and i really enjoy my time with it so far. 
one thing that annoyed me was that music could not be controlled with my media buttons on my keyboard.

# Nice to mention
this PR also includes 5_min_of_silence which i took from https://github.com/anars/blank-audio/blob/master/5-minutes-of-silence.mp3

this is used so that the browser thinks something is being played 

i thought about adding a feature flag for this since i can imagine that this is not the wanted behaviour for some people but i was not familiar enough with the construct so i left it be for now. If its wanted I (or you) can add it 

image


src/helpers/useMediaBrowserMetaData.ts
```diff
@@ -0,0 +1,46 @@
...
+} from "@/plugins/api/interfaces";
+import { store } from "@/plugins/store";
+import { onMounted, watch } from "vue";
+
+function playerMediaToMetadata(item: QueueItem) {
+  let artist: string | undefined;
+  let album: string | undefined;
+
+  //here we cast to Track to access properties that are only available on Track
+  if (item.media_item!.media_type === "track") {
```
marcelveldt: Let's use the MediaType Enum here

src/helpers/useMediaBrowserMetaData.ts
```diff
@@ -0,0 +1,46 @@
...
+    const currentMedia = item.media_item as Track;
+    artist = currentMedia.artists.map((a) => a.name).join(", ");
```
marcelveldt, line 12: What about the case where a non-track item is being played, such as a radio station or some raw audio file ?
I think the code should handle that instead of crashing

src/helpers/useMediaBrowserMetaData.ts
```diff
@@ -0,0 +1,46 @@
...
+  let currentMediaData: MediaMetadata | undefined;
+
+  //watch the current media to update the metadata
+  watch(
+    () => store.curQueueItem,
+    (newMedia) => {
+      if (!newMedia || !newMedia.media_item) return;
+      const newMediaMetaData = playerMediaToMetadata(newMedia);
+      //Lets make sure that the new media isnt spammed
+      if (currentMediaData?.title === newMediaMetaData.title) return;
```
marcelveldt: I would compare the uri here instead of title.

src/layouts/default/PlayerOSD/PlayerBrowserMediaControls.vue
```diff
@@ -0,0 +1,85 @@
...
+
+// MediaSession setup
+onMounted(() => {
+  window.addEventListener("click", () => {
+    //There is a safety rule in which you need to interact with the page for the audio to play
+    interactedRef.value = true;
+    updateMediaState(store.activePlayer?.state);
+  });
+
+  navigator.mediaSession.setActionHandler("play", () => {
```
marcelveldt, line 66: I'm not familiar with these browse api's but would there also be support for volume control ? Or are these player-related commands the only ones ?

src/App.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <router-view />
+  <PlayerBrowserMediaControls />
```
marcelveldt, line 3: What would be the downside of having this component always loaded ? Does it block playback of other browser tabs ? I think not right ? So it should be safe to just enable this by default for everyone and if somebody starts complaining, we can add a toggle to disable ?

marcelveldt (changes requested):
Nice PR, I'm sure people will love this.
I left a couple of small comments and questions

src/layouts/default/PlayerOSD/PlayerBrowserMediaControls.vue
DerDeathraven: no, those are sadly not supported. there is an example with all supported items here

https://developer.mozilla.org/en-US/docs/Web/API/MediaSession

src/helpers/useMediaBrowserMetaData.ts
DerDeathraven: i am not deep enough in the codebase. as far as i understand this code shouldnt crash in those cases and rather just show the file name or have i got something wrong there?

src/App.vue
DerDeathraven: on phones it could be that it blocks the sound from other sources. on PCs it shouldnt. i would just go with it and fix it if someone is complaining

src/helpers/useMediaBrowserMetaData.ts
DerDeathraven: done

src/helpers/useMediaBrowserMetaData.ts
DerDeathraven: done

src/helpers/useMediaBrowserMetaData.ts
marcelveldt: Yeah, I just spotted you already check if the media item is track, I missed that part last time. So all good.

marcelveldt (approved):
Nice work!

PR #798, opened 2024-12-30
Make DSP Configuration UI More Consistent

maximmaxim345:
## Overview

This PR primarily makes sliders and switches use the primary color, as found in the rest of Music Assistant (notably the existing Settings Menu).

The other visual adjustments, as seen in the screenshots, increase the contrast between the primary color and the background (and include some margin adjustments resulting from these changes).

## Screenshots

new-dark

new-light

### Reference Screenshots before this PR

old-dark

old-light

### DIFF
--- a/src/components/dsp/DSPParametricEQ.vue
+++ b/src/components/dsp/DSPParametricEQ.vue
@@ -1,21 +1,23 @@
 <template>
   <v-container class="pa-2">
     <!-- Frequency Response Graph with Dark Theme Support -->
-    <v-card class="mb-4" elevation="2">
+    <v-card elevation="0" color="transparent">
       <div ref="graphContainer" class="graph-container">
         <canvas ref="canvas" class="frequency-graph"></canvas>
       </div>
     </v-card>
 
     <!-- Band Management Section -->
-    <v-card class="mb-4" elevation="2">
-      <v-card-title class="d-flex align-center px-4 py-2" />
-
+    <v-card
+      elevation="0"
+      color="transparent"
+      class="border-t border-b rounded-0"
+    >
       <!-- Band Selection with Visual Indicators -->
       <v-card-text class="pa-0">
         <v-chip-group
           v-model="selectedBandIndex"
-          class="mb-4 pa-2"
+          class="mb-0 pa-2"
           mandatory
           selected-class="primary"
         >
@@ -38,7 +40,7 @@
     </v-card>
 
     <!-- Band Controls Card -->
-    <v-card v-if="selectedBand" elevation="2">
+    <v-card v-if="selectedBand" elevation="0" color="transparent">
       <v-card-text>
         <!-- Band Header -->
         <div class="d-flex align-center mb-4">
@@ -69,20 +71,15 @@
               :label="$t('settings.dsp.parametric_eq.filter_type')"
               variant="outlined"
               density="comfortable"
-              class="mb-4"
             />
           </v-col>
 
           <v-col cols="12">
-            <DSPSlider
-              v-model="selectedBand.frequency"
-              type="frequency"
-              class="mb-4"
-            />
+            <DSPSlider v-model="selectedBand.frequency" type="frequency" />
           </v-col>
 
           <v-col v-if="showGainParameter(selectedBand.type)" cols="12">
-            <DSPSlider v-model="selectedBand.gain" type="gain" class="mb-4" />
+            <DSPSlider v-model="selectedBand.gain" type="gain" />
           </v-col>
 
           <v-col cols="12">
--- a/src/components/dsp/DSPParametricEQ.vue
+++ b/src/components/dsp/DSPParametricEQ.vue
@@ -47,7 +47,7 @@
             :label="$t('settings.dsp.parametric_eq.enable_band')"
             density="comfortable"
             hide-details
-            inset
+            color="primary"
           />
           <v-spacer />
           <v-btn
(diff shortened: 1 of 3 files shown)


OzGav:
I appreciate your efforts. This looks good to me!

marcelveldt (approved):

OzGav:
I am just starting to play with this some more now. The following are just my thoughts/ opinions with no UI precedent so feel free to disagree!

I note that when adding a filter a blue/cyan dot appears on the left to indicate it is enabled (and selected or not) and no dot when it is disabled. I wonder if a green (enabled) and red (disabled) dot may be more intuitive? Or maybe leave the current enabled indication and just have a red dot (or an x) to show it is disabled?

I am no DSP whiz so this may be a dumb question but does it make sense to be able to add multiple of the same filter?

Lastly, and not a UI question, we need to add some info to the docs about this awesomeness. Probably some basic info here https://music-assistant.io/player-support/ and maybe advanced info here or amendments to this https://music-assistant.io/faq/tech-info/ If you have any raw text or links to info I can paraphrase then I can do the actual PRs

P.S. Are you on our Discord server?

maximmaxim345:
> I am just starting to play with this some more now. The following are just my thoughts/ opinions with no UI precedent so feel free to disagree!

No problem! Not many music players offer such a DSP feature. The only one that comes to mind is ROON: https://help.roonlabs.com/portal/en/kb/articles/muse

> I note that when adding a filter a blue/cyan dot appears on the left to indicate it is enabled (and selected or not) and no dot when it is disabled. I wonder if a green (enabled) and red (disabled) dot may be more intuitive? Or maybe leave the current enabled indication and just have a red dot (or an x) to show it is disabled?

Here are my thoughts on the current design:
- The line represents the audio path from the audio file (top) to the speaker (bottom)
- A dot represents a filter that changes the signal, so disabled filters don't have a dot since they don't touch the audio signal
- It would be great to change the appearance of the dot to more clearly communicate this; however, this uses Vuetify's timelines, so a completely custom widget would need to be written

In case we show a red dot, this could also be interpreted as either still running the filter partly (I don't know, maybe a filter has a built-in preamp that will still be applied when disabled). An 'x' could also be understood as cutting the audio signal.

> I am no DSP whiz so this may be a dumb question but does it make sense to be able to add multiple of the same filter?

I can think of a couple of scenarios:
- Having one EQ measured (with a microphone in REW or from https://autoeq.app/) and a second EQ applying adjustments to sound preference (like adding more bass)
- We already support multiple filter types (the complex PEQ and the simple Tone Control filter). So having multiple PEQs is not that far off
- I am also planning to add Compressors/Limiters in the future. The order of having those and EQs will change the sound (this is the reason why you can already change the order of filters)

> Lastly, and not a UI question, we need to add some info to the docs about this awesomeness. Probably some basic info here https://music-assistant.io/player-support/ and maybe advanced info here or amendments to this https://music-assistant.io/faq/tech-info/ If you have any raw text or links to info I can paraphrase then I can do the actual PRs

That would be very nice!
I will send you some resources/texts once I find some time.

> P.S. Are you on our Discord server?

Yes I am, but I don't check any notifications.

On another note, I will soon open PRs for the following features:
- Import and Export of EqualizerAPO Presets (also supports import of REW filters). The APO format is widely supported by most applications, so for example https://autoeq.app/ filters could then be simply imported.
- Saving and loading of DSP presets. Useful for having DSP for multiple different listening positions, or just using headphones on different devices.

OzGav:
No problems on the Discord question I just thought it might be easier to converse vice this medium but this is working!

Your explanation of the UI sounds good. Looking at it again having read your response it makes perfect sense to me. I think once the docs are in place this will then be clear for all.

As for the docs, thanks for the Roon link, I think I will borrow some of their explanatory ideas!

I am away on holidays at the moment but I’ll see if I can make a start on the docs as this is so good, and as you say a very unique feature, that I would like to make sure everyone knows of its existence and how to use it.

OzGav:
One thing I think is missing from the UI is an indication of when the DSP settings are active. Correct me if I’m wrong but right now I think the idea is they are active by default but with no optional filters. You can then deactivate the DSP setting completely. 

At the moment I can’t tell, without going into the player settings, if the DSP is active or if any optional filters are added. 

Question is should the settings be enabled by default? (I understand they do nothing by default as the gains are set to zero)

Regardless of the answer to that I think an indication of the status should at least be done in the player settings and maybe via an icon in the Player List (if it’s active)?

maximmaxim345:
> No problems on the Discord question I just thought it might be easier to converse vice this medium but this is working!

I just haven't had the time to clean up my Discord from all the spam. Maybe I'll DM you if I do so.

> One thing I think is missing from the UI is an indication of when the DSP settings are active. Correct me if I’m wrong but right now I think the idea is they are active by default but with no optional filters. You can then deactivate the DSP setting completely.

> Question is should the settings be enabled by default? (I understand they do nothing by default as the gains are set to zero)

You are mostly right. I initially set DSP to be active by default to accommodate users of the old EQ system (which was primarily available on Airplay). This would automatically add a Tone Control filter to their DSP settings. This approach was part of my original migration logic, though it was never merged into the main branch.

But I've since rewritten part of the migration logic (merged into server), so that we now could only enable the DSP if the user really used the old EQ.

> At the moment I can’t tell, without going into the player settings, if the DSP is active or if any optional filters are added.

This should not be an issue anymore if the DSP is disabled by default.
I will open a PR for that!

> Regardless of the answer to that I think an indication of the status should at least be done in the player settings and maybe via an icon in the Player List (if it’s active)?

This is definitely something we need to think about.
Maybe somewhere in/around/beneath the Streamdetails on the fullscreen player?
I mean replay gain is shown there, which is technically also a DSP??

PR #800, opened 2025-01-02
Add import and Export buttons to the Parametric Equalizer DSP

maximmaxim345:
## Overview

This PR adds support for importing and exporting Parametric Equalizer presets.

Equalizer APO's file format is widely adopted in various applications, and this import functionality also works seamlessly with REW's  `Export filter settings as text` file format.
While not all filter types are supported (Modal, BP, LSC x dB, HS x dB, AP, LS with dB slope, HS with dB slope), the most important ones are functional.

Currently, the `Preamp` option is not supported. This may change in the future.

## Testing

I tested this with:
- An Equalizer preset exported from REW
- An Equalizer preset exported from EasyEffects (as APO preset)
- An example REW-compatible preset from https://housecurve.com/
- Importing a Music Assistant exported preset into EasyEffects

## Screenshot

image

### DIFF
--- a/src/components/dsp/DSPParametricEQ.vue
+++ b/src/components/dsp/DSPParametricEQ.vue
@@ -1,5 +1,26 @@
 <template>
   <v-container class="pa-2">
+    <!-- Import/Export Buttons to Load/Save Equalizer APO Settings -->
+    <div class="d-flex justify-end">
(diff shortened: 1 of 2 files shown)


marcelveldt (approved):
Nice work again!

PR #801, opened 2025-01-04
Fix: Improve spacing/padding consistency accross DSP Settings

maximmaxim345:
Previously Input settings had considerably less padding then Output settings and the Tone Controls filter. This PR adds the missing `<v-card-item>` to fix this.
Previously padding was inconsistent across the DSP Settings. 
This PR addresses this by using `<v-card-item>` for items inside the `<v-card>`, except `<DSPSlider>`, which already has padding on its own.
Additionally this PR makes labels of `<DSPSlider>` a constant width.

# Screenshot of current beta
image

# Screenshots with this PR

input
tone control
peq
output

### DIFF
--- a/src/components/dsp/DSPParametricEQ.vue
+++ b/src/components/dsp/DSPParametricEQ.vue
@@ -1,7 +1,7 @@
 <template>
   <v-container class="pa-2">
     <!-- Import/Export Buttons to Load/Save Equalizer APO Settings -->
-    <div class="d-flex justify-end">
+    <v-card-item class="d-flex justify-end">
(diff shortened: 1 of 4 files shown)


marcelveldt:
question: why not remove the padding alltogether and remove the v-card-item from the output card ?

maximmaxim345:
> question: why not remove the padding alltogether and remove the v-card-item from the output card ?

Good point! I changed the PR so that `<v-card-item>` (or an equivalent padding) is only used on non sliders.
I also tried to remove padding completely, but that hurt readability by a bit.

marcelveldt (approved):
Looks good!

PR #808, opened 2025-01-07
fix: Use a better solution to watch params

felipetoffolo1:
As [mentioned here](https://github.com/orgs/music-assistant/discussions/954#discussioncomment-11758079), sometimes the URL query params are not updating correctly unless the cache is cleaned.

I'm changing the way we watch for params changes to use the [way vue-router recommends.](https://router.vuejs.org/guide/essentials/dynamic-matching.html#Reacting-to-Params-Changes)
I believe this should solve the issue.


marcelveldt (approved):
Nice catch!

PR #809, opened 2025-01-10
Fix: missing playername on webkit/safari

maximmaxim345:
#767 introduced a regression on webkit/safari, where the player name is not visible.

# Screenshot before
image
In the order: Chromium (Brave), Firefox, and Webkit (GNOME Web)

# Screenshot with this PR

image
In the order: Chromium (Brave), Firefox, and Webkit (GNOME Web)


marcelveldt (approved):
Nice catch!

PR #810, opened 2025-01-11
Display the used volume normalization mode/values instead of target

maximmaxim345:
This is PR is a rebase of #651.

# Original description

Displaying the target loudness in the stream details display isn't very useful, since that's a user-configured setting and does not change depending on what is being played. In place of that number, show information about what volume normalization is currently being applied.

If measurements are available, and are being used, it will display either the track or album measured loudness depending on whether `prefer_album_loudness` is set.

Otherwise, it will display "Dynamic volume normalization" if loudnorm is being used in dynamic mode - or "Fixed gain correction" if fixed gain is being applied.

If no volume normalization is being applied, the loudness information row will not be displayed.

image


marcelveldt (approved):
Nice!

PR #812, opened 2025-01-13
Fix beta and stable docs website addresses

OzGav:


marcelveldt:
I was already working on that

PR #819, opened 2025-01-13
fix: Prevent racing condition to cause issue

felipetoffolo1:
I believe that a racing condition is preventing preventing activePlayer to be set by the URL param.

That happens when the watch is executed before `api.players` is populated.

How it is today, the watch won't be executed again when `api.players` change

FYI: The problem does not happen locally (probably because the loading order will be a little different) , so it was not possible to be 100% sure that this fixes the issue


src/layouts/default/Default.vue
```diff
@@ -36,15 +36,15 @@ const framelessState = ref(false);
 
 const route = useRoute();
 watch(
-  () => route.query.player,
-  (newActivePlayer) => {
+  [() => route.query.player, () => api.players], // watch both the route and api.players
+  ([newActivePlayer, players]) => {
```
marcelveldt: players is a continuously changing, nested large object. Does this come with a performance penalty ?
Because in that case we could also set some boolean value in the store after "fetchState" in the api code.

felipetoffolo1: It was triggering mostly on route changes.
But you are right that it is not necessary.

I'm just checking for the "object length" so that it would re-trigger when moving from populated to non-populated only

marcelveldt (approved):

PR #821, opened 2025-01-13
chore: Hide double scroll

felipetoffolo1:
Vuetify forces `overflow-y` in the HTML element.

Not a popular choice

https://github.com/vuetifyjs/vuetify/issues/16994
https://github.com/vuetifyjs/vuetify/issues/1197

This overrides it back to `hidden`

apparently they will remove it soon


marcelveldt (approved):
Nice find!

PR #822, opened 2025-01-15
Feat: Add simple DSP indicator to Streamdetails to display the current DSP state

maximmaxim345:
This PR adds support for DSPDetails, introduced with https://github.com/music-assistant/server/pull/1875.

# Overview

This is currently a simple indicator that displays two states:
- When a DSP is disabled due to grouping in an incompatible `PlayerProvider`
- When a DSP is enabled and functioning

This indicator primarily helps users understand when a DSP is disabled due to grouping (thus avoiding confusion).
It will be replaced with a fully featured display of the DSP pipeline once I finish that (could still take some time).

# Screenshots

Screenshot From 2025-01-13 19-21-07

Screenshot From 2025-01-13 19-20-53
(text is capitalized in this PR, but not in the screenshots)


marcelveldt (approved):
Good as a first start - next step is redesigning that whole dialog I think.
Lets chat about that on discord

PR #823, opened 2025-01-17
Fix: Remove links to DSP Settings for PlayerGroups

maximmaxim345:
Group players only use the DSP of the containing players.
Therefore this PR hides all links to the DSP settings on `PlayerGroup`s.

### DIFF
--- a/src/helpers/player_menu_items.ts
+++ b/src/helpers/player_menu_items.ts
@@ -179,16 +179,18 @@ export const getPlayerMenuItems = (
   });
 
   // add shortcut to dsp settings
-  menuItems.push({
-    label: "open_dsp_settings",
-    labelArgs: [],
-    action: () => {
-      store.showFullscreenPlayer = false;
-      store.showPlayersMenu = false;
-      router.push(`/settings/editplayer/${player.player_id}/dsp`);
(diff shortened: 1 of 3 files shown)


src/helpers/player_menu_items.ts
```diff
@@ -179,16 +179,18 @@ export const getPlayerMenuItems = (
...
-    label: "open_dsp_settings",
-    labelArgs: [],
-    action: () => {
-      store.showFullscreenPlayer = false;
-      store.showPlayersMenu = false;
-      router.push(`/settings/editplayer/${player.player_id}/dsp`);
-    },
-    icon: "mdi-equalizer",
-  });
+  if (!player.provider.startsWith("player_group")) {
```
marcelveldt: we also have playertype, to catch all native playergroups.
For instance you also have cast groups, which you miss here by only checking playergroups.
Or do you only want to catch syncgroups here and rule out all dedicated group players ?

maximmaxim345: I actually missed that we have the `PlayerType` enum!
I adjusted this PR to use that instead.

marcelveldt (approved):

PR #824, opened 2025-01-17
Fix: Allow opening of DSP Settings on unavailable players

maximmaxim345:
We have no reason to block the user from opening the DSP Settings on unavailable Players.

### DIFF
--- a/src/views/settings/Players.vue
+++ b/src/views/settings/Players.vue
@@ -169,7 +169,6 @@ const onMenu = function (evt: Event, playerConfig: PlayerConfig) {
         editPlayerDsp(playerConfig.player_id);
       },
       icon: "mdi-equalizer",
-      disabled: !api.players[playerConfig.player_id]?.available,
       hide: api.players[playerConfig.player_id]?.type === PlayerType.GROUP,
     },
     {


marcelveldt:
Are you sure the code can handle this ? Last time I checked it couldn't hence the extra guard.
Does it make sense to update dsp settings for a player that is offline ?

maximmaxim345:
> Are you sure the code can handle this ? Last time I checked it couldn't hence the extra guard.

The backend endpoint calls [this function](https://github.com/music-assistant/server/blob/dev/music_assistant%2Fcontrollers%2Fplayers.py#L1170) to reload the DSP. The function has an early return if it can't find the player, so yes.

> Does it make sense to update dsp settings for a player that is offline ?

To update: Not that I think of.
To view: Yes! For example to export/save an EQ of an now unavailable device.

Additionally the user could already access the settings by:
- opening them through the player settings
- opening them through the player/speaker selector.

marcelveldt (dismissed):

maximmaxim345:
Fixed merge conflicts.

marcelveldt (approved):

PR #825, opened 2025-01-19
Implemented paging in browse feature of music providers

robsonke:
See #1888 in the server repo, I assume that needs to be merged first.

### DIFF
--- a/src/plugins/api/index.ts
+++ b/src/plugins/api/index.ts
@@ -256,16 +256,19 @@ export class MusicAssistantApi {
       album_types,
     });
   }
+
   public getLibraryTracksCount(
     favorite_only: boolean = false,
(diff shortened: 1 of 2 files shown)


robsonke:
Small note, sorting doesn't work yet, but that's case too without paging. Maybe I'll add that later.

marcelveldt:
Sorry, as discussed in discord its not as easy as this because you will loose sorting this way.
The beauty of browse currently is that for rather small count listings, the sorting and filtering can be done entirely in the frontend. Alternative could be to do all sorting, filtering and paging in the backend but that is rather complex/impossible due to all differences in the providers. It needs at least a thorough architecture discussion.

I'm going to deny the PR's for that reason, sorry for the work you've put in.

PR #852, opened 2025-02-05
feat: make players list expandable in the Audio Pipeline

maximmaxim345:
To reduce clutter when playing to many devices, the audio pipeline now collapses multiple players.
Pressing on a collapsed player will expand the list to show all group members, as was done previously.
Pressing again on the expanded list will collapse it again, showing only the first player.

# Screenshots

Screenshot From 2025-02-03 19-06-04

Screenshot From 2025-02-03 19-05-57


marcelveldt (approved):
Nice!

PR #885, opened 2025-02-26
Feat: improved Notification support

maximmaxim345:
# Overview
This PR reworks part of the Media Controls to solve a few bugs, and add a couple new features.
This includes:
- Play an inaudible sound to ensure that the notification will show up with all browsers. (including firefox on desktop)
- Show the notification even if the player is initially paused.
- Add support for displaying cover art in the notification.
- Show the current position and total track length.
- Support seeking from withing the notification.

This PR also refactors some code for the future built in player provider.

# MacOS/IOS
It seams like Apple hides track forwad/backward buttons if seek forward/backward is registered.
Since track forward/backward is more important, seek forward/backward will only be registered on non Apple platforms.
Regular seeking is still enabled on all platforms.

# Inaudible Tone
A almost silent tone with an inaudible frequency will play for 1 second on startup.
After that no sound will be played, even if the browser says that Music Assistant plays audio.
From my testing, the notification should now work with all browsers supporting the `MediaSession` API.

# Screenshots
Screenshot_20250216_184531_Brave
Screenshot From 2025-02-18 11-23-18
Screenshot From 2025-02-17 14-11-03


marcelveldt (approved):
Impressive work again @maximmaxim345 !

PR #886, opened 2025-02-26
Feat: Move the output limiter option from DSP to Player Settings

maximmaxim345:
Companion PR for https://github.com/music-assistant/server/pull/1981.


marcelveldt (approved):

OzGav:
Will it look something like this in the end? 

image

maximmaxim345:
> Will it look something like this in the end?

The options name will be "Enable limiting to prevent clipping".
But I like your Idea of moving DSP settings to "Audio" and the "DSP is enabled".
So, yes!

PR #887, opened 2025-02-27
Fix: Select all in search view

maximmaxim345:
This PR fixes pressing CTRL-A in Search views.

Additionally this adds a button to allow selecting everything without a keyboard:
image

### DIFF
--- a/src/components/ItemsListing.vue
+++ b/src/components/ItemsListing.vue
@@ -788,8 +788,7 @@ const keyListener = function (e: KeyboardEvent) {
   if (loading.value) return;
   if (e.key === "a" && (e.ctrlKey || e.metaKey)) {
     e.preventDefault();
-    selectedItems.value = pagedItems.value;
-    showCheckboxes.value = true;
+    selectAll();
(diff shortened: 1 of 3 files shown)


marcelveldt (approved):
Nice QOL improvement!

OzGav:
I am just testing this and looking at the report here https://github.com/orgs/music-assistant/projects/2/views/1?pane=issue&itemId=78197266

In the report it is reported that when viewing an artist or an album if you go to the TRACKS section of those views and press the multi-select then you can't use CTRL-A to select all of the tracks. This still doesn't work.

If the solution is rather than using CTRL-A that now the user will use the new select all icon then that is fine and I will add that info to the docs. However, I further noticed when using that icon that you only get 50 items at a time (yep understand the paged nature of the views) so just wanted to confirm the intent is that the user will press the button and then scroll down through the list until the bottom is reached to actually select all the tracks?

edit: Also I think related is this https://github.com/music-assistant/support/issues/2840 I just tried it again and the behaviour is as described in the report.

PR #888, opened 2025-02-27
Fix: Improved colors in fullscreen player

maximmaxim345:
The colors in the Fullscreen player where bothering me a bit so this PR makes a few adjustments:
- Watch for changes if the User Switches Dark mode on or off.
- Keep colors between the text and the slider consistent.
- Ensure that the text/slider always has a readable contrast ration for better accessibility (while keeping the color as close as possible to the album cover).

Currently, selection between black or white text is done by checking whatever gives a higher contrast.
But this could be changed, since the main contribution of this PR is verifying/adjusting the background colors readability/contrast.

# Screenshots

## Before
before 1

## After
after 1

## Before
before 2

## After
after 2

## Before
before 3

## After
after 3

### DIFF
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -303,35 +303,6 @@ export function findDarkColor(colors: RGBColor[]): string {
   return mostPleasantColor;
 }
 
-export function darkenBrightColors(
-  color: string,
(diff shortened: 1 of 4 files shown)


marcelveldt (approved):
Nice!

PR #898, opened 2025-03-03
Fix: More search fixes

maximmaxim345:
Fixes a few more bugs around the search/select-all I missed with https://github.com/music-assistant/frontend/pull/887.
Like:
- enables type to search for tracks in an album
- clear search when closing the search box
- pressing escape now closes the search box
- pass text handling to the text box, allowing ctrl a, ctrl c, and similar
- Make select all really select everything by loading all items first, asking for confirmation for large selections over 250 items

### DIFF
--- a/src/components/ItemsListing.vue
+++ b/src/components/ItemsListing.vue
@@ -178,6 +178,7 @@ import {
   nextTick,
   onMounted,
   watch,
+  watchEffect,
 } from "vue";
 import {
   MediaType,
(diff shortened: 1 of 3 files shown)


OzGav:
Thanks Maxim. Can you confirm the intent now will be to use the select all button only to select all the items or will CTRL-A still work if the search box isn't the focus?

maximmaxim345:
> Thanks Maxim. Can you confirm the intent now will be to use the select all button only to select all the items or will CTRL-A still work if the search box isn't the focus?

CTRL-A will select all items when the search box is not selected or hidden.
If searching, it will select the searched for text.

So both, the button and CTRL-A can be used.

You also asked if select all should select only all loaded items or everything.
I will look into that this or next week.

maximmaxim345:
Select all now loads all items first.
For large item counts, like selecting every track in the library, this will ask for a confirmation first:
image

marcelveldt (approved):
LGTM, thanks @maximmaxim345 !

PR #901, opened 2025-03-07
Allow RTSP URLs for Builtin Provider

jdelaney72:
The server already handles RTSP audio streams in the Builtin Provider but the UI does not accept RTSP URLs.


marcelveldt (approved):
Thanks @jdelaney72

PR #902, opened 2025-03-08
Feat: Built-in Web Player Provider

maximmaxim345:
For a full description, see https://github.com/music-assistant/server/pull/2009.

Depends on https://github.com/music-assistant/server/pull/2009

Translation support will come with a future PR once the final text is decided.

### DIFF
--- a/src/components/PlayerCard.vue
+++ b/src/components/PlayerCard.vue
@@ -1,5 +1,6 @@
 <template>
   <v-card
+    v-if="player !== 'web'"
     flat
     class="panel-item"
     :class="{
@@ -151,9 +152,21 @@
               player.state == PlayerState.PLAYING ? 'mdi-pause' : 'mdi-play'
             "
         /></Button>
+
+        <!-- power button for a local player -->
+        <Button
+          v-if="player.player_id == webPlayer.player_id"
+          variant="icon"
+          class="player-command-btn"
+          @click.stop="webPlayer.disable()"
+          ><v-icon
+            :size="getBreakpointValue({ breakpoint: 'phone' }) ? '30' : '32'"
+            >mdi-power</v-icon
+          ></Button
+        >
         <!-- power button -->
         <Button
-          v-if="player.power_control != PLAYER_CONTROL_NONE"
+          v-else-if="player.power_control != PLAYER_CONTROL_NONE"
           variant="icon"
           class="player-command-btn"
           @click.stop="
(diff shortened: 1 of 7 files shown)


src/components/PlayerCard.vue
```diff
@@ -151,9 +152,21 @@
...
+        <Button
+          v-if="player.player_id == webPlayer.player_id"
+          variant="icon"
+          class="player-command-btn"
+          @click.stop="webPlayer.disable()"
+          ><v-icon
+            :size="getBreakpointValue({ breakpoint: 'phone' }) ? '30' : '32'"
+            >mdi-power</v-icon
+          ></Button
+        >
```
marcelveldt, line 166: what is the exact goal of a power button to disable the player ?

src/components/PlayerCard.vue
```diff
@@ -191,6 +204,68 @@
       :allow-wheel="false"
     />
   </v-card>
+  <!-- pseudo player to register the web browser for playback -->
```
marcelveldt, line 207: I'm not entirely following why its needed to duplicate the player card.
If you need a place to enable/disable/configure the builtin player, I think we should add config to the config views, there's a dedicated section there for the frontend itself.

At some point I want to overhaul the config pages so we can make it better/prettier.

src/layouts/default/PlayerSelect.vue
```diff
@@ -53,6 +53,11 @@
           "
           @click="playerClicked(player)"
         />
+        <PlayerCard
+          v-if="!webPlayer.player_id"
+          player="web"
+          @click="localPlayerEnabled"
+        />
```
marcelveldt, line 60: Some comment as before, I think the player should be treated as any other player and we just add some filter to only show it on its own device.

src/components/PlayerCard.vue
maximmaxim345: Having this duplicate player card has 3 main reasons I found while testing:
- If the player was not yet enabled, it does not show up in the players list. So it does not have a player_id.
- If we alternatively enable the player automatically by default, each open browser would create a new player/queue. But that's not really a problem since, like you said, we could disable it by default (and/or auto discard the player/queue after some time). The main problem resulting of this is twofold:
  - The notification would no longer show the selected player, so only the local browser can be controlled.
  - Having two browser tabs with an enabled player (which would easily be the case if it's turned on automatically) would result in some chaos. I want to look at this later, but if it needs to be explicitly enabled in two tabs, the chance of having problems for now is a lot smaller.

Also the two tabs problem also happens with SnapWeb from my testing.

src/components/PlayerCard.vue
maximmaxim345: The main use case would be to get the other notification mode back.

But thinking again another solution would be implement native power controls in the provider. And then watch for a power of event in `web_player.ts` or `BuiltinPlayer.vue`.
But I don't think that approach would work with SnapWeb.

marcelveldt: ah! now I see.
Why not just see those 2 separate from eachother ?

In other words:

1) You have a (optional ?) feature that enabled the notifications of media playback.
2) You have an optional builtin player

Both can work independent from eachother.

src/components/PlayerCard.vue
marcelveldt: Ah OK, I know what you mean. In my early implementation back then I fixed that by storing the player id in the local storage. And then always check if the player is already registered before re-registering (to prevent the issue with multiple browser tabs).

On the other hand... why not just register each session ? so even allow playback into multiple browser tabs to make the implementation easier. And detach the browser notifications from the internal player. 

Ah, issue is perhaps that the notification stuff cant send silence while we're actually playing music with the builtin player ?

src/components/PlayerCard.vue
marcelveldt: YES, exactly 
As discussed on discord, we'll merge this now as-is so we can collect some early feedback and then iterate on it later in the beta cycle.

marcelveldt (approved):
Thanks @maximmaxim345 !

There a few small discussion items left but as discussed on discord, we'll merge this now as-is so we can collect some early feedback and then iterate on it later in the beta cycle.

PR #903, opened 2025-03-09
fix: Missing options when adding a new PEQ band

maximmaxim345:
Pressing "Add band" would previously show empty options instead of those of the newly added band.

Should probably also be back-ported to stable.

### DIFF
--- a/src/components/dsp/DSPParametricEQ.vue
+++ b/src/components/dsp/DSPParametricEQ.vue
@@ -121,7 +121,7 @@
   </v-container>
 </template>
 <script setup lang="ts">
-import { ref, onMounted, watch, computed } from "vue";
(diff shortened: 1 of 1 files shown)


marcelveldt:
> Should probably also be back-ported to stable.

That will be difficult now that the frontend already contains new features so I'd rather not do that unless really needed.
Otherwise we could consider releasing a version now before merging the web player.

maximmaxim345:
> That will be difficult now that the frontend already contains new features so I'd rather not do that unless really needed. Otherwise we could consider releasing a version now before merging the web player.

Ahh, I see. No problem!
This is pretty much just a visual bug, so it's not that important.

marcelveldt (approved):

PR #916, opened 2025-03-12
Various minor fixes

OzGav:
Images have white dots in the corners so new versions are supplied

Whilst looking at the above the URL was noticed to be pointing to the old custom_component repo

Also noticed a capitalisation issue in the Frontend


marcelveldt (approved):
Thanks @OzGav !

PR #923, opened 2025-03-17
Fix: Show the web player on all tabs of MA

maximmaxim345:
This is a quick workaround to make the "This Device" player show up on all tabs when hidden (so with default setting).
Will do a proper solution soon, but this should be enough for testing.


marcelveldt (approved):
Understood. We somehow need a way to mark players as "hide to others but not on this machine".

PR #961, opened 2025-04-11
UI Rework for Authentication Dialog: Added Close Button, Timeout, and…

MickLesk:
### Description:  
This PR addresses the issue outlined in [Issue #3836](https://github.com/music-assistant/support/issues/3836), where the authentication dialog in the configuration was blocking user interaction after a failed authentication attempt. The initial motivation was to improve the user interface by adding a close button ("X") to the dialog, which led to a complete rework of the authentication UI.

**Key Changes:**
- Added a close button ("X") to the authentication dialog to allow users to manually cancel the authentication process.
- Implemented a timeout for authentication failure, automatically closing the dialog after a set period.
- Enhanced UI to support dark mode and light mode seamlessly.
- Added feedback mechanisms using snackbars for authentication success or failure.

This update ensures better user control and improves the visual experience of the authentication process.

Some Impressions:
image

Timeout (60 s):
image

Manual Abort (with [x]) 
image

--

Dark + i18n:
image

Manual Abort (with [x]) 
image

Timeout (60s)
image


OzGav:
It is appropriate to amend en.json but the others are generated by lokalise. After the PR is merged and en.json is pulled into Loklise then you can do the translations there.

MickLesk:
Ah, all right. That was the only thing I wasn't quite sure about. Should I remove it myself?

OzGav:
I would say yes. Just remove it from your PR

src/plugins/api/index.ts
```diff
@@ -1454,33 +1454,47 @@ export class MusicAssistantApi {
...
+    const isAuthAborted =
+      "error_code" in msg &&
+      typeof msg.details === "string" &&
+      msg.details === "auth_aborted";
+
+    const isAuthTimeout =
+      "error_code" in msg &&
+      typeof msg.details === "string" &&
+      msg.details === "auth_timeout";
+
```
marcelveldt, line 1468: I don't think this may/should live here as its an implementation detail of the auth flow within config.
Also it will at some point cause confusion if our api is going to have auth on its own.
So I would move handling the auth error in the config view and maybe create a dedicated component which we can use for AddProvider and EditProviderConfig.

src/plugins/api/index.ts
```diff
@@ -1454,33 +1454,47 @@ export class MusicAssistantApi {
...
+      // Handle manual abortion by the user
+      else if (isAuthAborted) {
+        console.debug("[resultMessage] User aborted auth flow manually.");
+        showSnackbar($t("settings.auth_aborted"), "error"); // Show the correct message
+      }
+      else {
+        console.error("[resultMessage]", msg);
+        // Show generic error message
+        showSnackbar(msg.details || msg.error_code, "error");
+      }
```
marcelveldt, line 1486: same comment, this may not live here.

src/plugins/api/index.ts
```diff
@@ -1454,33 +1454,47 @@ export class MusicAssistantApi {
...
-      // always handle error (as we may be missing a resolve promise for this command)
       msg = msg as ErrorResultMessage;
-      console.error("[resultMessage]", msg);
-      store.activeAlert = {
-        type: AlertType.ERROR,
-        message: msg.details || msg.error_code,
-        persistent: false,
-      };
-    } else if (DEBUG) {
-      console.log("[resultMessage]", msg);
```
marcelveldt, line 1470: Suggestion to make the error bubble up towards the consumer is an extra parameter to set that we want the exception to bubble up and then keep the fallback catch all exception handler with the snackbar?

src/plugins/api/index.ts
```diff
@@ -1491,14 +1505,28 @@ export class MusicAssistantApi {
...
+      } else {
+        resultPromise.reject("auth_aborted");
+      }
     } else {
       msg = msg as SuccessResultMessage;
       resultPromise.resolve(msg.result);
     }
   }
 
+  public abortAllCommands(reason = "auth_aborted") {
```
marcelveldt, line 1521: Can you tell me why this part was needed ?
Because aborting all commands sounds a bit weird when just the auth failed

MickLesk: If you click the [x] or the timeout is arrived. Both on same function. Kill the Spinner & auth

OzGav:
@MickLesk any progress on this?

PR #963, opened 2025-04-13
de.json aktualisieren

ludgerbeckmann:


OzGav:
Sorry I can see you have done a lot of work here but all translations must be done on Lokalise https://www.music-assistant.io/help/lokalise/ Just copy and paste it into there and it will get pulled down in the next cycle.

PR #973, opened 2025-04-20
Add translation for builtin_playlists

OzGav:


src/translations/en.json
```diff
@@ -22,6 +22,7 @@
     "artist_toptracks": "Top tracks",
     "artists": "Artists",
     "browse": "Browse",
+    "builtin_playlists": "Playlist Settings",
```
marcelveldt: Are you sure this is the correct translation ? Shouldn't this be "Builtin playlists" ?

OzGav: I was trying to align with the other headings - Generic Settings, Advanced Settings, AirPlay Specific Settings but I do note we aren't that standardised

image

Maybe `Builtin Playlist Settings`?

marcelveldt (approved):

PR #1061, opened 2025-07-28
Add interactive breadcrumb navigation to Browse section

stvncode:
## Interactive Breadcrumb Navigation for Browse Section

Adds clickable breadcrumb navigation to the Browse section, allowing users to jump directly to any path level instead of navigating back one step at a time.

### Features
- Click any breadcrumb segment to navigate to that specific path
- Last segment (current path) is not clickable
- Visual feedback with underlines and hover effects

### DIFF
--- a/src/components/ItemsListing.vue
+++ b/src/components/ItemsListing.vue
@@ -9,7 +9,11 @@
       color="transparent"
       :menu-items="menuItems"
       @title-clicked="toggleExpand"
-    />
+    >
+      <template #title>
+        <slot name="title">{{ title }}</slot>
+      </template>
+    </Toolbar>
 
     <v-divider />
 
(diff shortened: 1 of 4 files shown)


marcelveldt (dismissed):
Nice QoL improvement! Thanks

marcelveldt (approved):

src/components/ItemsListing.vue
```diff
@@ -279,6 +284,8 @@ const props = withDefaults(defineProps<Props>(), {
   path: undefined,
   icon: undefined,
   restoreState: false,
+  onTitleClick: undefined,
+  customTitleRenderer: undefined,
```
bramkragten, line 288: Are these used?

src/components/Toolbar.vue
```diff
@@ -10,16 +10,18 @@
...
-        <v-badge
-          v-if="count && getBreakpointValue('bp4')"
-          color="grey"
-          :content="count"
-          inline
-        />
-      </button>
+    <template #title>
+      <slot name="title">
+        <button v-if="title" @click="emit('titleClicked')">
```
bramkragten, line 15: Is this event used?

src/components/ItemsListing.vue
stvncode: customTitleRenderer can be removed yes

src/components/Toolbar.vue
stvncode: Yes for ItemsListing.vue, we have a custom title render

PR #1073, opened 2025-08-07
Fix: Prevent player cards from collapsing in slider

FabionRabushja:
Fix for the issue: https://github.com/music-assistant/support/issues/3854 

- Fixed player card width to maintain consistent 305px size
- Adjusted swiper-slide width to auto for proper card spacing
- Removed calculated swiper slides-per-view to auto, cause forcing it to a calculated value made player cards collapse


marcelveldt (approved):
Thanks, @FabionRabushja !

PR #1074, opened 2025-08-10
DSP Presets

math625f:
This PR adds the needed features to the EditPlayerDSP page required for saving, loading, and deleting DSP presets.
It requires the PR to the server repository adding support for DSP presets: https://github.com/music-assistant/server/pull/2309

### DIFF
--- a/src/views/settings/EditPlayerDsp.vue
+++ b/src/views/settings/EditPlayerDsp.vue
@@ -10,7 +10,6 @@
       <v-toolbar-title>{{
         $t("settings.dsp.configure_on", { name: playerName })
       }}</v-toolbar-title>
-      <v-spacer />
       <v-menu offset-y transition="slide-y-transition">
         <template v-slot:activator="{ props }">
           <v-btn
@@ -19,7 +18,9 @@
             :class="getButtonClass()"
             prepend-icon="mdi-tray-arrow-down"
           >
-            {{ $t("settings.dsp.presets.load") }}
+            <span class="d-none d-md-inline">
+              {{ $t("settings.dsp.presets.load") }}
+            </span>
           </v-btn>
         </template>
         <v-list>
@@ -29,8 +30,8 @@
             </v-list-item-title>
           </v-list-item>
           <v-list-item
-            v-else
             v-for="preset in dspPresets"
+            v-else
             :key="preset.preset_id"
             @click="loadPreset(preset)"
           >
@@ -51,7 +52,9 @@
         prepend-icon="mdi-content-save"
         @click="showSavePresetDialog = true"
       >
-        {{ $t("settings.dsp.presets.save") }}
+        <span class="d-none d-md-inline">
+          {{ $t("settings.dsp.presets.save") }}
+        </span>
       </v-btn>
     </v-toolbar>
 
@@ -193,8 +196,8 @@
           </v-btn>
           <v-btn
             color="primary"
-            @click="savePreset"
             :disabled="!newPresetName.trim()"
+            @click="savePreset"
           >
             {{ $t("settings.dsp.presets.save") }}
           </v-btn>
--- a/src/views/settings/EditPlayerDsp.vue
+++ b/src/views/settings/EditPlayerDsp.vue
@@ -11,13 +11,9 @@
         $t("settings.dsp.configure_on", { name: playerName })
       }}</v-toolbar-title>
       <v-menu offset-y transition="slide-y-transition">
-        <template v-slot:activator="{ props }">
-          <v-btn
-            v-bind="props"
-            class="mr-4"
-            :class="getButtonClass()"
-            prepend-icon="mdi-tray-arrow-down"
-          >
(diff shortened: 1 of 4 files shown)


math625f:
Loading a preset will fully overwrite the current DSP config, so perhaps some kind of confirmation is needed? Should be pretty simple to add if desired.

marcelveldt:
@maximmaxim345 any opinions about the overwriting ?

maximmaxim345:
@math625f Thanks for implementing DSP Presets! I will definitely be using them myself (I actually had a half-finished implementation of them but never got around to finishing it)!

Having a confirmation dialog is a good idea, but I think if we add it, it should only show up if the settings would really be lost.
For now, we can just overwrite them.


One idea for how this could be done is to move `preset_id` from `DSPConfigPreset` to `DSPConfig`. With that, we would know if the current player's DSP settings are already saved or not, and it would also allow us to know if the settings have changed since being saved. So the confirmation would only be shown if the settings aren't already saved.

Knowing the `preset_id` would also make it possible to update a saved preset. A problem I can think of with this implementation:
1.  Save a preset with the name "Kitchen Speaker."
2.  Tweak the EQ.
3.  Save another preset with the name "Kitchen Speaker."
4.  Now you have two "Kitchen Speaker" presets, and you don't know which one is the old one (which might even be in random order due to the random UUID key).

src/views/settings/EditPlayerDsp.vue
```diff
@@ -10,6 +10,49 @@
       <v-toolbar-title>{{
         $t("settings.dsp.configure_on", { name: playerName })
       }}</v-toolbar-title>
+      <v-spacer />
```
maximmaxim345:
```suggestion
```

The spacer cuts of the device name on smaller screens.

src/views/settings/EditPlayerDsp.vue
```diff
@@ -10,6 +10,49 @@
...
+      <v-menu offset-y transition="slide-y-transition">
+        <template v-slot:activator="{ props }">
```
maximmaxim345: Fix the lint warning `Variable 'props' is already declared in the upper scope`

src/views/settings/EditPlayerDsp.vue
```diff
@@ -10,6 +10,49 @@
...
+        </template>
+        <v-list>
+          <v-list-item v-if="dspPresets.length === 0">
+            <v-list-item-title>
+              {{ $t("settings.dsp.presets.empty_warning") }}
+            </v-list-item-title>
+          </v-list-item>
+          <v-list-item
+            v-else
+            v-for="preset in dspPresets"
```
maximmaxim345:
```suggestion
            v-for="preset in dspPresets"
            v-else
```

src/views/settings/EditPlayerDsp.vue
```diff
@@ -131,6 +174,34 @@
...
+        </v-card-text>
+        <v-card-actions>
+          <v-spacer />
+          <v-btn @click="showSavePresetDialog = false">
+            {{ $t("cancel") }}
+          </v-btn>
+          <v-btn
+            color="primary"
+            @click="savePreset"
+            :disabled="!newPresetName.trim()"
```
maximmaxim345:
```suggestion
            :disabled="!newPresetName.trim()"
            @click="savePreset"
```

src/views/settings/EditPlayerDsp.vue
```diff
@@ -10,6 +10,49 @@
...
+      <v-spacer />
+      <v-menu offset-y transition="slide-y-transition">
+        <template v-slot:activator="{ props }">
+          <v-btn
+            v-bind="props"
+            class="mr-4"
+            :class="getButtonClass()"
+            prepend-icon="mdi-tray-arrow-down"
+          >
+            {{ $t("settings.dsp.presets.load") }}
```
maximmaxim345:
```suggestion
            <span class="d-none d-md-inline">
              {{ $t("settings.dsp.presets.load") }}
            </span>
```

Show just the icon for smaller screens like phones.

src/views/settings/EditPlayerDsp.vue
```diff
@@ -10,6 +10,49 @@
...
+            </template>
+          </v-list-item>
+        </v-list>
+      </v-menu>
+      <v-btn
+        :class="getButtonClass()"
+        prepend-icon="mdi-content-save"
+        @click="showSavePresetDialog = true"
+      >
+        {{ $t("settings.dsp.presets.save") }}
```
maximmaxim345:
```suggestion
        <span class="d-none d-md-inline">
          {{ $t("settings.dsp.presets.save") }}
        </span>
```
Show just the icon for smaller screens like phones.

math625f:
I ended up changing the prepend icons to some manual ones, because I couldn't figure out how to center the prepend icons.

maximmaxim345 (approved):
LGTM, Thanks @math625f !

PR #1090, opened 2025-08-20
Implement chapter time over full audiobook duration

tsipinakis:
Why? A picture is a thousand words:

2025-08-20T18-32-43

2025-08-20T18-32-17

### DIFF
--- a/src/components/navigation/Navigation.vue
+++ b/src/components/navigation/Navigation.vue
@@ -5,6 +5,7 @@
         src="@/assets/icon.svg"
         alt="Music Assistant"
         class="sidebar-header-logo"
+        @click="toggleCollapsed"
       />
       <div v-if="!collapsed" class="sidebar-header-title">Music Assistant</div>
     </div>


tsipinakis:
I fixed the possible undefined errors during build, note that this is the first time I ever touch TS and Vue, so let me know if there are more standard ways to do what I did :sweat_smile:

tsipinakis:
Ready for another lint run/review

src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -88,36 +88,75 @@ const isDragging = ref(false);
...
+    localStorage.getItem("frontend.settings.audiobook_chapter_time") == "true"
+  );
+});
+
 // computed properties
+const curChapter = computed(() => {
+  if (store.curQueueItem?.media_item?.metadata?.chapters) {
+    return store.curQueueItem.media_item.metadata.chapters.find((chapter) => {
+      return (
+        chapter.start! < store.activePlayerQueue?.elapsed_time! &&
```
stvncode: Using ! (non-null assertion) on optional chain expressions is unsafe and can cause runtime crashes. You need to check before if the values or here

src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -88,36 +88,75 @@ const isDragging = ref(false);
...
+  }
+  return null;
+});
+
+const playerCurQueueItemDuration = computed(() => {
+  if (
+    chapterTime.value &&
+    store.curQueueItem?.media_item?.media_type == MediaType.AUDIOBOOK
+  ) {
+    return curChapter.value?.end! - curChapter.value?.start!;
```
stvncode: Same here. This will crash if curChapter.value is null or if end/start are undefined.

src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -88,36 +88,75 @@ const isDragging = ref(false);
...
+    tempTime.value = curTimeValue.value;
+    return curTimeValue.value;
+  }
+
+  if (store.activePlayerQueue) {
+    if (
+      chapterTime.value &&
+      store.curQueueItem?.media_item?.media_type == MediaType.AUDIOBOOK
+    ) {
+      return store.activePlayerQueue?.elapsed_time - curChapter.value?.start!;
```
stvncode: Same problem here, unsafe non-null assertion.

src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -141,10 +180,9 @@ const startDragging = function () {
 const stopDragging = () => {
   isDragging.value = false;
   if (!isDragging.value && store.activePlayer) {
-    api.playerCommandSeek(
-      store.activePlayer.player_id,
-      Math.round(tempTime.value),
-    );
+    var seekTime = tempTime.value;
+    if (chapterTime.value) seekTime = curChapter.value?.start! + seekTime;
```
stvncode: This will crash if curChapter.value is null or start is undefined.

src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -88,36 +88,75 @@ const isDragging = ref(false);
...
-    tempTime.value = curTimeValue.value;
-    return curTimeValue.value;
-  }
-  if (store.activePlayerQueue) return store.activePlayerQueue.elapsed_time;
-  return 0;
-});
 
 const chapterTicks = computed(() => {
   const ticks: Record<number, string> = {};
+  if (chapterTime.value) return [];
```
stvncode: Maybe I'm wrong but this seems backwards. When chapter time mode is enabled, you're hiding all chapter ticks. Shouldn't chapter ticks be shown when not in chapter time mode?

src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -88,36 +88,75 @@ const isDragging = ref(false);
 const curTimeValue = ref(0);
 const tempTime = ref(0);
 
+const chapterTime = computed(() => {
+  return (
```
stvncode: Small syntax tips here, if you do no logic in a function, you can remove the return. Something like this:
```
const chapterTime = computed(() => 
localStorage.getItem("frontend.settings.audiobook_chapter_time") == "true"
  );
```

stvncode (commented):
Thanks for the great enhancement! Just some small comments on non-null assertion

src/layouts/default/PlayerOSD/PlayerTimeline.vue
tsipinakis: Whoops apologies, I had already fixed that, but only pushed to my personal branch not the one linked to this MR.

src/layouts/default/PlayerOSD/PlayerTimeline.vue
tsipinakis: Exactly, when you're using chapterTime (if is true) then there are no chapter ticks because the timeline starts from the beginning to the end of a chapter. If you're not using chapterTime (if is false) then ticks are shown.

OzGav:
Having just started to play with Audiobooks there are a number of considerations here. For audiobooks with not so many chapters having the full timeline means you can click on the chapter mark to jump to that chapter. Obviously in the example above this is not that relevant as there are so many chapters so maybe not something to worry about. However, while on this subject one thing the UI lacks is an indication of which chapter is currently playing. Is there anything we can do there? 

P.S. Check the lint probs...

tsipinakis:
Whoops indeed I pushed to the wrong branch again, pushed to the one I use for my own build, but forgot to also push to the MR one :see_no_evil: 

> However, while on this subject one thing the UI lacks is an indication of which chapter is currently playing. Is there anything we can do there?

Indeed, though out of scope for this MR, maybe next one :)

marcelveldt:
Without showing the user that this is a chapter playing, this is going to cause unclarity. We need to figure out the UX how to display/select chapters. Silently only showing the current chapter makes the progress bar nicer maybe not perse more easier.

tsipinakis:
> Silently only showing the current chapter makes the progress bar nicer maybe not perse more easier.

At least for me, now it takes it from completely unusable and dangerous (misclick means I lose my place and jump to a random place in the book), to useful as I can seek around.

I agree about the indication though - but I'd say this can be a first step to be merged as it's an opt-in and users should be aware that they have switched it on.

I do intend to look at the second part after.

OzGav:
You can just touch a chapter in the list to jump there

tsipinakis:
I just added audiobook chapter subtitles, this should be better now.

OzGav:
Have you got some screenshots?

tsipinakis:
2025-10-15T14-29-19
2025-10-15T14-29-29

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
```diff
@@ -151,6 +151,14 @@
               {{ $t("off") }}
             </v-card-subtitle>
 
+            <!-- SUBTITLE: player powered off -->
```
marcelveldt:
```suggestion
            <!-- SUBTITLE: current (audiobook) chapter -->
```

src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -88,36 +88,63 @@ const isDragging = ref(false);
 const curTimeValue = ref(0);
 const tempTime = ref(0);
 
+const chapterTime = computed(() =>
+  localStorage.getItem("frontend.settings.audiobook_chapter_time") == "true"
```
marcelveldt: Are we certain a computed property is the right way to do this ?
Like, will it only be computed once and not over and over ?

src/views/settings/FrontendConfig.vue
```diff
@@ -131,6 +131,30 @@ onMounted(() => {
...
+      type: ConfigEntryType.BOOLEAN,
+      label: "chapter_time",
+      default_value: false,
+      required: false,
+      multi_value: false,
+      category: "audiobooks_podcasts",
+      value:
+        localStorage.getItem("frontend.settings.audiobook_chapter_time") ==
+        "true",
+    },
```
marcelveldt: duplicate?

OzGav:
@marcelveldt This subtitle idea is somewhat aligned with the concept of providing another line in the now playing view to provide extra information for the Radio Paradise provider? Should we review that for a consolidated approach?

src/layouts/default/PlayerOSD/PlayerTimeline.vue
tsipinakis: I'm not sure, have not used vue before this. I'm open to ideas.

On the other hand: Does it matter? How much overhead is a map lookup?

src/views/settings/FrontendConfig.vue
tsipinakis: Yep, rebase error.

tsipinakis:
@marcelveldt Ready for another review round

tsipinakis:
@marcelveldt @OzGav ping?

PR #1091, opened 2025-08-21
Add shortcuts functionality to navigation and context menus

FabionRabushja:
- Implemented a shortcuts menu in the BottomNavigation and DrawerNavigation components, allowing users to view and interact with shortcuts grouped by media type.
- Added functionality to handle shortcut clicks, including playing tracks and navigating to item details.
- Integrated shortcut management in the ItemContextMenu for adding/removing shortcuts.
- Updated translations to include shortcut-related terms.
- Ensured existing shortcuts are hydrated on connection.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -18,6 +18,7 @@ import { api } from "@/plugins/api";
 import { onMounted } from "vue";
 import { useTheme } from "vuetify";
 import { store } from "@/plugins/store";
+import { hydrateShortcutsImages } from "@/helpers/shortcuts";
 import { i18n } from "@/plugins/i18n";
 import router from "./plugins/router";
 import { EventType } from "./plugins/api/interfaces";
@@ -106,6 +107,8 @@ onMounted(() => {
     store.libraryRadiosCount = await api.getLibraryRadiosCount();
     store.libraryTracksCount = await api.getLibraryTracksCount();
     store.connected = true;
+    // hydrate missing images for existing shortcuts when we connect
+    await hydrateShortcutsImages();
     // enable the builtin player by default if the builtin player provider is available
     if (allowBuiltinPlayer && api.getProvider("builtin_player")) {
       webPlayer.setMode(WebPlayerMode.BUILTIN);
--- a/src/helpers/shortcuts.ts
+++ b/src/helpers/shortcuts.ts
@@ -1,0 +1,145 @@
+import { reactive } from "vue";
+import { MediaType, MediaItem } from "@/plugins/api/interfaces";
+import api from "@/plugins/api";
+
+export interface Shortcut extends MediaItem {
+  id: string;
+  name: string;
+  icon: string;
+  itemType: MediaType;
+  itemId: string;
+  provider: string;
+  uri: string;
+}
+
+// Reactive shortcuts store
+export const shortcuts = reactive<Shortcut[]>(
+  JSON.parse(localStorage.getItem('shortcuts') || '[]')
+);
+
+// Helper function to get icon for media type
+function getIconForMediaType(type: MediaType): string {
+  switch (type) {
+    case MediaType.TRACK:
+      return 'mdi-music-note';
+    case MediaType.ALBUM:
+      return 'mdi-album';
+    case MediaType.ARTIST:
+      return 'mdi-account-music';
+    case MediaType.PLAYLIST:
+      return 'mdi-playlist-music';
+    case MediaType.RADIO:
+      return 'mdi-radio';
+    case MediaType.PODCAST:
+      return 'mdi-podcast';
+    case MediaType.AUDIOBOOK:
+      return 'mdi-book-open-variant';
+    default:
+      return 'mdi-bookmark';
+  }
+}
+
+// Helper function to save shortcuts to localStorage
+function saveShortcuts() {
+  localStorage.setItem('shortcuts', JSON.stringify(shortcuts));
+}
+
+// Add a new shortcut
+export async function addShortcut(item: MediaItem): Promise<void> {
+  // Resolve a full item from library or provider to ensure images are present
+  let fullItem: MediaItem = item;
+  try {
+    const fromLibrary = await api.getLibraryItem(item.media_type, item.item_id, item.provider);
+    if (fromLibrary) fullItem = fromLibrary as MediaItem;
+    if (!fullItem.metadata?.images || fullItem.metadata.images.length === 0) {
+      const fromProvider = await api.getItem(item.media_type, item.item_id, item.provider);
+      if (fromProvider) fullItem = fromProvider as MediaItem;
+    }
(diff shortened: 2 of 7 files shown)


src/helpers/shortcuts.ts
```diff
@@ -0,0 +1,145 @@
...
+  if (index !== -1) {
+    shortcuts.splice(index, 1);
+    saveShortcuts();
+  }
+}
+
+// Hydrate shortcuts with missing images
+export async function hydrateShortcutsImages(): Promise<void> {
+  try {
+    for (let i = 0; i < shortcuts.length; i++) {
```
bramkragten, line 99: Use `for ... of` instead to simplify the code.
```suggestion
    for (const sc of shortcuts) {
```

src/helpers/shortcuts.ts
```diff
@@ -0,0 +1,145 @@
...
+      const sc = shortcuts[i] as any;
```
bramkragten, line 100: Why is this typed as `any`?

src/layouts/default/DrawerNavigation.vue
```diff
@@ -41,13 +76,50 @@
...
+  MediaType.PLAYLIST,
+  MediaType.RADIO,
+  MediaType.PODCAST,
+  MediaType.AUDIOBOOK
+];
+
+// Group shortcuts by media type
+const getShortcutsByType = (type: MediaType) => {
+  return shortcuts.filter(shortcut => shortcut.itemType === type);
+};
```
bramkragten, line 105: I would rather go over the shortcuts once and group them all by type instead of going over the list of shortcuts for every media type:

```ts
const groupByMediaType = (
  shortcuts: Shortcut[]
): Record<MediaType, Shortcut[]> => {
  const result = {};
  for (const shortcut of shortcuts) {
    const type = shortcut.itemType;
    if (key in result) {
      result[key].push(shortcut);
    } else {
      result[key] = [shortcut];
    }
  }
  return result;
};
```

src/helpers/shortcuts.ts
```diff
@@ -0,0 +1,145 @@
...
+      const hasImages = sc?.metadata?.images && sc.metadata.images.length > 0;
+      if (hasImages) continue;
+      try {
+        let fullItem = (await api.getLibraryItem(sc.media_type || sc.itemType, sc.item_id || sc.itemId, sc.provider)) as any;
```
stvncode, line 104: As @bramkragten mentionned, you can remove the as any, typescript will infer the type. You can also remove all the as any in the file, it'll work without them.

src/layouts/default/BottomNavigation.vue
```diff
@@ -73,6 +159,71 @@ defineProps<Props>();
 const menuItems = getMenuItems();
 
 const isLibraryOpen = ref(false);
+const isShortcutsOpen = ref(false);
+
+// Define media types for grouping
```
stvncode, line 164: You are using mediaTypes in here and also in DrawerNavigation, it would be better to export it from a place and import it in both places to remove duplicate

src/helpers/shortcuts.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+
+// Check if an item is already in shortcuts
+export function isInShortcuts(mediaType: MediaType, itemId: string): boolean {
+  return shortcuts.some(s => s.id === `${mediaType}-${itemId}`);
+}
+
+// Get shortcuts filtered by media type
+export function getShortcutsByType(type: MediaType): Shortcut[] {
```
stvncode, line 137: getShortcutsByType is implemented in two places, here and in Bottom Navigation

src/layouts/default/BottomNavigation.vue
```diff
@@ -73,6 +159,71 @@ defineProps<Props>();
...
+  MediaType.ALBUM,
+  MediaType.ARTIST,
+  MediaType.PLAYLIST,
+  MediaType.RADIO,
+  MediaType.PODCAST,
+  MediaType.AUDIOBOOK
+];
+
+// Group shortcuts by media type
+const getShortcutsByType = (type: MediaType) => {
```
stvncode, line 176: This function is called multiple times in the template. You have 7 media types and if multiple shortcuts, every render, you have multiple iterations. What could be great here is to use a computed version, what do you think ?

src/App.vue
```diff
@@ -106,6 +107,8 @@ onMounted(() => {
     store.libraryRadiosCount = await api.getLibraryRadiosCount();
     store.libraryTracksCount = await api.getLibraryTracksCount();
     store.connected = true;
+    // hydrate missing images for existing shortcuts when we connect
+    await hydrateShortcutsImages();
```
stvncode, line 111: This blocks everything after it (if you have many shortcuts or slow api responses for example). What could be good to do is something like this: 
`hydrateShortcutsImages().catch(error => {
  console.warn('Failed to hydrate shortcut images:', error);
});`

src/helpers/shortcuts.ts
```diff
@@ -0,0 +1,145 @@
...
+  icon: string;
+  itemType: MediaType;
+  itemId: string;
+  provider: string;
+  uri: string;
+}
+
+// Reactive shortcuts store
+export const shortcuts = reactive<Shortcut[]>(
+  JSON.parse(localStorage.getItem('shortcuts') || '[]')
```
stvncode, line 17: Parsing localStorage without validation is risky. Could you add some validation please ?

src/layouts/default/BottomNavigation.vue
```diff
@@ -73,6 +159,71 @@ defineProps<Props>();
...
+        store.activeAlert = {
+          type: AlertType.ERROR,
+          message: "Not connected to server",
+          persistent: false
+        };
+        return;
+      }
+
+      // Play the track
+      await api.playMedia([shortcut.uri], QueueOption.PLAY);
```
stvncode, line 206: Could be nice here to add a try catch like this for example:
`try {
  await api.playMedia([shortcut.uri]);
} catch (error) {
  store.activeAlert = { type: 'error', message: 'Failed to play track' };
}`

src/helpers/shortcuts.ts
```diff
@@ -0,0 +1,145 @@
+import { reactive } from "vue";
+import { MediaType, MediaItem } from "@/plugins/api/interfaces";
+import api from "@/plugins/api";
+
+export interface Shortcut extends MediaItem {
```
stvncode, line 5: Do you need to actually extends the interface ? By doing that you'll have in Shortcut duplicate data like media_type and itemType and also for item_id and itemId. Later on you do some check like sc.media_type || sc.itemType that could be prevented. But maybe i'm wrong and I didn't catch something

stvncode:
Hi @FabionRabushja, just to check in, do you need any help to follow your work ? Thanks again for the time allowed to this 🙏

stvncode:
Closing for inactivity, don't hesitate to leave a message if you want to add shortcuts back

PR #1099, opened 2025-08-28
Podcast multi select menu played / unplayed

OzGav:
Requested here https://github.com/orgs/music-assistant/discussions/3673


stvncode:
Thanks for implementing this feature! I'm still relatively new to the codebase myself, I might be missing context or architectural decisions. A few thoughts:

- The core functionality looks solid and the logic flow makes sense
- I'm wondering if this belongs in getContextMenuItems() rather than getPlayMenuItems() since it's more of a content management action than a playback action?
- A couple of small suggestions around state detection and async handling (see other comments)

Overall this is a great addition! Let me know if any of my suggestions don't make sense or if there's context I'm missing.

src/layouts/default/ItemContextMenu.vue
```diff
@@ -448,6 +448,73 @@ export const getPlayMenuItems = async function (
...
+  // Multi-select mark as played/unplayed for podcast episodes
+  if (
+    items.length > 1 && 
+    items.every(item => 
+      item.media_type === MediaType.PODCAST_EPISODE &&
+      "fully_played" in item && 
+      "resume_position_ms" in item
+    )
+  ) {
+    const podcastEpisodes = items as PodcastEpisode[];
```
stvncode, line 460: I noticed the current logic for detecting played state might have an edge case. This considers any episode with a resume position (even 1ms) as "played". Maybe we should be more specific about what constitutes each state? Something like:
```
const hasProgress = (item: PodcastEpisode) => item.resume_position_ms > 0;
const isFullyPlayed = (item: PodcastEpisode) => item.fully_played;
const isUnplayed = (item: PodcastEpisode) => !item.fully_played && !hasProgress(item);

const allFullyPlayed = podcastEpisodes.every(isFullyPlayed);
const allUnplayed = podcastEpisodes.every(isUnplayed);
```

src/layouts/default/ItemContextMenu.vue
```diff
@@ -448,6 +448,73 @@ export const getPlayMenuItems = async function (
...
+    const allPlayed = podcastEpisodes.every(item => item.fully_played || item.resume_position_ms);
+    const allUnplayed = podcastEpisodes.every(item => !item.fully_played && !item.resume_position_ms);
+    
+    // If all items are played, show "mark unplayed" option
+    if (allPlayed) {
+      playMenuItems.push({
+        label: "mark_unplayed",
+        icon: "mdi-clock-fast",
+        action: async () => {
```
stvncode, line 475: Small optimization suggestion for the async operations. The current for loop processes episodes sequentially, which could be slow with many episodes. We could use Promise.all() to handle them in parallel like: 
```
action: async () => {
        await Promise.all(podcastEpisodes.map(async (item) => {
          await api.markItemUnPlayed(item);
          item.fully_played = false;
          item.resume_position_ms = 0;
        }));
      },
```

OzGav:
For the architectural rationale you will have to check with Marcel. When multiple items are selected a dialog appears at the centre bottom half of the UI. This is currently the only menu that is accessible when in multi-select mode to my knowledge hence this approach to place the options in there.

src/layouts/default/ItemContextMenu.vue
OzGav: Yes this is way better. I will roll this in.

src/layouts/default/ItemContextMenu.vue
OzGav: Ok also rolled in

OzGav:
@stvncode I think I have addressed the problems you identified. Let me know what you think now. Thanks for the review!

stvncode (approved):
Looks perfect to me now, thanks 🙏

PR #1105, opened 2025-09-02
Replace mods components

stvncode:
What changed:

- Replaced mods/ResponsiveIcon.vue → components/Icon.vue + composables/useIcon.ts
- Replaced mods/ListItem.vue → components/ListItem.vue + composables/useListItem.ts
- Replaced mods/Container.vue → components/Container.vue + composables/useContainer.ts

The mods components were legacy workarounds/hacks. We now follow what vuetify is doing and to also have no surprise with those compnoents.

- [x] Remove mods components
- [x] Replace them
- [x] Test everything


marcelveldt (approved):
Impressive work @stvncode 
This is a much appreciated big/first step in cleaning up the code !

PR #1106, opened 2025-09-05
Implement testing to MA and add pre-commit and ci 

stvncode:
Multiple goals for this pr:

- Implement vitest for basic component like the ones I changed recently (Button, Icon etc) and for the big utils functions. The plan is to try to start covering more functions/components in the future (we are at 4% for now, so still a long route to go).
- Implement pre-commit with husky. At least have for now the linting and the test (might remove the test at some point if it's taking to much time to people).
- Add a CI check for test

In order to make the pre-commit work, you should do a ```yarn prepare``` at the root.


marcelveldt (approved):
LGTM!

PR #1118, opened 2025-09-08
Update branding

OzGav:


marcelveldt:
Do you have a screenshot of the endresult how it looks like in the frontend ?

OzGav:
I'm actually not sure where these are used. I was just looking around and saw them and thought to update them. When I search the frontend repo for logo.png and logo.svg I don't see any results. 

Looks like icon.png is used in [src/layouts/default/DrawerNavigation.vue](https://github.com/music-assistant/frontend/blob/bff8f5d5f443b54f53bd4dcf851b43b0f7b551fd/src/layouts/default/DrawerNavigation.vue#L5)

icon.svg is used here [public/index.html](https://github.com/music-assistant/frontend/blob/bff8f5d5f443b54f53bd4dcf851b43b0f7b551fd/public/index.html#L7)

But I don't see where this actually appears in the UI? There is the browser tab favicon that also needs updating...

marcelveldt (approved):
let's merge it now and we can always adjust later of something doenst show up well

PR #1121, opened 2025-09-09
Update icons

OzGav:
Not sure where these are used but updating them!


marcelveldt:
These are the PWA icons so what you see on your phone/device when you install the page as app.
That means that we need a square icon, not a house shaped icon

OzGav:
Square images created

marcelveldt (approved):

PR #1122, opened 2025-09-10
Improve Multi select favourite

OzGav:
Currently you cant bulk remove from favourites. This change makes the logic work similarly to the recent mark played/unplayed change.

Also there was weird behaviour when making these sorts of changes as the checkboxes were getting cleared even though they were actually still selected. The simplest thing I found was to clear the checkboxes which resets the view and seems intuitive.

Addresses: https://github.com/orgs/music-assistant/discussions/3798


src/components/ItemsListing.vue
```diff
@@ -967,6 +968,12 @@ onMounted(async () => {
     loadData(true);
   }
 
+  // Listen for selection clearing events
+  (eventbus as any).on("clearSelection", () => {
```
stvncode: Instead of as any, can't we add to the Events type in the eventbus.ts file the clearSelection function ? Like : ```clearSelection: void``` 
This way, you can just do eventbus.on(...) without the any

stvncode (commented):
Except for the any, this looks really good 🙏

src/components/ItemsListing.vue
OzGav: You are the expert here @stvncode So happy for you to take this PR idea in any direction you like!

stvncode: Sure, it's just to prevent any if we can for more type safety 😊

stvncode (approved):

PR #1124, opened 2025-09-10
feat: Add the possibility to add folder to playlist

stvncode:
As the title say, this is adding to the folder menu a way to add to playlist.

### DIFF
--- a/src/layouts/default/AddToPlaylistDialog.vue
+++ b/src/layouts/default/AddToPlaylistDialog.vue
@@ -92,20 +92,19 @@
 <script setup lang="ts">
 import MediaItemThumb from "@/components/MediaItemThumb.vue";
 import ProviderIcon from "@/components/ProviderIcon.vue";
-import { MediaType } from "@/plugins/api/interfaces";
+import Toolbar from "@/components/Toolbar.vue";
+import api from "@/plugins/api";
 import type {
+  BrowseFolder,
   MediaItemType,
   MediaItemTypeOrItemMapping,
   Playlist,
-  Track,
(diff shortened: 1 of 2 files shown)


src/layouts/default/AddToPlaylistDialog.vue
```diff
@@ -192,18 +191,94 @@ const fetchPlaylists = async function () {
     }
   }
 };
+
+const expandFolderToTracks = async function (
```
marcelveldt, line 195: I'm pretty sure we already added this to the backend - did you test trying playMedia on a folder already ?

marcelveldt: Ha, yeah we indeed added support already in the backend but we never exposed it in the frontend :-)

https://github.com/music-assistant/server/blob/dev/music_assistant/controllers/player_queues.py#L1622

marcelveldt: Sorry ignore me here - I thought this was about playback of a folder but this is about adding the contents of a folder to a playlist, which is way more complicated.

marcelveldt:
Waiting for backend implementation

stvncode:
Closing for now, will create a new one when be will be updated

OzGav:
Woah I am 9 hours too late. I just came to look at this. Steven will you do something different or can I look at what you have done here and do the backend based off this?

stvncode:
The point Marcel made on this is that we were doing things here that could be done in the be, so i guess it'll be a different pr that's why i closed it. But if you do the be, I'll be happy to implement a new version of this 😃

PR #1139, opened 2025-09-18
Hiding the select button in case there's no reason to select items 

### LINKED ISSUE #1: Remove multi-select in top browse view levels (was: Useless overflow menu on the browse view)
The overflow menu on the "Browse" view is useless
Because of that, The multi-select option is also useless (There is no action behind it)

https://github.com/music-assistant/backlog/assets/5878296/790d3b96-6ba0-44ab-98a4-da262e923cc1

Gav - Most of this has been fixed. However I have editied the title as there is still wierdness in that you can multi-select at the top levels of the view which is non-sensical

2.4.0: You can still multi-select but the ACTIONS menu is blocked which is good. Just wanted to confirm this is the solution?

robsonke:
In case the view only shows folders and no playable items, there's no need to have checkboxes to do multi item operations.
Should fix https://github.com/music-assistant/backlog/issues/1

### DIFF
--- a/src/views/BrowseView.vue
+++ b/src/views/BrowseView.vue
@@ -6,7 +6,9 @@
       :show-library="false"
       :show-favorites-only-filter="false"
       :show-track-number="false"
-      :show-select-button="path && path != 'root' ? true : false"
+      :show-select-button="
+        path && path != 'root' && !hasOnlyFolders ? true : false
+      "
       :load-items="loadItems"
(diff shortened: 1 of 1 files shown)


OzGav:
@robsonke error in the build...

robsonke:
> @robsonke error in the build...

That's pretty silly, learned the hard way that `yarn dev` isn't too picky :)

stvncode (approved):
Thank you for the pr 😃

PR #1149, opened 2025-09-23
Update play icon for tracks and for home page

stvncode:
The goal of this pr is to reduce the opacity of the play icon in the home page and also center it:
Screenshot 2025-09-23 at 12 40 22

And also to change the play button on the track cover from this:
Screenshot 2025-09-23 at 12 44 21

To this: 
Screenshot 2025-09-23 at 12 39 43


marcelveldt:
Should we make the play icon on the album covers a bit more visible ? I would make it a bit bigger and a bit more opaque so you can better see its a play button/icon. The track one is fine

marcelveldt (approved):

PR #1150, opened 2025-09-23
First iteration of reworking the settings pages: the providers view

stvncode:
This is a proposal for a new settings ui:
https://github.com/user-attachments/assets/ead72ae7-fd39-4119-9ca1-ef907bf9f1f1


marcelveldt:
Feedback (but we can definitely discuss this next week either async or in a call):

- We should consider having all the provider cards on one single page (like HA integrations)
- Add filters to filter per provider type, show/hide disabled providers etc
- One single "add provider button"
- Have an icon on the provider card what type it is (e.g. a player for player provider)
- For Player provider: show the amount of players in the bottom of the card, which is also a link to players page (just like HA integration card links to devices)
- Player page also have a search box and filters to filter per provider etc.

So basically, let the settings screen for providers and players behave more like HA config page for integrations and devices.

Then after this step I think we should also split up the settings into different pages/sections, just like HA settings. So we can give a proper home to the core settings as well, a about/credits age and so on. But first lets focus on the providers and players pages! This first iteration already looks very promising!

OzGav:
This is beyond very promising! Love your work @stvncode

OzGav:
Looks like we should standardise the prrovider descriptions somewhat.....

marcelveldt (approved):
Amazing work @stvncode this looks amazing!
Cant wait to see this layout applied to the rest of the settings pages!

PR #1170, opened 2025-10-06
Let the user reduce the navbar

stvncode:
The main enhancement here is the possibility to have a dock icon at the bottom of the nav to reduce teh navabar to only icons.
On top of that, i added the return to home when you click to the icon/title at the top of the navbar + some reorder for providers

- [x] Navbar can be resized using localStorage for now
- [x] Sort providers by title and provider type
- [x] Return to hom on lcicking at the logo + name

Screenshot 2025-10-06 at 13 04 57
Screenshot 2025-10-06 at 13 04 45

### DIFF
--- a/src/components/navigation/Navigation.vue
+++ b/src/components/navigation/Navigation.vue
@@ -1,12 +1,12 @@
 <template>
-  <nav class="sidebar">
-    <div class="sidebar-header">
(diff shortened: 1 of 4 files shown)


marcelveldt (approved):
Nice work @stvncode !

PR #1171, opened 2025-10-06
New home for settings page

stvncode:
This is a small iteration for the settings home page. Instead of having the menu icons in the right toolbar that is not easy to find when you first enter the app, we now have a settings page (HA iso) with those elements.

I tried new icons and put some colors. Don't hesitate to propose others ideas.

Screenshot 2025-10-06 at 15 56 22


marcelveldt (approved):
Yes! This is a great start!

PR #1173, opened 2025-10-09
Modify the way volume controls work

stvncode:
This pr fix this issue: https://github.com/orgs/music-assistant/projects/2/views/1?pane=issue&itemId=132546511&issue=music-assistant%7Csupport%7C4294

- [x] On tap on the slider, the volume now do a volume up/down instead of going to the point
- [x] Visible draggable handle/circle 
- [x] Drag'n'drop don't change
- [x] Group filtering and sorting
- [x] Remove volume bar  when snt selected in groups


marcelveldt (approved):
Works very well, nice work!

PR #1174, opened 2025-10-16
Add the provider type on items on search

stvncode:
Screenshot 2025-10-16 at 08 38 19

Regarding the design, i tried to put on the top left corner of the card but i'm opened to other suggestions 😃


src/components/ItemsListing.vue
```diff
@@ -268,7 +270,7 @@ export interface Props {
 const props = withDefaults(defineProps<Props>(), {
   sortKeys: () => ["name", "sort_name"],
   showTrackNumber: true,
-  showProvider: Object.keys(api.providers).length > 1,
```
stvncode, line 271: @marcelveldt, if i don't put false here, it shows providers also in playlist (and maybe in others like audiobooks, radio etc since i don't have one of those yet). What was the need for this showProvider before ?

marcelveldt: It is what the title suggests, if the provider icons should be displayed. It's passed to the underlying view types. I think its only implemented however on the listitem and not on the panels but I'm not sure.

src/components/ProviderFilters.vue
```diff
@@ -12,7 +12,7 @@
     />
     <div class="d-flex ga-2 filter-buttons">
       <v-btn height="40" elevation="0">
-        {{ $t("sort.provider") }}
+        {{ $t("image_type") }}
```
marcelveldt: image type ?
maybe create a new translation key for provider_type instead ?

marcelveldt (approved):
Thanks, @stvncode !

OzGav:
This seems to have been positively received. I think it would also be good to have this on the PLAYLIST view when viewing by thumbs

stvncode:
@OzGav, can i let create a ticket on the backlog regarding this please ?

PR #1194, opened 2025-10-27
Volume slider tweaks

stvncode:
- [x] Make thumbsize bigger for now (for volume slider)
- [x] Haptic feedback (vibration on mobile while dragging)
- [x] Change the slider colours opacity when volume is muted


marcelveldt (approved):
Thanks, @stvncode !

PR #1203, opened 2025-11-09
Chapter time

tsipinakis:
Reopening [#1090](https://github.com/music-assistant/frontend/pull/1090). Which was closed after a wrong force push with an empty compare (github doesn't like it it seems).

@marcelveldt I see there was significant new changes in this part of the code, I'm not sure I have the time to re-implement this on top of the new things, it'd be great if you could help see this merged?

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
+++ b/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
@@ -151,6 +151,14 @@
               {{ $t("off") }}
             </v-card-subtitle>
 
+            <!-- SUBTITLE: player powered off -->
+            <v-card-subtitle
+              v-if="store.curChapter"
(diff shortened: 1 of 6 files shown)


marcelveldt:
Yeah, the big change you are talking about is that we now calculate the progress in the frontend instead of it being pushed every second by backend. Also the details in player.current_media now are always accurate, regardless the source as the player may be playing some live source instead of the our queue.

If you don't have time atm to adjust the PR to those changes, me or @stvncode can have a look next week

tsipinakis:
> If you don't have time atm to adjust the PR to those changes, me or @stvncode can have a look next week

I don't have a lot of time at the moment so if you can help resolve the conflicts it'd be great

teancom:
Obviated by https://github.com/music-assistant/frontend/pull/2548 but thank you!

PR #1211, opened 2025-11-20
POC: Add server endpoint for preloading next stream (frontend)

elialbert:
[backend PR](https://github.com/music-assistant/server/pull/2660)

This is a POC. It definitely needs some work in one spot - the hardcoded stream server port is obviously a no-go. I am happy to add that but I don't want to get into changing models and other stuff unless someone is going to be interested in helping me get this across the finish line.


marcelveldt:
As discussed on discord, this is not the way this should work. The next track is already preloaded by other parts of the backend

PR #1212, opened 2025-11-22
Replace the builtin player with Sendspin

elialbert:
This PR adds initial Sendspin support for the builtin web player.
Sendspin is a new audio protocol by the Open Home Foundation that allows for synchronized audio playback.
Now you can synchronize any device running the Music Assistant Web interface with any other Sendspin compatible speaker, all while playing perfectly in sync*.

\*You may need to tweak the sync delay under `Settings / User Interface`

# Known issues and Limitations

- The notification is still a bit buggy
- Only uncompressed PCM is supported for now
- You may still hear some stuttering and other audio glitches
- Not available through Home Assistant Cloud yet
- Not available through remote connections yet
- The [`outputLatency` Web API](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/outputLatency) is still fairly new. Accuracy varies widely by the Browser and OS used. If you have issues with inconsistent delay, try to disable it under `Settings / User Interface / Use browser latency detection`


src/views/settings/FrontendConfig.vue
```diff
@@ -33,13 +33,17 @@ import { onMounted } from "vue";
...
+import { store } from "@/plugins/store";
+import { getSendspinDefaultSyncDelay } from "@/helpers/utils";
 
 // global refs
 const router = useRouter();
 const config = ref<ConfigEntry[]>([]);
 const loading = ref(false);
 
 onMounted(() => {
+  const usingHaCloud = store.remoteAccessInfo?.using_ha_cloud ?? false;
```
marcelveldt: This is not the right way to detect this as this will be true when ha cloud is available, so even the local connection will then not be able to use sendspin.

There is some existing check somewhere in the frontend if we're in a ingress session.
So I would start by just disabling it when we detect ingress for now

marcelveldt (approved):
Amazing work @elialbert and @maximmaxim345 !

PR #1213, opened 2025-11-23
feat: Add Safari 15 / iOS 15 compatibility

sethbrammer:
## Summary

This PR adds compatibility for Safari 15 and iOS 15, enabling the Music Assistant frontend to work on older iOS devices that are still in widespread use (e.g., iPad Air 2, older iPads that cannot upgrade past iOS 15).

### Changes

- **`.browserslistrc`**: Added explicit `iOS >= 15` and `Safari >= 15` targets
- **`vite.config.ts`**: Added `safari15` build target for proper transpilation
- **`src/main.ts`**: Added `core-js` and `regenerator-runtime` polyfills for older WebKit
- **`src/styles/global.css`**: Added iOS Safari CSS compatibility tweaks:
  - `-webkit-overflow-scrolling: touch` for smooth scrolling
  - `-webkit-tap-highlight-color` for tap highlight removal
  - `-webkit-backdrop-filter` for blur effects
- **`package.json`**: Added required polyfill dependencies

### Why This Matters

Many users still have older iPads that are stuck on iOS 15 due to Apple's device support policy. These devices work perfectly fine but cannot run the current frontend due to missing ES6+ features and CSS compatibility issues.

### Testing

- Tested on iOS 15 Safari (iPad Air 2)
- Tested with Playwright WebKit

### Related Issues

Addresses requests from users with older iOS devices who cannot access the Music Assistant web interface.

---

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -4,6 +4,10 @@
  * Bootstraps Vuetify and other plugins then mounts the App`
  */
 
+// Polyfills for older browsers (Safari 15 / iOS 15 support)
+import "core-js/stable";
+import "regenerator-runtime/runtime";
+
 // Global styles
 import "@/styles/global.css";
 
--- a/src/styles/global.css
+++ b/src/styles/global.css
@@ -206,3 +206,20 @@ html {
   color: rgb(var(--v-theme-primary));
   font-weight: 800;
 }
+
+/* iOS Safari 15 compatibility tweaks */
+.scroll-container,
+.v-list,
+.v-navigation-drawer__content,
+.v-main__wrap {
+  -webkit-overflow-scrolling: touch;
+}
+
+* {
+  -webkit-tap-highlight-color: transparent;
+}
+
+.blur-background {
+  -webkit-backdrop-filter: blur(10px);
+  backdrop-filter: blur(10px);
+}
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,4 +1,6 @@
-> 1%
+> 0.5%
 last 2 versions
 not dead
 not ie 11
+iOS >= 15
+Safari >= 15
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+# Build stage
+FROM node:20-alpine AS builder
+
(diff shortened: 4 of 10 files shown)


src/styles/global.css
```diff
@@ -206,3 +206,20 @@ html {
   color: rgb(var(--v-theme-primary));
   font-weight: 800;
 }
+
+/* iOS Safari 15 compatibility tweaks */
+.scroll-container,
+.v-list,
+.v-navigation-drawer__content,
+.v-main__wrap {
+  -webkit-overflow-scrolling: touch;
```
stvncode, line 215: The doc say that it's non-standart (see https://udn.realityripple.com/docs/Web/CSS/-webkit-overflow-scrolling)

src/main.ts
```diff
@@ -4,6 +4,10 @@
  * Bootstraps Vuetify and other plugins then mounts the App`
  */
 
+// Polyfills for older browsers (Safari 15 / iOS 15 support)
+import "core-js/stable";
```
stvncode, line 8: What APIs does the codebase actually use that Safari 15 lacks ? It's just that importing all of core-js/stable is adding 259.8kB to the bundle and i want to know if this is necessary. Or maybe we can only polyfill what's actually needed to prevent adding the all lib

sethbrammer:
**Re: `-webkit-overflow-scrolling`**

You're right that it's non-standard - it's just a progressive enhancement for scroll momentum on older iOS. Browsers that don't support it ignore it.

**Re: `core-js/stable`**

Good call - 259kB is way too much. Tested on an actual iOS 15 device and only `AbortSignal.timeout()` needed polyfilling. Swapped it for a tiny inline shim:

```ts
if (typeof AbortSignal !== "undefined" && !AbortSignal.timeout) {
  (AbortSignal as any).timeout = (ms: number): AbortSignal => {
    const controller = new AbortController();
    setTimeout(
      () => controller.abort(new DOMException("TimeoutError", "TimeoutError")),
      ms,
    );
    return controller.signal;
  };
}
```

PR #1225, opened 2025-11-25
Fix validation for dropdowns when the selection value is 0

OzGav:
The validation rule was using a falsy check (!v) which incorrectly
  treated the integer value 0 as invalid for required fields. This
  affected sync interval dropdowns where "Disable automatic sync"
  has a value of 0.

image


stvncode (approved):
Thanks 😃

PR #1228, opened 2025-11-27
Fix content hidden for core settings server information

stvncode:
https://discord.com/channels/753947050995089438/1083643113484070953/1443540979792740354


marcelveldt (approved):
Thanks, @stvncode !

PR #1248, opened 2025-12-03
Add dot for players and providers filters when active + change dot color for toolbar items

stvncode:


marcelveldt (approved):
Thanks, @stvncode !

PR #1253, opened 2025-12-03
feat: Add Safari 15 / iOS 15 compatibility

sethbrammer:
## Summary
Adds compatibility for Safari 15 and iOS 15, enabling the frontend to work on older iOS devices (e.g., iPad Air 2).

## Changes
- **Minimal `AbortSignal.timeout` polyfill** (~200 bytes) - Safari 15 doesn't support this API
- **`-webkit-overflow-scrolling: touch`** - Progressive enhancement for scroll momentum on older iOS

## Why minimal polyfill?
Previous approach used full `core-js/stable` which added 259kB to the bundle. After testing on an actual iOS 15 device, only `AbortSignal.timeout()` needed polyfilling.

## Testing
- Tested on iOS 15 Safari (iPad Air 2)
- All unit tests pass

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -4,6 +4,19 @@
  * Bootstraps Vuetify and other plugins then mounts the App`
  */
 
+// Polyfill for Safari 15 / iOS 15 (AbortSignal.timeout not supported)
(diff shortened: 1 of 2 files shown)


stvncode (approved):
Thanks for implementing the changes, looks good now !

PR #1255, opened 2025-12-05
Open player group with an icon instead of the card

stvncode:


src/components/PlayerCard.vue
```diff
@@ -196,6 +196,18 @@
                 : 'mdi-play'
             "
         /></Button>
+        <Button
```
marcelveldt, line 199: would it be an idea if show a small badge with the number of speakers in the (sync)group ?
just like we do in the footer player thingie ?

stvncode: Oh yeah totally, that will help user understand the idea behind the icon !

marcelveldt (approved):
Nice work @stvncode !

PR #1259, opened 2025-12-07
Add the possibility to switch from listview to thumbs for players and providers

stvncode:


marcelveldt (approved):
Nice work again @stvncode !

PR #1269, opened 2025-12-10
Hide unavailable Sendspin web players from settings

maximmaxim345:
Hide unavailable Sendspin web players from settings

This prevents users from being confused by a lot of auto generated players
in the Players and Providers settings pages.

This functions similarly to how builtin web players were hidden before the switch to Sendspin.


marcelveldt (approved):
Thanks, @maximmaxim345 !

PR #1295, opened 2025-12-15
Change icons for player

stvncode:


src/components/PlayerCard.vue
```diff
@@ -177,34 +177,34 @@
       <!-- power/play/pause + menu button -->
       <template #append>
         <!-- play/pause button -->
-        <Button
+        <UiButton
```
marcelveldt, line 180: this name looks a bit weird "UiButton" - you'd almost say it should be UIButton.
Anyways, not important, just an observation

marcelveldt (approved):

src/components/PlayerCard.vue
stvncode: That's mb, I did that to keep the other instance of Button (vuetify) but I forgot to remove it. Will add that to another pr

PR #1300, opened 2025-12-16
Adjust PlayerSelect to always show all players, sorted on priority

marcelveldt:
Adjust PlayerSelect to always show all players, sorted on priority

@stvncode @Erioldoesdesign feedback wanted

I was always annoyed by the fact that we had some players at the top and some hidden behind "all players" and "all groups" without a good logic for the user to find out why a player would show at the top or in the collapsed section.

This PR adjusts this and will just show one big list of all your players, not hidden behind some collapse thingie.
I added a bit of priority filtering so playing players or ones you recently interacted with wil be shown at the top

BEFORE:
image

AFTER:

image


stvncode (dismissed):
This is a nice improvements !

Erioldoesdesign:
Good improvement but i do worry users will get 'list overwhelm' from a long list.

@stvncode suggested filters at the top of the right hand section to make a long list shorter via filters vs using Headings/expandable tabs/sections.

I also think a hr line divider below the suggested system players would be good e.g. a hr line below the current player or 'most used player'

I think there's something that can be done to make groups of players more visually clear but that's a separate issue for sure.

marcelveldt:
Lets keep this PR and branch open as draft so we can do some experimenting of what could work best

kepstin:
The specific issue I'm having (on a system that does not use any "This device" players) is that unless I expand the "All players" section, the only player I will see is the currently selected player; all the other players are hidden. This makes the menu rather annoying to use.

Since I don't use the "This device" players (and would actually [prefer to completely disable them](https://github.com/orgs/music-assistant/discussions/4505)), my list of available players is already small and curated, and I would prefer to see all of them all the time.

Maintaining a (per-user with global fallback?) "last used" timestamp on the players might allow auto-sorting them in a useful way. This could be combined with some default filters to e.g. hide unavailable or powered-off players, perhaps?

marcelveldt:
I moved ahead with this one and opted for a variation where the "all players" list is still there but it remembers the collapsed state (and shows both normal players and group players) - also added a search box when there are more than 6 players in the system. Top listing only shows 3 (higher prio) players based on recent interactions, play state etc.

Works pretty well and solves the immediate urgency. Can we later iterate on it from here

marcelveldt:
image

kepstin:
> opted for a variation where the "all players" list is still there but it remembers the collapsed state

Is this remembered per-device, per-user, or globally? In my case, I would prefer globally, or at least per-user, to avoid user interface inconsistencies when switching between devices.

PR #1313, opened 2025-12-17
Enable useCredentials in VitePWA.config so PWA also works behind basi…

cephos:
…c auth of reverse proxy


marcelveldt:
We now have support baked in for authentication, so you shouldn't be using basic auth anymore to access the PWA

cephos:
> We now have support baked in for authentication, so you shouldn't be using basic auth anymore to access the PWA

Ah, perfect. Thats even better. Thank you for the hint. Just updated and using it this way now.

PR #1318, opened 2025-12-21
Auto-focus on search box in add provider dialog

OzGav:
I would like to be able to start typing straight away when I click on the ADD A NEW PROVIDER button.


stvncode (approved):
Perfect, thanks 🙏

PR #1329, opened 2025-12-22
Speaker name (3rd row in player-bar at bottom) not showing up in Safari (+Mobile)

Thyraz:
Fix for this issue:
https://github.com/music-assistant/support/issues/4575#issuecomment-3680258591

The problem had two root-causes which caused unwanted line-clamping.
- We used the same CSS class-names as used in Tailwind, which applied unwanted styles
- A previous override to get rid of clamping applied from vuetify didn't work anymore, as it was defined in an scoped CSS block (maybe wasn't scoped in the past.

Behavior is now the same as on Chrome (which most likely don't listen to the old -webkit-* properties anymore and therefore wasn't affected.


MarvinSchenkel:
Just wondering, why did you rename the css class names? Feels like an unnecessary change to me? I think the only relevant changes here are the additions of `.player-track-detail` and `class="player-track-details"`?

I think @stvncode has a better view on this.

Thyraz:
> Just wondering, why did you rename the css class names? Feels like an unnecessary change to me? I think the only relevant changes here are the additions of `.player-track-detail` and `class="player-track-details"`?

Hi Marvin,

it's because the problem was caused by two clamping settings. 
And only when I fixed both the player name showed up for me.

One is caused by the scoped CSS, which avoided the over-ride to take effect.
This is the one you mentioned where I removed the scoping for this CSS block and added class="player-track-details".

The other one has a different source.
In `src/styles/style.css` we import the tailwind utilities with `!important`:
```JS
@import "tailwindcss/utilities.css" important;
```
This imports tailwainds own clamping styles:
image
In the compiled code these result to the same class names than our clamping rules:
- line-clamp-1
- line-clamp-2
- ...

If you look in the dev tools, you will therefore see 2 style blocks applied to `.line-clamp-1` in the frontend.
Only the second block contains code that we have in our own code-base.
The first one contains the style from tailwind.

I'm pretty sure this is a side-effect and not intended behavior?
image
Even more that it's one of the reasons why the player name is hidden in Safari.

Therefore I renamed the class names in our code, so the tailwind don't get applied to these elements anymore.
But maybe there's a more elegant way to solve this ...

stvncode (commented):
This is nice thanks ! The thing is that since we are adding shadcnui, we will have 2 css configuration (shadcn + vuetify), so possibly some conflicts like this unfortunately. Did you test if that didn't break anything for other part of the app ?

Thyraz:
> Did you test if that didn't break anything for other part of the app ?

I guess the only other effect this could have have, would be, that some of the wrongly applied CSS became beneficial for the current behavior of the app. ;)

I took a look at the listing options (like list, panel-view, ...) for item listings for the different media types and artists.
Both in Chrome and Safari.

Everything beside the fixed player name doesn't seem to be affected or changed.

stvncode (approved):
Thank you 🙏

PR #1334, opened 2025-12-24
Fix Auto focus search box

OzGav:
Looks like the last changes stopped this from working


stvncode (approved):
That's probably my bad sorry. Thanks for the fix 👍

PR #1335, opened 2025-12-24
Store some settings per user

OzGav:
User Interface settings now separated into two groups, those that are stored per user and those that are stored per browser


stvncode (dismissed):
I totally agree. Since we now can have multiple user, it's important for the configurations to be personal

OzGav:
Thanks Steven. Do you want to merge this as is and then do a follow up with the settings storage location change or do it all at once?

stvncode:
Would be lovely @OzGav 🙏

OzGav:
@stvncode OK I have tested these changes and they seem to work OK but this is way past what I would normally try and do frontend wise so please check carefully! Don't be shy pointing out the faults!

marcelveldt:
I dont like the categories "Per User Settings" and "Per Browser Settings". A category should describe the category of the setting(s), not where its technically stored.

If you even want to make it clear where they are stored (which I already doubt), you can add it to the description.

marcelveldt:
Is there a real user demand to add this feature ?
Note that with every feature we add (big or small) we add maintenance burden so lets try to focus on high demand features first please

OzGav:
The feature request is linked in the first post. I was going through all the requests and noticed this one with five votes and thought I could implement it fairly easily. Five votes might not seem like many but five votes feature requests are on page 3 of 13.

I have changed the section titles to “Personalization” and “Device Layout”

src/translations/en.json
```diff
@@ -382,6 +384,10 @@
                 "el_GR": "Greek"
             }
         },
+        "startup_view": {
+            "label": "Startup view",
+            "description": "Select which view to display when Music Assistant starts. Choose from Home, Search, Artists, Albums, Tracks, Playlists, Audiobooks, Podcasts, Radio, or Browse."
```
marcelveldt:
```suggestion
            "description": "Select which view to display when Music Assistant starts."
```

src/translations/en.json
```diff
@@ -301,6 +301,8 @@
             "audio": "Audio",
             "advanced": "Advanced settings",
             "generic": "Generic settings",
+            "per_user": "Personalization",
+            "per_browser": "Device Layout",
```
marcelveldt: I still don't think these names are good.

OzGav: Perhaps Preferences / Display ?

MarvinSchenkel:
How about we just always store preferences per user? If a user wants specific settings for let's say a wall-mounted tablet, they can create a dedicated user for that.

Rationale:
A user is likely logged in on MORE than 1 device , while a device is always logged in with 1 user.

OzGav:
Yes could work. This force mobile layout option, I recall, was to satisfy people who still wanted the bottom menu so they probably want that everywhere?

marcelveldt:
> Yes could work. This force mobile layout option, I recall, was to satisfy people who still wanted the bottom menu so they probably want that everywhere?

I think that would be device specific ?

marcelveldt:
Question I have is why we need this feature ?
You can simply open the URL directly on the page you want ? Why would this be needed to be configurable ?
What is the usecase and how many users request this ?

OzGav:
My understanding of the feature request is to allow the startup view to be configured no matter from where MA is accessed. So including the companion apps or ingress. The feature request currently has 6 votes putting it on page 3 of 15 (72/353) of the non-implemented feature requests.

src/plugins/router.ts
```diff
@@ -10,7 +10,14 @@ const routes = [
     children: [
       {
         path: "",
-        redirect: "/home",
+        redirect: () => {
+          // Read from user preferences first, fallback to localStorage for migration
```
marcelveldt: why ? this is a new setting, why does it need migration?

src/translations/en.json
```diff
@@ -315,6 +315,8 @@
             "playback": "Playback settings",
             "advanced": "Advanced settings",
             "generic": "Generic settings",
+            "per_user": "Preferences",
+            "per_browser": "Display",
```
marcelveldt:
```suggestion
            "preferences": "Preferences",
            "display_settings": "Display",
```

OzGav: Done

src/plugins/router.ts
OzGav: Yep mistake got carried away with copying and pasting. Removed. Also added comments to remove the other migration code in the future.

stvncode:
Hey @OzGav, what's the update on this ?

OzGav:
@stvncode I think it is readu to go. I merged Marcel's suggestions which was the last I did other than resolving some conflicts.

stvncode:
@Would it be ok to wait for Marcel to come back from PTO ? Since he started the review

OzGav:
Of course

marcelveldt (dismissed):

marcelveldt (approved):
Looks good!

PR #1337, opened 2025-12-24
Refacto settings page

stvncode:
https://github.com/user-attachments/assets/359a39b0-2636-4204-bbbc-3fe5c0330d60


stvncode:
@MarvinSchenkel, @OzGav wdyt about that ?

MarvinSchenkel:
What would this look like on mobile? And on desktop with sidebar expanded?

stvncode:
Column of one card. You have 1-2-3 cards depending of the screenwidht

OzGav:
Looks clever to me!

MarvinSchenkel:
I'm honestly not sure. To me, a settings page feels most natural as a list. Maybe we can have a similar approach here as for providers and players, where we have the option to switch between a list and tiles?

@Erioldoesdesign curious what you think

OzGav:
To be honest I thought this was the tile option for one of two options (the other being list). Sorry I lost track of what we currently have... :-(

stvncode:
@MarvinSchenkel, we can have both with an icon at the top. I just thought the previous version could need a refresh

MarvinSchenkel:
> @MarvinSchenkel, we can have both with an icon at the top. I just thought the previous version could need a refresh

That's nice! If it is consistent with the other views then I am fine with this. Will the default view be a list?

Erioldoesdesign:
> I'm honestly not sure. To me, a settings page feels most natural as a list. Maybe we can have a similar approach here as for providers and players, where we have the option to switch between a list and tiles?
> 
> @Erioldoesdesign curious what you think

@stvncode 

My suggestion would be dun dun dunnnnnnn Secret NEW OPTION! (the typical standard designer repsonse... 🙇)

The sizing and spacing here is terrible, please ignore that (i did a quick screenshot design!) but essentially our two sections that are most important/complex/sections people spend the most time in or need to find quickly in the settings are Providers and Players.

So they should not be the same 'size' as the other, less important or used 'boxes of settings. 

But yes I agree there should also be a list version that puts Providers and then Player 1 and 2 and then lists out the others.
Screenshot 2026-01-08 at 17 49 00

But yeah pelase ignore the 'balancing' of spacing here because it's not good:

stvncode:
https://github.com/user-attachments/assets/47bc68f9-0be1-481f-9136-ff520813b117

What do you think of this ? @Erioldoesdesign @MarvinSchenkel @OzGav ?

MarvinSchenkel:
> Screen.Recording.2026-01-12.at.11.59.53.mov 
> What do you think of this ? @Erioldoesdesign @MarvinSchenkel @OzGav ?

I love it.

Erioldoesdesign:
> Screen.Recording.2026-01-12.at.11.59.53.mov
> 
> What do you think of this ? @Erioldoesdesign @MarvinSchenkel @OzGav ?

Beautiful, Amazing, Love it. No notes :shipit: 

(thank you very much for considering/agreeing and implementing the design suggestion re. size 🙏 )

MarvinSchenkel (approved):

PR #1339, opened 2025-12-27
Fix PEQ decimal input boxes 

OzGav:
Allows decimal inputs to 2 places for frequency and gain and 3 places for q-factor. 
Retains existing behaviour of sliders
Adds handling of commas in the boxes


OzGav:
Marked as drafted as I haven’t had a chance to fully test it but wanted to get some feedback

PR #1342, opened 2025-12-31
Update of the fullscreen song page and also the footer

stvncode:


OzGav:
Would be great to close this as a result of this improvement https://github.com/music-assistant/support/issues/4286

marcelveldt:
Is there a screenshot and/or description to describe what exactly this brings ?

Thyraz:
Thanks @ozgav for linking this issue. 

I also would love to see it solved, as the overflow of the screen or in other cases large whitespace around the whole fullscreen player UI leads to UX problems or not that good looking results on some screen resolutions / window sizes. 

Would prefer, if the controls at the bottom would have a more predictable spacing to the bottom 

Also the song / quality info could use less rows / vertical space. 
Cover art is often tiny even on large screen. 

And players like smart clocks / media hubs with a horizontal pi screen (has a resolution of 800x480) are currently cut off at the bottom of the fullscreen player (while all the other parts of MA work fine, as they have a more clear layout).

Would really love it the fullscreen player would also be more usable here. 

I will try to find some time later STVN (and ongoing when this draft evolves) to run this PR locally to give feedback. 

Thanks a lot for working on this. 🙂

Thyraz:
I really like these new icon set with the more rounded glyphs a lot. 🙂👍
Do you plan to tone them done a bit as before (the old ones have only been completely opaque on hovering), or is the goal a little bit more contrast on this view?

And I guess I'm not really a fan of the white / black (based on theme) fill of the play icon, instead of seeing the background color through.
But not sure if this is intentionally, or just wip. 😉

Thyraz:
I‘ve seen Eriol‘s link here to the Figma file, and with the transparency of the icons, it looks really good. 👍 

https://github.com/music-assistant/backlog/issues/46

image

stvncode:
I'm closing this due to research that we are doing regarding design. Will create a new one when we'll be ready

PR #1343, opened 2026-01-02
Apply styles to the new volume sliders to match the rest of the UI

Thyraz:
Not sure how you all feel about the current look of the new volume slider.
I know it just started to finally behave like it should UX wise, thanks to the amazing work from _stvncode_
So please don't take this as critique in any way.

But I guess there's room for UI improvements now, so I tried to make it look more consistent with the rest of the UI.

**Don't be afraid: The knob is just visually looking smaller. It's real size and the hit target has not been changed.
So it remains easy to use on touch / mobile devices.**

Beside of that I tried to recreate the look of the old slider as close as possible (which is still used for the track position in the player views, so both can be visible at the same time).
I still like it a lot and think it fits the UI very well while also providing a modern look.

This animated GIF should show a few comparison shots of the current  vs. the customized look.

vol-slider

But not sure if this is the way you guys want to go with the new UI.
I created this draft with the changes I made.
Feel free to comment or to ignore it. 🙃


marcelveldt:
I personally like the blue-ish slider and the smaller knob. So a bit in-between what you are proposing.
I'll leave final review with @stvncode

stvncode (commented):
I agree with Marcel's point on the color. Otherwise looks good. We can start with the knob improvement and see if users like it. Otherwise we can switch back to dark/white

Erioldoesdesign:
UX/UI perspective here 👋 

So (as always) there's a wider challenge here re. MA. The #18BCF2 blue colour is used in many places in MA (and HA) to indicate to users 'Hey, you can edit/change/manipulate this!' and/or 'Hey, this is actively selected or selectable and does something!'

So blue = editable, changeable by the user. It's a visual clue in the UI that the user is in control.

Moving to just grey doesn't give users that visual clue anymore but! that's kind of ok because with a volume slider particularly, people know they can change it based on the position of the slider's fullness/extended bar and also by the icon. Also people just know instictually they can turn volume up or down using this (or using keys on a keyboard)

So my personal preference is a like the grey, it looks more consistent and i don't think users will have a problem locating this to edit the volume without the blue. But i don't think there's a big UX/UI problem with the original either.

The discussion this opens is 'How and when do we use #18BCF2 blue to indicate what 'state' the UI is in or can do?' <- that's a whole seperate new discussion to have.

Re. the knob/circle size on the slider. My preference is always for larger hit areas for UI you can click/tap/change due to accessibility reasons. Folks with low dexterity, hand tremors or folks that struggle with fine motor control need bigger buttons and hit areas. not JUST bigger hit areas that are invisible because then the visible indicator operates differently to the function which can be confusing. Imagine this is real life, you have a button on your oven that is 2 inches in diameter and you press 4 inches to the left of it by accident and oh no! i've accidentally turned my oven on because i didn't know the invisible button was bigger than the button i can see!

Obviously buttons on music assistant are not starting something as dangerous as a fire, but when the visible and the function don't match, users get confused or make mistakes that they cannot understand if they didn't know somehow that invisible button is bigger than see-able button.

Thyraz:
@Erioldoesdesign maybe it looks too small then and we need to choose something in between? 
But having it as large as the hit-area looks also kinda out of place to me. 

If you look at mobile apps, they all do the same thing. It's already build into the UI frameworks from Apple, Google, ...
If I remember correctly, Apple even had a UI talk at the developer conference about that concept back in time when they allowed public development and started the App Store with iPhone OS 2.0.
As this was the era where the world switched from tiny UI elements on stylus operated handhelds, to finger operated capacitive touchscreens.

So the touch area should always be large enough to forgive a not-so-perfect touch.
But increasing all elements visually to reflect that also looks kinda odd, even more on small device screens where you can't have too much whitespace around everything.

But you're most likely right: By recreating the style of the track position slider it's now going to the other extreme.

About the color: I also think blue is ok here.
But the current colors look way better on the dark theme than the bright one. Could also be the combination with the bright knob.
Also the bigger visual difference in colors compared to the track position slider is more standing out on the bright theme.

I'll play around with it a bit more over the next days and show some alternatives with a little bit bigger knob and try to look how a colorful version might end up fitting the bright theme more.
Will show some alternatives, then we can see if something is ok to be used for a short-term improvement on the volume bar or not ...

Erioldoesdesign:
> @Erioldoesdesign maybe it looks too small then and we need to choose something in between? But having it as large as the hit-area looks also kinda out of place to me.
> 
> If you look at mobile apps, they all do the same thing. It's already build into the UI frameworks from Apple, Google, ... If I remember correctly, Apple even had a UI talk at the developer conference about that concept back in time when they allowed public development and started the App Store with iPhone OS 2.0. As this was the era where the world switched from tiny UI elements on stylus operated handhelds, to finger operated capacitive touchscreens.
> 
> So the touch area should always be large enough to forgive a not-so-perfect touch. But increasing all elements visually to reflect that also looks kinda odd, even more on small device screens where you can't have too much whitespace around everything.
> 
> But you're most likely right: By recreating the style of the track position slider it's now going to the other extreme.
> 
> About the color: I also think blue is ok here. But the current colors look way better on the dark theme than the bright one. Could also be the combination with the bright knob. Also the bigger visual difference in colors compared to the track position slider is more standing out on the bright theme.
> 
> I'll play around with it a bit more over the next days and show some alternatives with a little bit bigger knob and try to look how a colorful version might end up fitting the bright theme more. Will show some alternatives, then we can see if something is ok to be used for a short-term improvement on the volume bar or not ...

I hear you re. the the 'big enough for a mis-tap/click'. I think perhaps this is me going too hard a stance on what could arguably be an 'accessibility mode' for people that need much bigger UI components for disabilities etc. and then have a medium, in-between for most folks who are reasonably dexterious and able-bodied right now. Happy to be having the discussion though and appreacite the effots here - I opened up a discussion thread on our discord server to see what other folks might think about these specifics too

Erioldoesdesign:
@stvncode messaged me re. this PR. I'm fine with this being merged and feedback being gathered in ths related discord post: https://discord.com/channels/753947050995089438/1458485616592289944

I'm stillnot sure this sizing/colour combination is as inclusive as we could be while also taking inspirationf rom other music streaming players but let's see what people sya :)

Thyraz:
Like stated above, 
I'm also not sure anymore if we should really go "that small" for the handle.

But beside of that, I still like the overall styling and colors of the old slider style that this PR mimics.
At least this PR decouples visual and functional size and we can adapt the visual size regarding to user feedback later (or already start with a little bit larger size)?

If we go for a different sized handle than this small one, I would also suggest to adopt the track position slider the same way.
But I bet you had a lot concept talks / ideas about the new player-view lately.
So not sure what the current plans are.

Would still be happy to see this merged as a starting point for the slider style in case it fits the new UI ideas.
But it doesn't need to be merged, just for the sake of merging. ;-)

MarvinSchenkel:
> @stvncode messaged me re. this PR. I'm fine with this being merged and feedback being gathered in ths related discord post: https://discord.com/channels/753947050995089438/1458485616592289944
> 
> 
> 
> I'm stillnot sure this sizing/colour combination is as inclusive as we could be while also taking inspirationf rom other music streaming players but let's see what people sya :)

@Erioldoesdesign 
I would love to see this PR get merged even though we are not 100% sure yet. I much rather get feedback on 90% of the PR than no feedback at all.

The only thing we need from you (while you're figuring out the color etc.) is just whether or not this is a step in the right direction. So happy to hear this is the case.

We can simply change the color in a follow up PR while we are getting feedback on this change. I think small incremental steps are better than trying to figure out the whole thing at once.

stvncode (approved):

PR #1344, opened 2026-01-03
Feature: Party mode

apophisnow:
This adds a party mode which does a few things.

New Party Mode dashboard view shows played, currently playing, and upcoming tracks in a Jukebox style carousel.
QR code on the Party Mode view allows guest access without having to log in. Guests are restricted to a page which only allows adding songs to the queue on the assigned party mode player.
Configurable throttling for "Play Next" and "Add to queue" for guests.
There is also a models and server PR for this feature.


stvncode (commented):
@MarvinSchenkel told me that it was almost done and you wanted an early review. Instead of adding a lot of comments, I'll just make it one for readability.

- You are using vuetify ui component and not shadcn component (we are switching slowly but adding more vueitfy will require work to refacto later). The thing is that it can be annoying for you and long, so I think we can merge like this and I'll later change them (if that's ok for you for sure).
- File size. For some file like GuestView for example, it's more than 1k lines, who's difficult to review but will be difficult to maintain with time as well. Files should not have more than 300 or 350 max max in my opinion. SO if you can split them into more components, it would be lovely (files are GuestView, PartyView for the biggest)
- For the ts file (useRateLimiting, useGuestSearch, useGuestQueue), it would be nice to have unit test
- They are some duplicate codes that can be extracted to only one component (in GuestView for example)

They are some other minor issues, but I'll address them after this is done.

This is only for the code, I didn't test the ux/ui.

Sorry for all these but the feature is pretty big. I just wanted to say that this is insane, thanks a lot for your work 🙏 

Don't hesitate if you need help on fixing those issues, I can definitively help on this 😃

src/plugins/router.ts
```diff
@@ -228,6 +246,38 @@ const routes = [
...
+                    resolve();
+                  }
+                },
+                { immediate: true },
+              );
+            });
+          }
+          // Only allow access if party mode plugin is enabled
+          if (!store.enabledPlugins.has("party_mode")) {
+            next({ name: "home" });
```
stvncode:
```suggestion
            next({ name: "discover" });
```

stvncode: This changed not so long ago sorry

src/views/PartyModeGuestView.vue
```diff
@@ -0,0 +1,546 @@
...
+      :skipping-song="skippingSong"
+      :skip-token-countdown="skipTokenCountdown"
+      :boost-badge-color="boostBadgeColor"
+      :request-badge-color="requestBadgeColor"
+      @skip="skipCurrentSong"
+      @queue-scroll="handleQueueScroll"
+    />
+
+    <!-- Snackbar for feedback -->
+    <v-snackbar v-model="snackbar.show" :color="snackbar.color" :timeout="3000">
```
stvncode: We use toast now everywhere from vue-sonner (toast.success, toast.error, toast.info etc)

src/plugins/auth.ts
```diff
@@ -53,12 +97,37 @@ export class AuthManager {
...
+    return this.claims?.[key];
+  }
+
+  /**
+   * Check if this is a party mode guest session.
+   * Party mode guests authenticate via QR code/join code and have
+   * restricted UI access (only the guest view).
+   */
+  isPartyModeGuest(): boolean {
+    return this.claims?.username === "party_mode_system";
```
stvncode: This is fragile no ? Any real user with this username would be permanently locked into guest mode. Since the JWT already contains a role and permissions array, why should we not use that ?

src/composables/useRateLimiting.ts
```diff
@@ -0,0 +1,328 @@
...
+   * Apply server config to rate limiting parameters.
+   * Called after fetching party_mode/config from the server.
+   */
+  const configure = (config: PartyModeConfig) => {
+    rateLimitingEnabled.value = config.enable_rate_limiting ?? true;
+    addQueueEnabled.value = config.enable_add_queue ?? true;
+    boostEnabled.value = config.enable_boost ?? true;
+    skipSongEnabled.value = config.enable_skip_song ?? true;
+
+    ADD_QUEUE_MAX_TOKENS.value = config.add_queue_limit || 10;
```
stvncode: Maybe use ?? instead of || for this function

src/layouts/PartyModeGuestLayout.vue
```diff
@@ -0,0 +1,11 @@
+<template>
```
stvncode, line 1: Why use v-app/v-main and not a div since it doesn't need vuetify layout ?

stvncode (commented):
On top of that:
- Isn't that a bit weird that you cannot click on a queue item ?
- The navigation need to be improved imo. For example, the only way of coming bak from a search to queue is to remove the search. Maybe some tabs could help ?
- Better matching color with OHF. Colors look really ai generated

Except for those couple of things, it's really awesome 🙏

src/layouts/PartyModeGuestLayout.vue
apophisnow: Fixed

MarvinSchenkel (approved):
It's PARTY TIME!!!! 🎉

PR #1345, opened 2026-01-03
add translation keys for abs and itunes podcasts

fmunkes:
PR to make recommendations and browse folders of abs and itunes podcasts translatable


marcelveldt (approved):
Thanks, @fmunkes !

PR #1357, opened 2026-01-07
Navigate from Now Playing View to detailed view if available otherwise search

OzGav:
This change adds the following functionality when clicking on:
* Track/Artist/Album which is in any connected provider's library or the MA library→ Navigate to that provider's detail page
* Track/Artist/Album when a radio station is playing → Lookup up the item in the MA library. If found navigate to the items detail view
* Track/Artist/Album not in any library  → Fall back to global search
* Album field which shows the radio station name → Do nothing


stvncode (approved):
Nice 👍

PR #1370, opened 2026-01-16
Reduce audio distortion with the Sendspin web player

maximmaxim345:
Upgrades `sendspin-js` to 1.0.1 which includes an improved syncing strategy that reduces pitch distortion during connection startup.

The web player now uses "quality-local" correction mode when playing solo and switches to "sync" mode only when grouped with other players. Solo playback no longer applies pitch shifting, eliminating unnecessary audio artifacts.

Also fixes syncing issues after ungrouping and regrouping a web player.


MarvinSchenkel (approved):
LGTM!

PR #1376, opened 2026-01-19
EditPlayer: If player is not available hide save button and show a banner instead

gnumpi:
If a player is not available, the Edit Player page currently does not show any configuration entries but still displays the Save, Close, and Restore Defaults buttons. This PR instead shows a banner informing the user that the player is currently unavailable and hides the edit configuration buttons.


marcelveldt (approved):
Thanks, @gnumpi !

PR #1377, opened 2026-01-19
Fix: Auto re-sync remote mode when the sw is restarted

MarvinSchenkel:
This PR attempts to fix a race condition in the service worker that cause 404s on the imageproxy.

The race condition happens when the service worker gets restarted and sets [`isRemote = false`](https://github.com/music-assistant/frontend/blob/ad06c2ee140d9a493e693f69eb5d054f30709b6b/public/sw.js#L232) by default. This causes image requests from app.music-assistant.io to [not be proxied ](https://github.com/music-assistant/frontend/blob/ad06c2ee140d9a493e693f69eb5d054f30709b6b/public/sw.js#L89)to the backend, thus accessing them directly.

Solution is to listen for an event that tells us the service worker has restarted and then setting remote mode, so requests are properly proxied again.

### DIFF
--- a/src/plugins/remote/http-proxy.ts
+++ b/src/plugins/remote/http-proxy.ts
@@ -69,6 +69,9 @@ class HttpProxyBridge {
       try {
         await navigator.serviceWorker.register("./sw.js");
 
+        // Set up SW restart detection
+        this.setupServiceWorkerReconnect();
+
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):
LGTM

PR #1380, opened 2026-01-22
Add helpers/logic to handle companion app(s)

marcelveldt:
Add a tiny bit of handling to allow companion apps to leverage the frontend (or just components of it).

- Allow send messages to/from companion app (native) layer to the frontend
- Disable sendspin player and browser controls in companion mode
- Allow companion app running sendspin to be detected as "this device" - just like web player
- Fix a bug in remembering last used player so we can correctly remember the last used player (even if its temporary unavailable)

@maximmaxim345 @formatBCE

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -19,12 +19,12 @@
       webPlayer.interacted == true
     "
   />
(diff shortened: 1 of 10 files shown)


maximmaxim345 (approved):
LGTM

PR #1383, opened 2026-01-23
EditPlayer: Fix typo in configEntry/translation key

gnumpi:
Currently there is key mismatch between the ConfigEntry key for the 'DSP not supported in group' banner and the translation keys. As it is only an ui helper configEntry this PR simply changes the key of the configEntry itself to match the translation keys.


stvncode (approved):
Thanks 🙏

PR #1384, opened 2026-01-23
Separate UI-only config entries from API schema; inject DSP link via UI entry helpers

gnumpi:
## Summary

This PR removes frontend-only config entry types from the API schema and introduces a UI-layer model for injected configuration entries.  
DSP settings are now exposed in the UI via injected entries instead of extending the server `ConfigEntryType`.

This keeps the API contract aligned with actual server responses while allowing the UI to render and handle additional, non-persisted settings cleanly.

---

### API contract cleanup
- Removed the frontend-only `DSP_SETTINGS` value from `ConfigEntryType`
- Ensures the enum strictly reflects values sent by the backend

### UI config entry model
- Introduced `ConfigEntryUI` and injected entry helpers in `helpers/config_entry_ui`
- Added type guards (`isInjected`, `isDspLinkEntry`) for safe narrowing
- Added a factory (`makeDspLinkEntry`) to construct injected entries with proper defaults

### Settings UI updates
- `ConfigEntryField.vue` now renders DSP settings via `isDspLinkEntry` instead of checking an API enum
- DSP action button respects disabled/read-only state

### Form handling & persistence
- `EditConfig.vue` ignores injected entries when:
  - detecting unsaved changes
  - collecting values for submission
- Removed the previous workaround that deleted DSP keys before save

### Player settings injection
- `EditPlayer.vue` injects the DSP settings link using `makeDspLinkEntry`
- Group-related informational labels are now marked as injected UI entries

---

## Why

- Prevents UI-only concepts from leaking into the backend type system
- Makes injected UI elements explicit, type-safe, and non-persisted by design
- Eliminates special-case cleanup logic during save
- Improves long-term maintainability of the settings model

---

## Notes / Testing

- DSP settings link renders for non-group players and respects disabled state
- Injected entries do not trigger “unsaved changes” warnings
- Injected entries are never included in the payload sent to the backend
- Group informational labels continue to render correctly

### DIFF
--- a/src/helpers/config_entry_ui.ts
+++ b/src/helpers/config_entry_ui.ts
@@ -1,0 +1,54 @@
+import type { ConfigEntry, ConfigEntryType } from "@/plugins/api/interfaces";
+
+export type ConfigEntryUIOnlyType = "dsp_settings_link";
+export type ConfigEntryUIType = ConfigEntryType | ConfigEntryUIOnlyType;
+
+export type InjectedConfigEntry = Omit<ConfigEntry, "type"> & {
+  injected: true;
+  type: ConfigEntryUIOnlyType;
+  i18nParams?: Record<string, string | number>;
+};
+
+export type DspLinkConfigEntryUI = InjectedConfigEntry & {
+  type: "dsp_settings_link";
+  note_key?: string;
+};
+
+// server entries stay unchanged
+export type ServerConfigEntryUI = ConfigEntry & {
+  injected?: false;
+};
+
+export type ConfigEntryUI = ServerConfigEntryUI | InjectedConfigEntry;
+
+export const isInjected = (e: ConfigEntryUI): e is InjectedConfigEntry =>
+  "injected" in e && e.injected === true;
+
+export const isDspLinkEntry = (e: ConfigEntryUI): e is DspLinkConfigEntry =>
+  isInjected(e) && e.type === "dsp_settings_link";
+
+const DEFAULT_DSP_LINK_ENTRY: Omit<
+  DspLinkConfigEntryUI,
+  "value" | "default_value"
+> & {
+  value: boolean;
+  default_value: boolean;
+} = {
+  injected: true,
+  type: "dsp_settings_link",
+  key: "dsp_settings_link",
+  category: "audio",
+  label: "",
+  required: false,
+};
+
+export function makeDspLinkEntry(
+  overrides: Partial<DspLinkConfigEntryUI> = {},
+): DspLinkConfigEntryUI {
+  return {
+    ...DEFAULT_DSP_LINK_ENTRY,
+    ...overrides,
+    injected: true,
+    type: "dsp_settings_link",
+  };
+}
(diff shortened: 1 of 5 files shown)


marcelveldt:
Conceptual I think this is fine but I'd like another pair of eyes from @maximmaxim345 (due to the DSP related config) and @stvncode for frontend architecture

gnumpi:
Thank you, actually the DSP related config is the main reason for this changes. I will create a (draft) PR in just a moment explaining the overall idea....

gnumpi:
> Thank you, actually the DSP related config is the main reason for this changes. I will create a (draft) PR in just a moment explaining the overall idea....

https://github.com/music-assistant/frontend/pull/1400

marcelveldt:
What exactly does this fix/offer ?
Also, be aware that backend can actually send config entries of type UI, for example alerts and labels.
They just do not need to be tracked or sent back

gnumpi:
> What exactly does this fix/offer ? Also, be aware that backend can actually send config entries of type UI, for example alerts and labels. They just do not need to be tracked or sent back

What this introduces is a frontend-side mechanism to *augment* the list of config entries with additional render-only items and metadata, in a way that scales beyond the current DSP link/button special-casing.

Concretely, this:

- **Generalizes the existing DSP link/button pattern**, where render-only items are injected into the edit view, but makes it extensible and reusable.
- Introduces a clear marker (`injected: true`) so injected entries can be **reliably filtered out** (for example when building payloads to send back to the backend), instead of relying on hard-coded key checks.
- Allows injected entries to carry **additional UI-only parameters** (`note_key`, `i18nParams`, action parameters, etc.), which is required for the dynamic player settings work (PR #1400), for example to support links to other players in alerts or labels.
- Makes it possible to include config entries that **originate from e.g. another player** as **read-only, inherited entries**, without tracking them as local changes or accidentally sending them back to the API.

I intentionally avoided changing the API interfaces, as they should remain a 1:1 representation of backend data. Instead, this introduces a UI-layer union type (`ConfigEntryUI = ConfigEntry | InjectedConfigEntry`) that keeps the backend model untouched while providing a scalable way to handle UI-only and read-only injected entries on the frontend.

Also worth mentioning: I’m still fairly new to TypeScript, so I’m very open to feedback and to reworking this toward a preferred pattern if there’s a better approach.

maximmaxim345 (dismissed):
Looks good to me!
I put it in `ConfigEntryType` initially just because it was easier to implement, but this looks like more future proof approach.
Thanks @gnumpi !

stvncode:
@gnumpi Can I let you rebase the pr please ? So we can merge it afterwards 😃

gnumpi:
> @gnumpi Can I let you rebase the pr please ? So we can merge it afterwards 😃

sure, I will do today.

gnumpi:
> @gnumpi Can I let you rebase the pr please ? So we can merge it afterwards 😃

sorry, just found the time to rebase this PR now. It's done.

stvncode (approved):

PR #1386, opened 2026-01-25
Use Player current media for Browser Media Session API to support non-track media types

tamwahba:
This PR updates media browser metadata helper to use the `player.current_media` its info instead of the player `QueueItem`. This change should bring the info shown by the media session API more inline with the built-in `PlayerOSD` since that [uses](https://github.com/music-assistant/frontend/blob/main/src/layouts/default/PlayerOSD/PlayerTrackDetails.vue#L109-L138) the player current media.

I noticed this issue when trying out the Pandora provider, but this is an issue with all radio providers (tested with the RadioBrowser provider as well) since the QueueItem is the whole radio station and only includes track details in the streamdetails.


marcelveldt (approved):
LGTM
Thanks, @tamwahba !

PR #1400, opened 2026-01-27
PlayerEdit: Dynamic Group Playback Settings

gnumpi:
Context
When a player is part of a group, certain settings are applied at the group level, while others remain player-specific. Which settings apply where depends on the player’s role within the group, the group type, and the capabilities of the player (provider). When a player is not part of a group, all settings apply at the individual player level.

Proposed change
This PR experiments with reflecting this distinction dynamically in the player settings UI. It introduces a dedicated Group Playback panel that is shown only when a player is part of a group. The panel displays group-level settings, including settings inherited from the group leader, which are shown as read-only where applicable. This helps clarify which settings are controlled at the group level versus per player.

In addition, this approach is intended to pave the way for supporting DSP-related settings at the group level in a future iteration.

Player not in a group:
settings_ungrouped

Group leader with no supported dsp settings per player:
group_leader_no_dsp

Group member with supported per player dsp settings:
group_player_dsp

Group member with not supported per player dsp settings:
member_player_no_dsp


stvncode:
Hey @gnumpi, this pr is still in draft, is this still something needed ?

gnumpi:
> Hey @gnumpi, this pr is still in draft, is this still something needed ?

Thanks, for the reminder! The UI has diverged quite a bit since then and probably made this PR obsolete. I will close it now and re-open a new one if needed.

PR #1404, opened 2026-01-27
Add group mute support to volume controls

scyto:
## Summary

- Group mute button calls new `group_volume_mute` server API instead of looping through individual members
- Group mute icon reflects aggregate state (muted only when ALL members are muted)
- Bottom bar mute icon is also group-aware
- Child player volume sliders remain interactive when muted (server uses `ATTR_MUTE_LOCK` to preserve mute state during group volume changes)
- Add `playerCommandGroupVolumeMute` and `playerCommandGroupMuteToggle` API methods
- Fix main volume slider not calling API (upstream regression from bb0dbf9a)

## Context

This is the frontend counterpart to music-assistant/server#3034 which adds the `cmd_group_volume_mute` API endpoint and `ATTR_MUTE_LOCK` attribute.

The maintainer's commit d1600778 added a frontend-only workaround that loops through group members calling individual mute, with a TODO: "revisit this when api/server supports group mute toggle". This PR replaces that workaround with a single server API call.

**Note:** This PR also fixes an upstream regression where the main volume slider's `@update:model-value` handler was reduced to just `mainDisplayVolume = $event` (no API call), making the volume slider a visual-only no-op. The fix restores the API call and makes it group-aware.

### Behavior changes

| Action | Before | After |
|--------|--------|-------|
| Click group mute button | Only leader muted | All members mute/unmute together via server API |
| Group mute icon | Reflected leader's state | Muted only when ALL members are muted |
| Group volume slider (with muted members) | Auto-unmuted members | Muted members stay muted (ATTR_MUTE_LOCK) |
| Child player slider (muted) | Slider disabled | Slider interactive, volume changes preserved |
| Main volume slider | Not calling API (regression) | Calls group_volume or volume_set correctly |
| Solo player | No change | No change |

### Files changed
- **VolumeControl.vue** — `isGroupMuted()` helper, group-aware mute icon/slider logic, volume slider API fix
- **VolumeBtn.vue** — group-aware mute icon in bottom bar
- **index.ts** — new `playerCommandGroupVolumeMute` and `playerCommandGroupMuteToggle` API methods
- **helpers.ts** — `handlePlayerMuteToggle` uses single group API call instead of member loop
- **utils.ts** — `getVolumeIconComponent` accepts optional `isMuted` override parameter

## Dependencies

Requires music-assistant/server#3034 for the `group_volume_mute` server API.

## Test plan

- [ ] Click group mute button → all members mute
- [ ] Click group unmute button → all members unmute  
- [ ] Mute 2 of 4 members individually → group icon shows unmuted
- [ ] Mute all 4 members individually → group icon shows muted
- [ ] Drag group volume slider with muted members → muted members stay muted
- [ ] Drag muted child player slider → volume changes, player stays muted
- [ ] Drag group volume slider → volume actually changes (API called)
- [ ] Solo player mute/unmute behavior unchanged
- [ ] Solo player volume slider works correctly


scyto:
demo docker image  `ghcr.io/scyto/ma-server-patched:group-mute`
updated with latest commits 2026-02-03

scyto:
backend PR now merged

src/components/VolumeControl.vue
```diff
@@ -293,11 +296,26 @@ const canExpand = computed(() => {
   return compProps.player.group_members.length > 0;
 });
 
+const isGroupMuted = function (player: Player): boolean {
+  if (!player.group_members.length) {
+    return !!player.volume_muted;
+  }
+  // For group players, only show muted if ALL members are muted
```
marcelveldt: For a future improvement, we can consider extending the Player model with an additional group_muted so we can do this logic in the backend, available for all clients of the MA api.

https://github.com/music-assistant/models/blob/main/music_assistant_models/player.py#L333

scyto: thanks @marcelveldt is that required before this PR can move forward?

scyto:
@marcelveldt updated to fix upstream bug and rebase this PR fork/branch with latest frontend changes

as alwasy built and tested, dev image is at `image: ghcr.io/scyto/ma-server-patched:group-mute`

src/components/VolumeControl.vue
```diff
@@ -264,6 +274,20 @@ const canExpand = computed(() => {
   return compProps.player.group_members.length > 0;
 });
 
+const isGroupMuted = function (player: Player): boolean {
```
stvncode: Duplicate here and in VolumeBtn. Also why is this a function and the other a computed value ?

src/plugins/api/index.ts
```diff
@@ -1395,6 +1395,20 @@ export class MusicAssistantApi {
...
+  public playerCommandGroupVolumeMute(
+    playerId: string,
+    muted: boolean,
+  ): Promise<void> {
+    return this.playerCommand(playerId, "group_volume_mute", {
+      muted,
+    });
+  }
+  public playerCommandGroupMuteToggle(playerId: string): Promise<void> {
+    return this.playerCommandGroupVolumeMute(
```
stvncode: Maybe I'm wrong but playerCommandGroupMuteToggle reads !this.players[playerId].volume_muted (the leader's raw flag) to decide
  whether to mute or unmute, but the UI icon uses isGroupMuted() which only returns true when all members are muted. These two can disagree.

The mute toggle reads the leader's volume_muted flag to decide the next state, but the icon uses an allPlayers check (isGroupMuted). These can disagree. If the leader is muted but one player isn't, the icon shows "unmuted" but clicking will send muted: false and unmute everything instead of muting.
                                                                                                              
A fix can be: !isGroupMuted(player) as the target state instead of !leader.volume_muted.

src/components/VolumeControl.vue
scyto: Good catch — extracted isGroupMuted into a shared plain function in helpers.ts (takes a Player param) and both components now import it from there. 

In VolumeBtn, it was previously a standalone computed because it always read from store.activePlayer. Now it's called as a plain function inside the volumeIconComponent computed, which still needs to be a computed itself since it selects the icon based on both mute state and volume level. Reactivity is preserved since Vue tracks the dependencies when the computed re-evaluates.

the real why is my lack of understanding of good programming and relying on claude, let me know if i need to take another direction

will mark as resolved once i have a tested commit

src/plugins/api/index.ts
scyto: You're right, i missed this in my testing, the leader's volume_muted flag and the all-members isGroupMuted() check could disagree. 

Fixed by removing playerCommandGroupMuteToggle (which used the leader flag) and having handlePlayerMuteToggle call playerCommandGroupVolumeMute(playerId, !isGroupMuted(player)) directly — so the toggle now uses the same logic as the icon.

will mark as resolved once i have a tested and pushed commit

src/components/VolumeControl.vue
scyto: and in case you were asking why its in both files:

VolumeControl.vue uses isGroupMuted(player) in the template to determine the mute icon and to disable the volume slider when the group is muted (lines 57 and 72)

VolumeBtn.vue uses isGroupMuted(player) inside the volumeIconComponent computed to pick the correct icon for the bottom bar (line 149)

in an attempt to think about this beter, i asked claude what the better way would be, is the below what you were getting at? @stvncode 

--
Thinking about it — VolumeBtn's volumeIconComponent computed is essentially duplicating the same icon-selection logic that getVolumeIconComponent in utils.ts already does. VolumeBtn could use that shared function instead of its own computed, which would look like:

```
const volumeIconComponent = computed(() => {
  const player = store.activePlayer;
  if (!player) return Volume2;
  return getVolumeIconComponent(player, displayVolume.value, isGroupMuted(player));
});
```
That eliminates the duplicated icon logic in VolumeBtn, but it still needs to import isGroupMuted — so it doesn't fully solve the reviewer's concern.

To go one step further, we could make getVolumeIconComponent automatically use isGroupMuted when no isMuted override is passed, instead of falling back to just player.volume_muted. 

That works because isGroupMuted on a non-group player (no group_members) already just returns player.volume_muted. Then neither VolumeControl nor VolumeBtn would need to import isGroupMuted for icon rendering at all — they'd just call getVolumeIconComponent(player, displayVolume).

VolumeControl would still need isGroupMuted for the slider disabled state though, so the import doesn't go away entirely there.

scyto:
@stvncode i used your feedback to help me think about how to push claude to help me be better, i hope this addresses what you were getting at, this approach seems simpler to me and removed 15 lines of duplicative and custom code from my original PR.

these commits have been tested on my live (patched) production system

### Commit 1: Deduplicate isGroupMuted and fix mute toggle logic

#### Problem: 
isGroupMuted was copy-pasted in two components (VolumeControl.vue as a function, VolumeBtn.vue as a computed), and playerCommandGroupMuteToggle in the API class used !leader.volume_muted to decide mute/unmute — which could disagree with the icon that used the all-members isGroupMuted check.

#### Fix:

Extracted isGroupMuted to a single shared function in helpers.ts
handlePlayerMuteToggle now calls api.playerCommandGroupVolumeMute(id, !isGroupMuted(player)) directly, so the toggle always matches the icon
Removed playerCommandGroupMuteToggle from the API class (it encoded the wrong logic)

### Commit 2: Make getVolumeIconComponent group-mute aware

#### Problem: 
Every caller of getVolumeIconComponent had to know about group mute and either pass isGroupMuted(player) explicitly or duplicate the icon-selection logic entirely (as VolumeBtn did with 15 lines of hand-rolled code). Player.vue didn't pass it at all — a bug.

#### Fix:

Changed getVolumeIconComponent in utils.ts to fall back to isGroupMuted(player) instead of just player.volume_muted when no explicit isMuted override is passed. This is backward-compatible since isGroupMuted returns !!player.volume_muted for non-group players.
VolumeControl.vue drops the third arg from its icon call
VolumeBtn.vue replaces 15 lines of duplicated icon logic with a single getVolumeIconComponent call
Player.vue gets correct group mute icons for free with no code change

stvncode:
@scyto Thanks for the changes, this looks better 😃 
Can i let you rebase your pr ? You have conflicts. For the code, i have nothing to say. Will test and approve after the rebase if everything's correct

scyto:
DO NOT MERGE THIS PR rebasing broke it horribly :-( working on it

scyto:
## Status update — group mute now fully working

After rebasing onto the latest upstream, group mute broke for protocol-synced players (e.g., Sendspin). The root cause was on the **server side**: `cmd_group_volume_mute` used `player.group_members` (raw property, empty for protocol-synced players) instead of `player.state.group_members` (computed state with protocol-translated members). The existing `cmd_group_volume` already used `player.state` correctly — this was an inconsistency in our function.

### Frontend changes in this push (commit 5f04cd70):
- Pass `isGroupMuted(player)` explicitly to `getVolumeIconComponent` so the icon reflects aggregate group mute state correctly
- Pass individual child mute state (`!!childPlayer.volume_muted`) for child player icons
- Skip members with null `volume_muted` in `isGroupMuted` (players that don't support mute control)
- Only disable mute button for non-group players when `mute_control == PLAYER_CONTROL_NONE`

### Server fix required:
- music-assistant/server#3205 — one-line fix: `player.group_members` → `player.state.group_members` in `cmd_group_volume_mute`

Both PRs are needed for full group mute functionality with protocol-synced players.

scyto:
@stvncode sorry about that churn some of the other work going on by marcel broke me when i rebased and it took a  while for me to get claude to figure it out

I built and tested a complete docker image `ghcr.io/scyto/ma-server-patched:group-mute` without the server change the child mutes worked but the the group mute didn't, i see the server change already got merged.

This was the final analysis

----
**ME: was protocol synced players the new upstream feature that broke us?**

Yes, exactly. The upstream protocol linking refactor changed the contract for how group members are accessed:

Before the refactor: player.group_members (raw) contained all members directly, so your original cmd_group_volume_mute code worked fine.

After the refactor: For protocol-synced players (like your Sendspin setup), raw group_members can be empty. The real member list now lives in player.state.group_members, which runs through __final_group_members to translate protocol player IDs to visible player IDs.

The upstream code was updated accordingly — cmd_group_volume already uses player.state.group_members. But since cmd_group_volume_mute was your new code in the PR, it didn't get that update during the refactor. After rebasing onto the refactored code, the mismatch surfaced.

**ME: will group mute still work for non-protocol synced players?**

Yes. The player.state.group_members property is a superset of player.group_members — it includes everything from the raw property plus protocol-translated members.

For non-protocol-synced players, the __final_group_members logic in [player.py](vscode-webview://0s19sir7doobald5p6357up7rsc266dfc8gjs6lptn824072e87v/music_assistant/models/player.py) essentially passes through the raw group_members unchanged (since there are no protocol IDs to translate). So player.state.group_members returns the same list as player.group_members.

This is also confirmed by the fact that cmd_group_volume (the volume slider for groups) has been using player.state.group_members all along and works for all player types — our fix just makes cmd_group_volume_mute consistent with that existing pattern.

scyto:
@stvncode sorry to bug you, hoping someone might get to this before the upstream changes so much we have conflicts again

scyto:
a conflict again, i will rebase, recompile and retest and fix

scyto:
conflict resolved, build tested on live system with 8 players

scyto:
Cool use of copilot!

marcelveldt:
So, I was discussing this with @stvncode the other day and we realized that we maybe want to add a new property to the player model for the group_volume status ? We already have a group_volume property which gets calculated by the backend, why not also add a group_muted status ?

It would be relatively cheap/easy to add that, and then both the frontend and other api consumers can benefit.
Want me to add that to the backend @scyto ?

scyto:
I think a server backend function/property that allows consistency via API irrespective of front end code always makes more sense for any product.

I guess its down to do you want to get feedback on the user experience first and then refactor or hold this PR and wait for the backend PR.

the key for me is the maintenance of the user experience on the front end

All my coding is done with claude, what i am good at is testing.  So happy to go with what you say.

And yes you making the new backend property means it will get done right first time :-)

marcelveldt:
OK, added that and implemented the frontend part in https://github.com/music-assistant/frontend/pull/1508

scyto:
> OK, added that and implemented the frontend part in #1508

thanks, i will test and see if all the functionality from this PR made it, if not will do a new PR for any remaining

--edit--
i will do a PR for the two misisng features
1. changing group volume should never unmute a member (leader or child)
2. muting a member should not block the ability to change volume setting of member

PR #1407, opened 2026-01-28
Don't force reload on all config changes

apophisnow:
This PR pairs with the one of the same name in the server repository. The goal is to enable skipping a full reload on every config change, and improving the UI to handle it gracefully.

- Add success toast "Settings saved successfully" after saving
- Disable save button until changes are made
- Use toast for error messages instead of alert()
- Remove navigation that incorrectly went to provider settings


marcelveldt (approved):
Looks good!

PR #1408, opened 2026-01-29
Fix save button disabled after OAuth actions complete

apophisnow:
When configuring providers that use OAuth (like Tidal), the OAuth flow sets tokens via actions which triggers a configEntries update. Previously, this would reset oldValues, making hasUnsavedChanges return false and disabling the save button.

Now oldValues is only captured on the first load, so action-triggered changes are still detected as unsaved changes.


marcelveldt (approved):
LGTM
Thanks, @apophisnow

PR #1426, opened 2026-02-02
Fix Refresh Item not updating tracks for provider playlists

sfortis:
## Summary

When using "Refresh Item" from the context menu on provider playlists (like Deezer's dynamic playlists), the metadata/image would update but the **track listing would not refresh**.

## Root Cause

Non-library items (provider playlists that aren't added to the library) don't trigger `MEDIA_ITEM_UPDATED` events from the backend when refreshed. The backend's `refresh_item` method only emits events for library items:

```python
if library_id is None:
    return media_item  # No event emitted for non-library items
```

The previous frontend implementation used `Object.assign()` to update the item in-place, which:
1. Mutated the object without changing its reference
2. Vue's watcher didn't reliably detect the deep mutation
3. Child components (track listings) never knew to refresh

## Fix

1. **Made `signalEvent()` public** in the API class to allow emitting synthetic events
2. **Emit synthetic `MEDIA_ITEM_UPDATED` event** from the context menu after refresh completes
3. **Added `refresh-on-parent-update` prop** to PlaylistDetails to auto-reload tracks when parent item changes

This ensures the same event flow happens regardless of whether the item is in the library or not.

## Testing

Tested on dev environment with Deezer provider playlists - working as expected.

**To verify:**
1. Browse to a provider playlist (e.g., Deezer recommendations)
2. Open playlist details to see tracks
3. Right-click → "Refresh Item"
4. Verify tracks refresh automatically (no manual browser refresh needed)

Suggested label: `bugfix`

### DIFF
--- a/src/plugins/api/index.ts
+++ b/src/plugins/api/index.ts
@@ -1842,7 +1842,6 @@ export class MusicAssistantApi {
 
   /**
    * Signal an event to all registered listeners.
-   * Can be used to emit synthetic events from the frontend.
    */
   public signalEvent(evt: MassEvent) {
     // signal event to all listeners
--- a/src/plugins/api/index.ts
+++ b/src/plugins/api/index.ts
@@ -1840,7 +1840,11 @@ export class MusicAssistantApi {
     });
   }
 
-  private signalEvent(evt: MassEvent) {
+  /**
+   * Signal an event to all registered listeners.
+   * Can be used to emit synthetic events from the frontend.
+   */
+  public signalEvent(evt: MassEvent) {
     // signal event to all listeners
     for (const listener of this.eventCallbacks) {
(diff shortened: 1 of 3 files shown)


src/plugins/api/index.ts
```diff
@@ -1840,7 +1840,11 @@ export class MusicAssistantApi {
     });
   }
 
-  private signalEvent(evt: MassEvent) {
+  /**
+   * Signal an event to all registered listeners.
+   * Can be used to emit synthetic events from the frontend.
```
marcelveldt:
```suggestion
```

sfortis: The suggestion has been applied

marcelveldt (approved):
It feels a bit strange to do it this way but on the other hand you added a proper comment so lets just do it this way

PR #1431, opened 2026-02-05
Remove bottom navigation and replace it by a sidebar like in HA

stvncode:


marcelveldt (approved):
Yeah, let's do this so we can collect feedback from user sin nightly how they think about it without the bottom navigation.

PR #1454, opened 2026-02-10
Add sound modes and player options

fmunkes:
This is an attempt add adding something to the frontend. Sound modes are easy, can just copy/ paste the sources approach.

I see if I can do something for the PlayerOptions, I might be able to simply add them in another category in the player settings. If it doesn't work out in the end, that's fine...


src/helpers/player_menu_items.ts
```diff
@@ -200,6 +200,33 @@ export const getPlayerMenuItems = (
...
+  );
+  if (selectableSoundModes.length > 0) {
+    menuItems.push({
+      label: "select_sound_mode",
+      labelArgs: [],
+      icon: "mdi-music-note-eighth",
+      subItems: selectableSoundModes
+        .map((s) => {
+          return {
+            label: s.name,
```
marcelveldt: shouldnt this be the translation_key ?

fmunkes:
Added something for player options too. I am aware, that this is in the "does the job" category, all these design things are rather knew to me, and I am not good at design in general. Screenshots:

Small menu in player view:
menu_1

Extra button, like dsp, in player settings:
player_menu

And the options themselves:
options

I realized, that the PlayerOptionEntry of the models should not have a type (which must be the same as the one of the PlayerOption themselves), and a value (or name), just one of these, like the ConfigValueOption - I would change that if you agree.
2026-02-14-142824-screenshot-region

src/views/settings/EditPlayerOptions.vue
```diff
@@ -0,0 +1,137 @@
...
+<script setup lang="ts">
+import api from "@/plugins/api";
+import {
+  EventType,
+  PlayerOption,
+  PlayerOptionType,
+  PlayerOptionValueType,
+} from "@/plugins/api/interfaces";
+import { onBeforeUnmount, computed, ref, watch, Ref } from "vue";
+import Slider from "@/components/ui/slider/Slider.vue";
```
stvncode: Unused import

src/views/settings/EditPlayerOptions.vue
```diff
@@ -0,0 +1,137 @@
...
+      let arrString: PlayerOption[] = [];
+      let arrSelect: PlayerOption[] = [];
+      let arrSensor: PlayerOption[] = [];
+
+      player.options.forEach((option) => {
+        if (option.read_only) {
+          arrSensor.push(option);
+        } else if (option.options && option.options.length > 0) {
+          arrSelect.push(option);
+        } else if (option.type == PlayerOptionType.STRING) {
```
stvncode: Could you use === for type-safe comparisons ?

src/views/settings/EditPlayerOptions.vue
```diff
@@ -0,0 +1,137 @@
...
+          arrString.push(option);
+        } else if (option.type == PlayerOptionType.BOOLEAN) {
+          arrBool.push(option);
+        } else {
+          arrNumber.push(option);
+        }
+      });
+
+      let arrConcat: PlayerOption[] = [];
```
stvncode: You don't need this. You can do something like this instead: 
`playerOptions.value = [
    ...arrBool.sort((a, b) => a.key.localeCompare(b.key)),
    ...arrNumber.sort((a, b) => a.key.localeCompare(b.key)),
    etc etc
  ];`

src/views/settings/PlayerOptionField.vue
```diff
@@ -0,0 +1,196 @@
...
+import { PlayerOptionType } from "@/plugins/api/interfaces";
+import { $t } from "@/plugins/i18n";
+import { computed, ref } from "vue";
+
+const props = defineProps<{
+  playerOption: PlayerOption;
+  playerId: string;
+}>();
+
+const isThumbHidden = ref(true);
```
stvncode: Unused variable

src/views/settings/PlayerOptionField.vue
```diff
@@ -0,0 +1,196 @@
...
+  key: string,
+  value: PlayerOptionValueType | PlayerOptionValueType[] | null | undefined,
+) => {
+  if (!props.playerId || value === null || value == undefined) return;
+  // we only have single numbers
+  if (Array.isArray(value)) value = value[0];
+  try {
+    await api.playerCommandSetOption(props.playerId, key, value);
+  } catch (error) {
+    console.error("Error setting option:", key, value);
```
stvncode: Could you use a toast error in case something goes wrong (toast.error(""))

src/views/settings/PlayerOptionField.vue
```diff
@@ -0,0 +1,196 @@
...
+      title: optName,
+      value: orgOption.value,
+    };
+    options.push(option);
+  }
+  return options;
+});
+</script>
+<style>
+.optionGrid {
```
stvncode: Looks like your css is unused

src/helpers/player_menu_items.ts
```diff
@@ -200,6 +200,33 @@ export const getPlayerMenuItems = (
     });
   }
 
+  // add 'select sound mode' menu item
+  const selectableSoundModes = player.sound_mode_list.filter(
+    (s) => !s.passive || s.id == player.active_sound_mode,
```
stvncode, line 217: Use strict equality === for all the == in this function

stvncode (commented):
Regarding the code, only minor issue. Otherwise it looks good 😃

src/views/settings/EditPlayerOptions.vue
fmunkes: Removed them - could we add
```json
"compilerOptions": {
        ...
        "noUnsedParameters": true,
        "noUnusedLocals": true
}
```
these to our tsconfig.json? At least for me an error is hard to ignore in my editor config :)

src/views/settings/EditPlayerOptions.vue
fmunkes: changed everywhere

src/views/settings/EditPlayerOptions.vue
fmunkes: Ah, I see, changed, but I added an .flat() - or can I avoid this too?

src/views/settings/PlayerOptionField.vue
fmunkes: removed.

fmunkes:
> Regarding the code, only minor issue. Otherwise it looks good 😃

Thanks a lot for the review. I (hopefully) addressed all your feedback. Regarding the remark I made on the python code, I'll create some new PR if needed at the end of this week. Should we then remove something I'd of course come back for a new frontend PR  :)

src/views/settings/EditPlayerOptions.vue
stvncode: Yeah definitively, can I let you add it pls ? :)

src/views/settings/EditPlayerOptions.vue
stvncode: Yeah, using flat means that you didn't add the spread (...). Just do `playerOptions.value = [
  ...arrBool.sort((a, b) => a.key.localeCompare(b.key)),
  ...arrNumber.sort((a, b) => a.key.localeCompare(b.key)),
  ...arrString.sort((a, b) => a.key.localeCompare(b.key)),
  ...arrSelect.sort((a, b) => a.key.localeCompare(b.key)),
  ...arrSensor.sort((a, b) => a.key.localeCompare(b.key)),
]`

src/views/settings/EditPlayerOptions.vue
fmunkes: sure - and done :) - but this causes a lot of lint errors now, which we could fix. let me know, if you'd like to keep it that way.

fmunkes: Ok, my suggestion would be, that I revert that tonight, such that this PR can be merged. I'll make a follow up PR for TS config change, where I adapt all these files.

stvncode: Oh yeah that would be lovely. Didn't know we have that much unused elements sorry 😢

fmunkes: Was able to do that in the web interface :)

fmunkes:
I just resolved to new merge conflict - @stvncode is there anything else you would like to do me on this PR? :)

src/views/settings/EditPlayer.vue
```diff
@@ -279,6 +279,17 @@ const config_entries = computed(() => {
...
   }
+  // add player options should they exist
+  if (player && player.options.length > 0) {
+    entries.push({
+      key: "player_options",
+      type: ConfigEntryType.OPTIONS,
+      label: "",
+      default_value: "",
+      required: false,
+      category: "options",
```
marcelveldt, line 290:
```suggestion
      category: "options",
      injected: true,
```

marcelveldt: I think this now needs to be so to "injected"

fmunkes: I verified that the options still appear as they should, thanks!

marcelveldt (approved):
Nice work @fmunkes !

PR #1456, opened 2026-02-12
Save current queue to playlist feature

chrisuthe:
Adding UI to save queue as playlist. This does use the dialog pieces not a prompt() for the name like the base create playlist does. I can refactor with that if needed. I'm unsure of the desire to have the confrimation with ability to open the playlist as well, that could obviously be removed completely if preferred. This relies on https://github.com/music-assistant/server/pull/3149.
createplaylistfromQueue


chrisuthe:
This should be testable now that the pre-req merged. Unsure if it needs the backend change label or not in that case.

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
```diff
@@ -510,12 +510,80 @@
         </div>
       </div>
     </v-card>
+
+    <!-- Save queue as playlist dialog -->
+    <Dialog
```
stvncode: Could you create another file for this ? PlayerFullscreen file is starting to be really big and for visibility purposes, it would be nice if we can put the content in another file 🙏

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
```diff
@@ -1243,6 +1312,53 @@ onMounted(() => {
...
+  if (!saveQueuePlaylistName.value) return;
+  showSaveQueueDialog.value = false;
+  try {
+    createdPlaylist.value = await api.queueCommandSaveAsPlaylist(
+      saveQueueQueueId.value,
+      saveQueuePlaylistName.value,
+    );
+    showPlaylistCreatedDialog.value = true;
+  } catch (e) {
+    alert(e);
```
stvncode: For error, we are using toast.alert now

src/helpers/player_menu_items.ts
```diff
@@ -172,6 +173,17 @@ export const getPlayerMenuItems = (
       icon: "mdi-cancel",
     });
   }
+  // add 'save queue as playlist' menu item
+  if (playerQueue?.items) {
```
stvncode: We should maybe put playerQueue?.items?.length > 0 instead of just playerQueue?.items to avoid showing the menu when the queue is empty ?

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
```diff
@@ -510,12 +510,80 @@
...
+            @click="doSaveQueueAsPlaylist"
+          >
+            {{ $t("settings.save") }}
+          </UiButton>
+        </DialogFooter>
+      </DialogContent>
+    </Dialog>
+
+    <!-- Playlist created dialog -->
+    <Dialog
```
stvncode: This is a bad behavior imo. What we could do is do a toast.success after saving to playlist and put a link to the playlist there. An example can be found here: https://vue-sonner.vercel.app/, in types -> action : 

Image

Need to test if it's good but could be: 
`toast.succes(t("playlist_created"), {
  action: {
    label: 'See playlist',
    onClick: openCreatedPlaylist
  },
})`

stvncode (commented):
Small comments. Thanks for your work 🙏

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
chrisuthe: Yes, this is much better, I agree.

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
chrisuthe: https://github.com/music-assistant/frontend/pull/1456/commits/228f5e1eefd50cb2604b5d304dc48923817cfec0 for this request!

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
chrisuthe: https://github.com/music-assistant/frontend/pull/1456/commits/1acbbcecf6aaef1b0ef7c0fecc898e62e301dbe5 shifts to this pattern, good call!

src/helpers/player_menu_items.ts
chrisuthe: Done!

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
chrisuthe: So, I took this a step further, created a reusable view and added it to the "Create playlist" area as it was still using the old console input. https://github.com/music-assistant/frontend/pull/1456/commits/29d2491a1527a46c7e7822e59e302a0d1b1840f6

marcelveldt (approved):
Nice work, @chrisuthe !

PR #1458, opened 2026-02-13
[Bug] Fix player settings filters lost on save

chrisuthe:
Use router.back() instead of router.push() in onSubmit and enablePlayer so URL query params (type/provider filters) are preserved when returning to the player list.


stvncode (approved):
Thanks, good catch 😃

PR #1470, opened 2026-02-17
Implement Feature/Enhancement: Copy Album or Artist name from page title

SimeonAT:
Implementation for Feature/Enhancement that is described in GitHub Discussion https://github.com/music-assistant/support/discussions/4726. 

GitHub Discussion #4726 asked for the ability for users to highlight and copy-paste the title of an album, artist, or playlist. I have added this enhancement, taking care to minimize the amount of changes I made in the front-end source code when doing so.

The below images demonstrate that this change now enables the title an album, artist, or playlist can be highlighted in both mobile and desktop.

**Desktop:**
copy playlist title in desktop

**Mobile:**
copy playlist title in mobile


src/components/InfoHeader.vue
```diff
@@ -519,6 +521,19 @@ const artistLogo = computed(() => {
 </script>
 
 <style scoped>
+.selectable {
+  -webkit-user-select: text;
+  /* Safari */
+  -khtml-user-select: text;
```
stvncode, line 527: Do we really need Konqueror support ?

stvncode (approved):
Good job 😃

src/components/InfoHeader.vue
SimeonAT: I added that line as [`user-select: none` is also set for Konqueror in `global.css`](https://github.com/music-assistant/frontend/blob/main/src/styles/global.css#L54). My goal was for `.selectable` to be the exact inverse of the `user-select` properties in `global.css`, in order to ensure that it worked across all browsers.

PR #1474, opened 2026-02-18
Show ADD TO PLAYLIST for podcast episodes and audiobooks

OzGav:
Extend the playlist functionality (originally added for radio stations in PR #1356) to also support podcast episodes, and audiobooks. Like radio stations, these items can only be added to playlists stored in the builtin provider.

I have added the WAITING ON BACKEND label as although the backend has been done I discovered a bug while working on this so want to get that fixed before we push this through.


src/layouts/default/ItemContextMenu.vue
```diff
@@ -839,12 +839,14 @@ export const getContextMenuItems = async function (
...
   }
 
-  // remove from playlist (playlist tracks and radio items)
+  // remove from playlist (playlist tracks, radio, podcast, podcast episode, and audiobook items)
   if (parentItem && parentItem.media_type === MediaType.PLAYLIST) {
     const playlist = parentItem as Playlist;
     if (
       (items[0].media_type === MediaType.TRACK ||
-        items[0].media_type === MediaType.RADIO) &&
+        items[0].media_type === MediaType.RADIO ||
```
stvncode: Are we doing a check if items.length > 0 before ? Imagine items is empty, we will have a runtime error

stvncode (dismissed):

src/layouts/default/ItemContextMenu.vue
OzGav: Took me a bit but WAY back at line 575-577 there's an early return. I also noticed the const firstitem there that could be used throughout to make it clearer that we are working with a non-null variable... I'll leave it to you if you want a global replace done or just leave it

stvncode (approved):
Thanks a lot @OzGav 🙏

PR #1475, opened 2026-02-19
Moved all Unit Tests to `tests` Directory

SimeonAT:
I had a discussion with @stvncode in the Music Assistant Discord, and he suggested that all unit tests should be organized in a `test` directory. In this pull request, I moved all the unit tests to a newly created `test` directory. 

I only needed to move all the test files in `composable` and `helpers`, and changed the imports in each test file based on their new locations. I didn't have to do any changes to the Vitest configuration -- even with these changes, the pre-commit was able to run the unit tests without any issues.


stvncode (approved):
Thanks 🙏

PR #1476, opened 2026-02-20
Add UI to select playback speed

andykelk:
Add a UI component `PlaybackSpeedBtn` which appears on the full player screen. This brings up a menu of different speeds which is then passed back to the server. Requires that the server handles `player_queues/set_playback_speed`.

`PlaybackSpeedBtn.vue` reads `playerQueue.extra_attributes.playback_speed` and calls the new API command on selection. The button highlights when speed ≠ 1×.

Related: https://github.com/orgs/music-assistant/discussions/3790

screenshot_20260220_123350

screenshot_20260220_123330

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerControlBtn/PlaybackSpeedBtn.vue
+++ b/src/layouts/default/PlayerOSD/PlayerControlBtn/PlaybackSpeedBtn.vue
@@ -1,0 +1,76 @@
+<template>
+  <v-menu
+    v-if="isVisible && playerQueue"
+    location="top"
+    :close-on-content-click="true"
+  >
+    <template #activator="{ props: menu }">
+      <Button
+        v-bind="{ ...menu }"
+        variant="icon"
+        :title="$t('playback_speed')"
+        :disabled="!playerQueue.active || playerQueue.items === 0"
+        :style="
+          currentSpeed !== 1.0 ? { color: 'rgb(var(--v-theme-primary))' } : {}
+        "
+      >
+        <span class="speed-label">{{ formatSpeed(currentSpeed) }}</span>
+      </Button>
+    </template>
+
+    <v-list density="compact" nav>
+      <v-list-subheader>{{ $t("playback_speed") }}</v-list-subheader>
+      <v-list-item
+        v-for="speed in SPEED_PRESETS"
+        :key="speed"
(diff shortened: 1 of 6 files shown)


marcelveldt:
Since this is a rather advanced setting and the amount of buttons is already overwhelming, could we consider adding it to the contextmeny instead ?

@stvncode what do you think ?

andykelk:
> Since this is a rather advanced setting and the amount of buttons is already overwhelming, could we consider adding it to the contextmeny instead ?
> 
> @stvncode what do you think ?

An alternative would be to dynamically display for speech content (audiobooks, podcast, podcast episode) only.

i.e. always have it in the context menu, but promote it to the main display if a speech item is playing.

image
image

stvncode:
@marcelveldt, totally agree on the position. We have too much buttons for now (and this page will be refacto soon as well). We should add it into the context menu for now and later, if design people find a way too show it in a better place, will do 👍

marcelveldt:
I also had to make changes to the backend implementation for this as it caused some issues.

The filter was applied to the wrong section of the code.
Instead it will now be applied to a queue item only.

src/layouts/default/PlayerOSD/PlayerControlBtn/PlaybackSpeedBtn.vue
```diff
@@ -0,0 +1,76 @@
...
+  activeColor?: string;
+}
+
+const props = withDefaults(defineProps<Props>(), {
+  isVisible: true,
+  activeColor: "rgb(var(--v-theme-primary))",
+});
+
+const currentSpeed = computed<number>(() => {
+  const speed = props.playerQueue?.extra_attributes?.playback_speed;
```
marcelveldt, line 54: so the playback_speed will now be set on the (current) queue item instead and not at queue level

marcelveldt: See here for details:
https://github.com/music-assistant/server/commit/90f369da23027c5a880d7a48854f214757e31156

stvncode:
@andykelk, any update on this ? 😃

OzGav:
Thanks for the contribution but this has been superseded by https://github.com/music-assistant/frontend/pull/1787

PR #1479, opened 2026-02-23
Enable VSCode Intellisense on Unit Tests Files and Use NVM versions of Node.js in Husky Pre-commit

SimeonAT:
This pull request makes two small modifications to the front-end unit testing setup:

1. The `include` in `tsconfig.json` is changed so that VSCode Intellisense and syntax highlighting are enabled when viewing all unit test files in `test/*`.

2. Husky pre-commit is modified to use NVM versions of Node.js and `npm` if they are found. If NVM is used, the Node.js and `npm` versions used by NVM is printed to the console:
    image

    I have Yarn, along with all Music Assistant front-end dependencies, installed on my NVM versions of Node.js and `npm`. As a result, I initially had some trouble running the Husky pre-commit script as my `PATH` versions of Node.js were outdated. I believe that this fix could be helpful for contributors who use NVM to manage their Node.js versions.


.husky/pre-commit
```diff
@@ -3,5 +3,15 @@
 # Ensure yarn is available in PATH
 export PATH="/usr/local/bin:/opt/homebrew/bin:$PATH"
 
+# Use default NVM versions of Node.js and `npm` if yarn is installed there.
+if [ -d "$HOME/.nvm" ]; then
```
stvncode: Husky hooks should be minimal, just run lint and tests. Environment setup belongs in the developer's dotfiles. NVM is just one version manager, others aren't handled, so this only solves the problem for a subset of users.

This might be better to add a .nvmrc to the repo but I'll handle it later. Can you remove this part please ?

tsconfig.json
```diff
@@ -30,8 +30,8 @@
         "src/**/*.tsx",
         "src/**/*.d.ts",
         "src/**/*.vue",
-        "tests/**/*.ts",
-        "tests/**/*.tsx"
+        "test/**/*.ts",
```
stvncode: Also, why the folder is not tests and test. Since they are many test in it we should use tests and not test. So this is good imo

stvncode (changes requested):

tsconfig.json
SimeonAT: I initially named it `test` and not `tests` to follow the convention used in [the examples shown in the Vitest GitHub repository](https://github.com/vitest-dev/vitest/tree/main/examples). I'll change the directory name from `test` to `tests`.

.husky/pre-commit
SimeonAT: I just reverted the Husky pre-commit to what it was before. 👍

stvncode (approved):
Thanks 🙏

PR #1494, opened 2026-02-24
Feat/genre-merge

jozefKruszynski:
Add genre merge functionality, fix infoheader title for genres, use eventbus for delete
Also allow filtering of empty genres, with hide empty being default


stvncode (approved):
Nice 🥳

PR #1499, opened 2026-02-25
Disable podcasts and audiobooks in main menu when library is empty

OzGav:
Previously all library menu items (artists, albums, tracks, playlists, radios, genres, podcasts, audiobooks) were hidden when their count was 0. Now only podcasts and audiobooks are hidden when empty — all other items are always shown in the menu.

Discussed here https://discord.com/channels/753947050995089438/1476323267513028714

### DIFF
--- a/src/components/navigation/AppSidebar.vue
+++ b/src/components/navigation/AppSidebar.vue
@@ -28,6 +28,7 @@ const navItems = computed(() => {
       title: t(item.label),
       url: item.path,
       icon: item.icon,
+      disabled: item.disabled,
     }));
 });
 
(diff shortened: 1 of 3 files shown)


marcelveldt:
Better idea: we disable them if they don't have items ?
The whole entry can be removed from the menu

OzGav:
Are you saying have them there greyed out so the user is prompted to go to the settings to then remove them if they wish?

marcelveldt:
Yeah... does that make sense ?

OzGav:
Yes I think so. Let me look at that.

OzGav:
OK this has been implemented. Tested on dev server as working

marcelveldt (approved):
LGTM
Thanks, @OzGav !

PR #1500, opened 2026-02-26
Fix issue "checkbox are not aligned with text when multi-selecting"

SimeonAT:
Pull request fixes the issue where the checkboxes are not aligned with the title of a list of genres or songs when the multi-select UI is enabled. This issue is discussed in https://github.com/orgs/music-assistant/projects/2/views/1?pane=issue&itemId=157687149. 

**On Desktop:**
image

**On Mobile:**
IMG_5286 Medium


src/components/ListviewItem.vue
```diff
@@ -8,52 +8,37 @@
...
-          />
-        </template>
-        <span>{{ $t("tooltip.explicit") }}</span>
-      </v-tooltip>
+    <template v-if="!showCheckboxes" #title>
+      <ListviewitemTitle
+        :display-name="displayName"
+        :item="item"
+        :show-checkboxes="showCheckboxes"
+      />
```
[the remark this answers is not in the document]
SimeonAT: Resolved in commit https://github.com/SimeonAT/music-assistant-frontend/commit/213bbaf80aef1873298ef879e62e3a5d29aa490e.

src/components/ListviewItemTitle.vue
```diff
@@ -0,0 +1,78 @@
...
+      v-if="item && item.metadata && parseBool(item.metadata.explicit || false)"
+    >
+      <TooltipProvider>
+        <Tooltip>
+          <TooltipTrigger>
+            <v-icon
+              :class="{ 'explicit-icon-margin-left': showCheckboxes }"
+              icon="mdi-alpha-e-box"
+              width="35"
+            />
```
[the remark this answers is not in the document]
SimeonAT: I don't believe this change needs to be made. When trying it out on my end, clicking the explicit icon of a song (when it appears) does *not* toggle the checkbox. Only clicking the checkbox or the song name will toggle the checkbox.

src/components/ListviewItem.vue
```diff
@@ -8,52 +8,37 @@
...
     </template>
```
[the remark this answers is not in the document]
SimeonAT: In theory, the fact that the subtitle moves into the title slot could be a potential issue. However, when looking at it visually, I don't think it is that much of a problem.

**On Desktop:**
image

**On Mobile:**
IMG_5285 Medium

It is true that when checkboxes are shown, the checkbox + title and subtitle are slightly unaligned. I talked to @stvncode about this on Discord -- he told me that the alignment doesn't need to be perfect, and that this alignment is an already an improvement. 

Nonetheless, I adjusted the margin bottom in commit https://github.com/SimeonAT/music-assistant-frontend/commit/204f9130f6e45a3c3145f437d3b23ff7facf030f.

SimeonAT:
Regarding Co-Pilot's comments that were suppressed due to low confidence:
 
>   `compProps` is declared with `withDefaults(defineProps(...))` but never used; the template accesses `displayName/item/showCheckboxes` directly. Consider either using `compProps` consistently (to apply defaults) or removing it to avoid dead code and confusion.
> 
> 
> ```
> const compProps = withDefaults(defineProps<Props>(), {
>   displayName: "",
>   showCheckboxes: false,
> });
> ```

I added default values for `isPlaying` as well in commit https://github.com/SimeonAT/music-assistant-frontend/commit/b0a894b84fdac77d87b88a8bc09ea2adb9aa9300.  

Following the convention in `ListviewItem.vue`, I did not add a default value for the `item` prop in `compProps`.

>  Component naming is inconsistent here: `ListviewitemTitle` uses a lower-case `i` in `item`, while the file is `ListviewItemTitle.vue` and other component imports in this file use PascalCase. Rename the import/local component name (and the template tag) to `ListviewItemTitle` for consistency and readability.
> 
> 
> ```
> import ListviewitemTitle from "./ListviewItemTitle.vue";
> import { Checkbox } from "@/components/ui/checkbox";
> ```

I don't believe it is necessary to change the name of the component. The `ListviewItemTitle` is a subcomponent of `ListviewItem`, which uses a lowercase "v" for "view" and capital "I" for "Item".

src/components/ListviewItem.vue
```diff
@@ -8,52 +8,39 @@
...
           @update:model-value="
-            (x: boolean | null) => {
-              if (x != null) emit('select', item, x);
+            (x: boolean | 'indeterminate') => {
+              if (x != 'indeterminate') emit('select', item, x);
             }
           "
         />
+        <label :for="`listviewitem-checkbox-${item.item_id}`">
+          <ListviewitemTitle
```
stvncode: Never put a div inside a label (ListviewitemTitle contain a div at the root), this is invalid HTML. You can change it from <div> to <span>. This would also fix the checkbox-label font-weight scoping since <span> is inline, not a block element.

src/components/ListviewItemTitle.vue
```diff
@@ -0,0 +1,84 @@
...
+  displayName: string;
+  item: MediaItemType;
+  showCheckboxes: boolean;
+  isPlaying: boolean;
+}
+
+// global refs
+const { t } = useI18n();
+
+const compProps = withDefaults(defineProps<Props>(), {
```
stvncode: compProps is declared but never used. Remove it pls

src/components/ListviewItem.vue
stvncode: Should be ListviewItemTitle and not ListviewitemTitle

stvncode (commented):
Also, you don't have to do many commits for a small feature like this. Just do one please, it's not easy for us to have that many commits (you did 33 commits btw for a pr that even with review should have 4-5 commits max)

src/components/ListviewItem.vue
SimeonAT: Renamed components to `ListviewItemTitle` in commit https://github.com/SimeonAT/music-assistant-frontend/commit/98f1fa9795a87aaf71f2a63df3edebb1691f3c07.

src/components/ListviewItemTitle.vue
SimeonAT: Removed the use of `compProps` in commit https://github.com/SimeonAT/music-assistant-frontend/commit/98f1fa9795a87aaf71f2a63df3edebb1691f3c07.

src/components/ListviewItem.vue
SimeonAT: To fix this issue in https://github.com/SimeonAT/music-assistant-frontend/commit/98f1fa9795a87aaf71f2a63df3edebb1691f3c07, I removed the topmost `div` of `ListviewitemTitle`, and applied the `checkbox-label` CSS class to the `span` elements directly.

src/components/ListviewItemTitle.vue
```diff
@@ -0,0 +1,80 @@
...
+    </span>
+  </span>
+  <!-- explicit icon -->
+  <template
+    v-if="item && item.metadata && parseBool(item.metadata.explicit || false)"
+  >
+    <TooltipProvider>
+      <Tooltip>
+        <TooltipTrigger>
+          <v-icon
```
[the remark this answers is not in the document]
stvncode: Just need to change that and we are good to go 😃

MarvinSchenkel:
Hey @SimeonAT , thanks for your efforts!

Just had a quick look and this experience is not nice. We should maintain the same space as the icon, now the text jumps back and forth. Also, the font seems to change when we start multi selecting, this should remain the same. Here is a video:

https://github.com/user-attachments/assets/b0c4387a-709c-48b7-8d47-726473a834b2

SimeonAT:
Hey @MarvinSchenkel! Thank you very much for your feedback. I will get right to work on fixing the space and font issues, and I'll let you know as soon as I fix them!

src/components/ListviewItemTitle.vue
```diff
@@ -0,0 +1,86 @@
...
+      ({{ new Date(item.metadata.release_date).getFullYear() }})
+    </span>
+  </span>
+  <!-- explicit icon -->
+  <template
+    v-if="item && item.metadata && parseBool(item.metadata.explicit || false)"
+  >
+    <Tooltip>
+      <TooltipTrigger as-child>
+        <span :aria-label="t('tooltip.explicit')" role="button" tabindex="0">
```
stvncode: role="button" on a non-interactive tooltip trigger is semantically incorrect, you can remove it

src/components/ListviewItemTitle.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <span
+    v-if="item.media_type == MediaType.FOLDER"
+    class="v-list-item-title"
```
stvncode, line 4: class="v-list-item-title" hardcodes a Vuetify internal class name outside of a Vuetify component context. This is fragile and could break on Vuetify updates.

stvncode (commented):
Almost good. Would it be possible to align the checkboxes with the icon at the top. There is a padding: 7px !important. Just remove the padding right 0px !important and do: ```padding: 7px 0 !important``` so both right and left have 0

Image

stvncode (approved):
I made the changes myself. Good to go now

PR #1501, opened 2026-02-26
Add manual media item to genre linking

jozefKruszynski:
Add components to allow linking media items manually to genres.


src/layouts/default/ItemContextMenu.vue
```diff
@@ -983,6 +983,20 @@ export const getContextMenuItems = async function (
       icon: "mdi-link",
     });
   }
+  // link to genre (any non-genre media item)
```
marcelveldt: should this be limited to library items only ?

jozefKruszynski: That would be very sensible.

stvncode (commented):
Code looks good, will let copilot review as well in case i missed something

stvncode (approved):
LGTM

PR #1503, opened 2026-02-27
Add missing genre route definitions and filters

jozefKruszynski:
Podcast and Audiobook route definitions were missing for redirects to library pages, as were filters from said library pages.


marcelveldt (approved):
Thanks, @jozefKruszynski !

PR #1505, opened 2026-02-28
Rename 'home' to 'discover' across the entire app

radiohe4d:
### Summary
After the new changes to the bottom navigation, the 'Home' button has been renamed to 'Discover'.

This PR updates all references, URLs & translations of 'Home' to 'Discover' across the entire app.

### Notes
* I have done my best to update the translations but they may not be accurate for those I had to manually translate.
* I have added a URL redirect from '/home' to '/discover' as I was worried anyone with bookmarks would get blank screens after this update.
* There are components and methods that still reference 'home' in their names. I wasn't sure if I should make those changes all in this PR or do them separately, let me know which and I can sort those out too.

### Screenshots
Screenshots showing the updated URL, page title and menu items updated to 'Discover'.

Screenshot From 2026-02-28 20-44-14
Screenshot From 2026-02-28 20-43-43


src/translations/cs.json
marcelveldt, line 1: So, normally you don't have to do the translations as this is handled through Lokalize.
Only the en.json is sufficient. But nice work anyways!

radiohe4d: Ah my mistake, I'll look that up to see how that works in future, thanks.

marcelveldt (approved):
Looks good, thanks @radiohe4d !

PR #1509, opened 2026-03-01
Improve player popover responsive layout + refactor event handling

radiohe4d:
### Summary
The player popover tip arrow no longer lines up with the players button on the bottom controls both on desktop and mobile.

This change makes the arrow point as close to the players button as close as possible at all horizontal resolutions (within reason).

### Notes
* The global event handlers did not seem to be needed as the popover was using @focus-outside, @pointer-down-outside and @escape-key-down. This handler was also always being triggered even after the tip was closed.
* The local 'tipWasShown' was redundant as the 'store.playerTipShown' can handle this from a central place.
* I've removed the check to not show the tip when (store.activePlayerId === webPlayer.player_id) because it relies on a race condition. The webPlayer object isn't always populated by the time the tip is loaded, so it makes this check unreliable.
* The click handler on the BottomNavigation wasn't needed here as it can be handled internally on line 14 of ActivePlayerPopover.

### Videos

Before the change:

https://github.com/user-attachments/assets/0bc68505-ff3c-4aa9-b60c-4595bca89dc1

After the change:

https://github.com/user-attachments/assets/3e126750-ba53-4f79-bb26-d8019b583fe9

### DIFF
--- a/src/components/ActivePlayerPopover.vue
+++ b/src/components/ActivePlayerPopover.vue
@@ -194,25 +194,31 @@ watch(
   left: 50%;
 }
 
-@media screen and (max-width: 361px) {
+@media screen and (max-width: 313px) {
   .player-tip-end .tip-arrow {
-    left: 155px;
+    left: 176px;
   }
 }
 
-@media screen and (min-width: 360px) and (max-width: 406px) {
+@media screen and (min-width: 312px) and (max-width: 411px) {
   .player-tip-end .tip-arrow {
-    left: 148px;
+    left: 167px;
   }
 }
 
-@media screen and (min-width: 405px) and (max-width: 471px) {
+@media screen and (min-width: 410px) and (max-width: 511px) {
   .player-tip-end .tip-arrow {
-    left: 138px;
+    left: 154px;
   }
 }
 
-@media screen and (min-width: 470px) and (max-width: 770px) {
+@media screen and (min-width: 510px) and (max-width: 621px) {
+  .player-tip-end .tip-arrow {
+    left: 141px;
+  }
+}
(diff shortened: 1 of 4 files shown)


marcelveldt:
Did you, by any chance also fix the weird border around the players buton when you dismiss it ?

image

radiohe4d:
I wasn't sure if that was intended or not. I can fix that too.

radiohe4d:
Also, with the addition of the Search button in the BottomNavigation, the alignment is wrong again. Ill have to readjust that too.

May need a better way to track the location of the Players button as this will continue to be an issue.

radiohe4d:
@marcelveldt ~~It seems the weird border issue has resolved itself.~~ The popover has default behaviour that puts focus on the next element (the Player button in this case) and it doesn't appear to have a way to disable it. I've had to select the button by ID and manually blur it. A bit hacky, so if anyone has other fixes it would be great to change this,

 I've updated the media-queries for the BottomNav for now until I can find a better solution.

marcelveldt:
I think this is fine for now. A follow up approach could be that we just drop the popover completely and look for another way to draw the attention to the user what player is the active one at startup.

Spotify for example on desktop has a big green bar telling you the active player (which is also a bit overkill imo) but on mobile its more subtle and there they just highlight the active player in bright green.

What we could potentially do is have the playername just blink a couple of times at startup in the blue color ?

Anyways, I'll merge this as-is because it is an improvement to what we have now.
If you come up with a better solution such as my idea above, feel free!

marcelveldt (approved):
Thanks, @radiohe4d !

PR #1511, opened 2026-03-01
Fix refresh on new version and replace it with a toast

stvncode:
@balloob fyi, this fix the refresh

Screenshot 2026-03-01 at 13 27 03


marcelveldt (approved):
Thanks, @stvncode !

balloob:
Noooo, I hate this toast with passion on the HA side. Let's not add it here.

balloob:
If we have stale assets, how come the initial screen loaded? how does the caching work?

marcelveldt:
It is your browser doing the caching and it should fetch new version at startup. The PWA has a service worker to do so but is hosted in HTTPS, the component in HA doesn't, especially not since it's in HTTP context. 

Since you run nightly, you have a new version of the frontend almost every day

marcelveldt:
@stvncode just had the toaster and clicking it did not refresh the page, the toaster just kept coming back.
Maybe we can do something else to prevent the issue but I have no idea what. In the end it's the browser that still has an outdated version of the javascript and should be told to refresh the page.

marcelveldt:
Background info: https://medium.com/mobileaction-tech/vue-spa-chunk-load-error-causes-and-solutions-2f86c02c761f

That service worker works correctly for the PWA but it does not work (apparently) when accessing the page using HTTP

PR #1512, opened 2026-03-01
Fix "This player is disabled" banner flash

apophisnow:
Fixes a bug where clicking a player in settings briefly shows "This player is disabled" before loading the settings page.


marcelveldt (approved):
Thanks @apophisnow !

PR #1513, opened 2026-03-02
Allow access to the BottomNavigation when the PlayerMenu is open

radiohe4d:
### Summary
When in mobile view and the player menu is open, it is difficult to navigate away and reaching the close button on the top on mobile is cumbersome.

This makes the BottomNavigation accessible on mobile view when the PlayerMenu is open.

Tapping the 'Players' button then toggles the menu.

### Notes
* Updated the player menu open transition to match the main app menu with both speed and curve.
* I've moved the buttons outside of the ActivePlayerPopover as it was triggering the popover when the buttons were pressed. This is not expected functionality.
* The player tip popover surfaced more bugs when making this change, so I've done more refactoring to fix and remove redundant code.

### Videos

https://github.com/user-attachments/assets/5d9c936a-cdfd-4b0f-9c62-116f0e501b2f


marcelveldt (approved):
Nice work, @radiohe4d !

PR #1515, opened 2026-03-02
Temporary fix so the user doesnt get stuck on the players menu

radiohe4d:
I've introduced this bug in the earlier PR, this is a temporary fix so the user doesn't get stuck on the players menu when navigating from the fullscreen player.


marcelveldt (approved):
Thanks, @radiohe4d !

PR #1525, opened 2026-03-02
Rounded style for overlayed menus

marcelveldt:
I personally found the overlayed menus (players on the right and the menu on the left) a bit clunky on mobile because the screen gets cutoff so I added a bit of a rounded styling so these panels look more like popouts from the side.

also adjusted the navigation menu on mobile to also hover above the footer menu.
This way the footer menu is simply always present on mobile and you can toggle these menus.

image

image


marcelveldt:
@stvncode @radiohe4d FYI

radiohe4d:
I really like this, more consistent functionally and looks great with the rounded corners.

MarvinSchenkel (approved):
🎉

PR #1529, opened 2026-03-03
Align the players controls so they are in more consistent positions

radiohe4d:
### Summary
Due to the order of the player control buttons and that some are optionally shown, the play/pause button doesn't always exist in a consistent position within the player card. 

I have also moved the power button so the group members button is always beside the play/pause button.

### Notes
N/A

### Screenshots

Before:
player_menu_align_controls_before

After:
player_menu_align_controls_after


stvncode (approved):
Nice thanks 🙏

PR #1539, opened 2026-03-05
Optimize NowPlaying badge animation to minimize CPU usage

apophisnow:
Fixes an issue with NowPlaying badge animation that was causing high CPU usage.


MarvinSchenkel (approved):
Thanks @apophisnow 🙏

PR #1540, opened 2026-03-06
Optimize more animations

apophisnow:
This optimizes more animations using transition and opacity where possible for more GPU vs. CPU.

Also smooths the progress bar so it runs per frame instead of 500ms tick.


MarvinSchenkel (approved):
Thanks @apophisnow !

PR #1542, opened 2026-03-06
Reduce render cost from scrollbar-color and slider transitions

apophisnow:
Two sources of unnecessary rendering overhead during playback:

1. scrollbar-color on * selector: Chromium attempted to animate scrollbar-color on every DOM element during re-renders, failing compositing each time ("Unsupported CSS property: scrollbar-color"). Fix: move scrollbar-width/scrollbar-color from * to html — these properties inherit naturally, so the browser only resolves them on elements that actually have a scrollbar. Also removes the *:hover toggle which triggered style recalc on every hover event.

2. Vuetify slider transitions: The v-slider track and thumb elements had CSS transitions on width/left, causing layout-triggering animations on every progress update. Since the value is already updated frequently via requestAnimationFrame, these transitions are redundant. Fix: disable transitions on slider internals.


apophisnow:
Removing in favor of full shadcn migration

PR #1543, opened 2026-03-06
Improve now playing screen phone

MarvinSchenkel:
Just thought it was fun to play a bit with the now playing screen on the phone to leverage the unused space a bit more, without changing functionality. Here are some before / after screens. Also now hide the album title and player button on really small screens (iphone SE / Galaxy S8+ like).

image
image

image
image

image
image

image
image

Example of album title and player bar hidden on the Galaxy S8+:
image


stvncode (dismissed):
Good improvement, thanks @MarvinSchenkel (needs a rebase tho)

stvncode (approved):
❤️

PR #1544, opened 2026-03-06
Party mode enhancements

apophisnow:
This branch is to collect and fix any bugs or enhancements found during beta testing for Party Mode prior to the GA release.


src/components/party-mode/PartyModeQR.vue
```diff
@@ -4,19 +4,22 @@
...
-      <a
-        :href="qrCodeUrl"
-        target="_blank"
-        rel="noopener noreferrer"
-        class="qr-link"
-      >
+      <div class="qr-link" @click="copyUrlToClipboard">
         <canvas ref="qrCanvas"></canvas>
-      </a>
+        <Transition name="copy-toast">
```
radiohe4d, line 16: Would we not use a component for these type of popups/notifications? Like Snackbar/Alert?

@MarvinSchenkel I know the plan is to move to Shadcn, so not sure whether I should be recommending Vuetify components or Shadcn in this instance?

src/components/party-mode/PartyModeQR.vue
```diff
@@ -175,21 +167,41 @@ onMounted(async () => {
...
-      await generateQRCode();
-    } else {
-      guestAccessEnabled.value = false;
-      qrCodeUrl.value = "";
-    }
-  });
+  unsubscribe = () => {
+    unsubProviders();
+    unsubCoreState();
+  };
```
radiohe4d, line 204: Vue has a built in function to handle unsubscribing before unmount. I believe this should be used in place of this manual unsubscribe assignment.

Then the if statement can be removed from the onUnmounted() event.

Edit:
I've seen Marcel use this method inside the 'onMount' event a few times, but there are more examples of it being used outside on its own. I believe both can be used, so worth giving it a go. 

```suggestion
  onBeforeUnmount(() => {
    unsubProviders();
    unsubCoreState();
  });
```

src/components/party-mode/PartyModeQR.vue
```diff
@@ -234,6 +247,43 @@ onUnmounted(() => {
...
+  opacity: 0;
+  transform: translate(-50%, -50%) scale(0.8);
+}
+
+.copy-toast-leave-to {
+  opacity: 0;
+  transform: translate(-50%, -50%) scale(0.8);
+}
+
 .qr-display canvas {
```
radiohe4d, line 287: I'm not sure if we should be creating whole components from scratch. @MarvinSchenkel thoughts on this?

src/components/party-mode/PartyModeQR.vue
```diff
@@ -249,13 +299,32 @@ onUnmounted(() => {
...
 }
 
-.qr-disabled p {
-  margin-top: 0.5rem;
-  color: rgba(255, 255, 255, 0.5);
+.qr-disabled-title {
+  font-size: 2rem;
+  font-weight: 600;
+  margin-bottom: 1rem;
+  color: rgba(255, 255, 255, 0.9);
```
radiohe4d, line 321: I believe there are variables that can be used here, so when switching dark/light mode they will change with that theme change. Check out /src/styles/styles.css.

Same comment for line 326 below.

src/components/party-mode/PartyModeQueueItem.vue
```diff
@@ -4,42 +4,62 @@
...
+        :loading="boosting"
+        :disabled="boostDisabled"
+        class="boost-btn"
+        :style="{ backgroundColor: boostBadgeColor }"
+        @click.stop="$emit('boost', item)"
+      >
+        <v-icon start size="small">mdi-rocket-launch</v-icon>
+        {{ $t("providers.party_mode.boost") }}
+      </v-btn>
+    </div>
```
radiohe4d, line 62: I think the visual bug i posted where the boost button was floating outside the queue item is partially due to this DIV sitting outside the 'queue-item-row' DIV above. Moving it inside that and setting max-width on the button should resolve this.

src/components/party-mode/PartyModeQR.vue
MarvinSchenkel: Yeah we use toast from vue-sonner:
e.g. `toast.error(t("error_loading_genres"));`. Could you adjust the toasts in the plugin @apophisnow ?

src/views/PartyModeGuestView.vue
MarvinSchenkel: Are these watches still required?

src/views/PartyModeDashboardView.vue
```diff
@@ -76,8 +89,23 @@ const theme = useTheme();
...
+    "party_mode/player",
+  );
+  if (partyPlayerId) {
+    store.activePlayerId = partyPlayerId;
+    if (!api.players[partyPlayerId]) {
+      accessError.value = "no_access";
+    } else {
+      accessError.value = "";
+    }
+  }
```
MarvinSchenkel, line 102: I would add an `else` here to always clear the error when no speaker is configured:
```
else {
    accessError.value = "";
}
```

src/views/PartyModeDashboardView.vue
```diff
@@ -398,7 +419,13 @@ onMounted(async () => {
...
   });
 
-  unsubscribeFunctions.value = [unsub1, unsub2];
+  // Subscribe to provider updates to detect party mode player config changes
+  const unsub3 = api.subscribe(EventType.PROVIDERS_UPDATED, async () => {
+    await refreshPartyPlayer();
+    fetchQueueItems(true);
+  });
+
+  unsubscribeFunctions.value = [unsub1, unsub2, unsub3];
```
radiohe4d, line 427: See earlier comment about 'onBeforeUnmount'.

src/views/PartyModeDashboardView.vue
```diff
@@ -557,6 +584,32 @@ watch(
...
+  align-items: center;
+  text-align: center;
+  gap: 1rem;
+  padding: 2rem;
+}
+
+.access-error-title {
+  font-size: 1.5rem;
+  font-weight: 600;
+  color: rgba(255, 255, 255, 0.9);
```
radiohe4d, line 603: See earlier comment about using variables.

src/views/PartyModeGuestView.vue
```diff
@@ -365,6 +388,44 @@ const addToQueue = async (item: Track | Artist, position: "next" | "end") => {
   }
 };
 
+const boostQueueItem = async (item: QueueItem) => {
+  if (!boostEnabled.value) {
+    toast.warning($t("providers.party_mode.boost_disabled"));
```
radiohe4d, line 376:
```suggestion
    toast.warning($t("providers.party_mode.guest_page.boost_disabled"));
```

src/views/PartyModeGuestView.vue
```diff
@@ -365,6 +388,44 @@ const addToQueue = async (item: Track | Artist, position: "next" | "end") => {
...
+    return;
+  }
+
+  if (rateLimitingEnabled.value && boostTokens.value <= 0) {
+    const minutesUntilNext = getTimeUntilNextToken();
+    toast.warning(
+      $t("providers.party_mode.boost_limit_reached", [minutesUntilNext]),
```
radiohe4d, line 383:
```suggestion
      $t("providers.party_mode.guest_page.boost_limit_reached", [minutesUntilNext]),
```

src/views/PartyModeGuestView.vue
```diff
@@ -365,6 +388,44 @@ const addToQueue = async (item: Track | Artist, position: "next" | "end") => {
...
+
+    if (rateLimitingEnabled.value) {
+      consumeBoostToken();
+    }
+
+    const name = item.media_item?.name || item.name;
+    toast.success($t("providers.party_mode.guest_page.item_boosted", [name]));
+  } catch (error) {
+    console.error("Failed to boost queue item:", error);
+    toast.error($t("providers.party_mode.add_to_queue_failed"));
```
radiohe4d, line 406:
```suggestion
    toast.error($t("providers.party_mode.guest_page.add_to_queue_failed"));
```

src/views/PartyModeGuestView.vue
```diff
@@ -429,31 +501,30 @@ onMounted(async () => {
...
-    console.debug("[GuestView] partyPlayerId:", partyPlayerId);
-    console.debug("[GuestView] store.activePlayerId:", store.activePlayerId);
-    console.debug("[GuestView] store.activePlayer:", store.activePlayer);
-    console.debug("[GuestView] api.players keys:", Object.keys(api.players));
-    console.debug("[GuestView] api.queues keys:", Object.keys(api.queues));
-    console.debug("[GuestView] queue state:", queue.currentQueue.value?.state);
+    await refreshPartyPlayer();
   } catch (error) {
     console.error("Failed to fetch party mode player:", error);
   }
```
radiohe4d, line 492: I would move the try catch into the 'refreshPartyPlayer' method, that way its all together as one and always handled when called.

src/components/party-mode/PartyModeQR.vue
apophisnow: I want the popup to happen where the user clicked so they don't miss the toast and wonder why clicking didn't do anything. If we have another component that can match the current UI for this I'm all for it, but I don't think toast is the best UX here.

apophisnow: This is currently implemented similar to what discord does when you click something that copies a link.

src/views/PartyModeDashboardView.vue
apophisnow: Will fix.

MarvinSchenkel (approved):
LGTM, thanks @apophisnow

PR #1545, opened 2026-03-07
Stop sidebar showing when mobile layout forced

OzGav:
Fixes https://github.com/orgs/music-assistant/discussions/5055#discussioncomment-16029779


stvncode (approved):
Thanks @OzGav

PR #1547, opened 2026-03-07
Add genre exclusion feature to UI

jozefKruszynski:
This adds ui elements for excluding discovered genres from media items.


MarvinSchenkel (approved):
Great work, thanks @jozefKruszynski

PR #1548, opened 2026-03-08
Improve 'Now playing' screen for other resolutions as well

MarvinSchenkel:
Follow up to [this PR](https://github.com/music-assistant/frontend/pull/1543). Now also apply the same design changes to other resolutions as well.

Some old vs new for ipad air / desktop with queue opened:

image
image
image
image

image
image
image
image

image
image
image
image


stvncode (approved):
Nice !

PR #1560, opened 2026-03-09
Fix for disabling the web player on party mode routes

maximmaxim345:
Refactors web player mode selection into `web_player.ts` so `resolvePreferredMode()` decides from route policy, party guest state, companion mode, and user preferences in one place.

Adds a router level property `meta.disableWebPlayer` so future pages can also easily opt out of the web player.

Also fixes a bug where the internally used broadcast channel (for selecting the tab used for the web player) was active even when the web player was disabled.


MarvinSchenkel (approved):
Thanks @maximmaxim345

PR #1561, opened 2026-03-09
Improve Sendspin Web Player syncing

maximmaxim345:
## Summary

Bumps `sendspin-js` to 2.0.3 with a lot faster initial syncing and considerably increased syncing precision.
Also switches from playing audio from a `<audio>` element to direct playback through `sendspin-js` (for desktop browsers).

## Switch from `<audio>` to direct playback on desktop

For some reason, using `<audio>` for playback on Chromium browsers on linux caused a pitched up distortion for exactly the first 10s of playback.
This sounded like a pitched up distortion with gaps in between each chunk. I don't know why this happens, but directly outputting without a `<audio>` seams to work without issues.

Direct playback is only done on desktop browsers for now

## Syncing improvements of `sendspin-js` 2.0.3
With the switch to a burst strategy, playback can now almost immediately start in sync, with less corrections and distortions on new browsers sessions.

Sync accuracy is also considerably improved now, measuring to under a millisecond in a real world scenario.
Random out of sync jumps are also a lot more rare now, only happening once in an hour from my tests.

test-final

# Testing
Tested on all devices I have, so basically everything except Safari on Mac.


MarvinSchenkel (approved):
Thanks @maximmaxim345

PR #1564, opened 2026-03-10
Fix/nowplaying badge overflow

apophisnow:
After optimizing the visualizer animation so that it can now take advantage of GPU, it gets cut off in the footer player controls. This PR fixes that issue and migrates to shadcn. I think moving the NowPlaying visualizer to be overlayed in the bottom right corner of the album art image gives the user a consistant place to expect to see it, rather than it being in a different location depending on what page you are looking at.

Broken animation:
Screenshot 2026-03-10 at 12 27 57 PM

After fix:
Screenshot 2026-03-10 at 1 01 40 PM

Fullscreen view after fix:
Screenshot 2026-03-10 at 12 52 55 PM

### DIFF
--- a/src/components/SendspinPlayer.vue
+++ b/src/components/SendspinPlayer.vue
@@ -347,7 +347,10 @@ onMounted(() => {
       if (!shouldBePlaying || !audioRef.value) return;
 
       audioRef.value.play().catch((error) => {
-        console.warn("Sendspin: Failed to recover audio element playback:", error);
+        console.warn(
+          "Sendspin: Failed to recover audio element playback:",
+          error,
+        );
       });
     });
   }
--- a/src/layouts/default/PlayerOSD/Player.vue
+++ b/src/layouts/default/PlayerOSD/Player.vue
@@ -248,7 +248,7 @@ watch(
 
 .mediacontrols-left {
   margin-inline-end: auto;
-  width: 20%;
+  min-width: 20%;
   > div {
     padding: 0px !important;
   }
--- a/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
+++ b/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
@@ -244,8 +244,22 @@
                     @mouseleave="hoveredQueueIndex = -1"
                   >
                     <template #prepend>
-                      <div class="media-thumb listitem-media-thumb">
+                      <div class="media-thumb listitem-media-thumb relative">
                         <MediaItemThumb size="50" :item="item" />
+                        <div
+                          v-if="
+                            item.queue_item_id ===
+                              store.curQueueItem?.queue_item_id &&
+                            store.activePlayer?.playback_state !=
+                              PlaybackState.IDLE
+                          "
+                          class="absolute bottom-0 right-0"
+                        >
+                          <NowPlayingBadge
+                            :show-badge="false"
+                            :show-icon="true"
+                          />
+                        </div>
                       </div>
                     </template>
                     <template #title>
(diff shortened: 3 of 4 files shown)


MarvinSchenkel:
Let's see what @Erioldoesdesign thinks about this one. I personally prefer it next to the speaker name because it draws attention to **where** the music is currently playing:
image

apophisnow:
> Let's see what @Erioldoesdesign thinks about this one. I personally prefer it next to the speaker name because it draws attention to **where** the music is currently playing: image

My counterpoint would be that on the fullscreen view we also have the animation, and it is nowhere near any indicator for the speaker it is playing on. So if the goal is to show where the music is playing, we are not being consistant about it.

I personally don't see the equalizer bar and think "that's the speaker my music is playing on", but rather "ok that track is currently playing." As far as where is it playing, as a new user this took me a bit to figure out, but now understand the mdi-speaker icon to indicate that. So I think if anywhere we were wanting to indicate something about a speaker, that iconagraphy was used.

In your screenshot, at a glance and as a new user, I might see that speaker name and think it is either an Artist or Album name.

MarvinSchenkel:
Definitely agree with you here that we should be consistent with our animations. But before we remove the EQ animation from the speaker (and promote it to be exclusively used next to song title text only), we should come up with a solution for drawing attention to the speaker that is currently playing music, as it can be quite disruptive if people play music to the wrong speaker..

I just had an idea that we could make a speaker icon animation that animates MA blue music notes when the speaker is playing music and use that instead of the EQ animation. Kinda something like https://media1.giphy.com/media/v1.Y2lkPTc5MGI3NjExMDRyNXB2eTdueTU2NmRhZjM1b2g0Y3QwYjd0NnliM3UxNGM4MmFqeiZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/qZelsWAXfWpAA/giphy.gif

@Erioldoesdesign ?

Erioldoesdesign:
Yeah i prefer this next to the speaker name too! the job of the animation is to draw attention to what speaker is playing music and not what album is playing right now since the more painful experience is accidentally playing LOUD music on a speaker you do not want to play LOUD music on (just one example!) 

So yeah the animation should stay contextually next to the speaker name and we need the UI to support helping the user know and recognise as much as possible what speaker is currently playing music - since it could be a speaker, somewhere that you cannot hear right now aka a speaker several rooms/floors away from where you are operating Music Assit.

Erioldoesdesign:
Something to add here is that the now playing bar and the now playing page are very different mental modes/contexts.

Now playing minimised bar: Needs to give contextually correct info at a glance, user can be in an active 'adding tracks to my queue/playlist mode' or replying a list that was paused. High likelyhood of accidental 'oops not that player' happening here given the way to change players is via the expanded play button on a track/album/artist or via the pop-out play drawer. 

Now playing maximised page: Is about sitting back and enjoying your chosen playlist/queue, you are less likely to be actively adding new music to a play queue (given the view) and more likely passively observing the list of tracks or enjoying album art/other trac details. The speaker UI here is also much more visibily clear and different to the now playing bar where the stack of texts are similar styles/sizes and hard to immediately differentiate. I imagine users have less problems noticing the speaker that is being used on the now playing full screen page than the now playing bar.

Given these different brain-spaces i'm not mad from a UX perspective that on the now playing maximised page, the animation is on the track listing not next to the player, like in the now playing minimised bar.

apophisnow:
It looks like I forgot to put my idea for visual separation of the speaker name (it was in the UX channel in discord), but I think anywhere we display the speaker name, it should perhaps be accompanied by the speaker icon. If we also make it a badge, then I think it gets less visually lost among the album, artist name etc...
image

apophisnow:
In my example above it is getting mixed with the gradient background, but ignore that.

Erioldoesdesign:
> anywhere we display the speaker name, it should perhaps be accompanied by the speaker icon.

Yes! totally agree with this! i can't think of any places where we say the speaker name and it would be 'bad' to put the speaker. There might be some exceptions but unsure of where they would be for now.

A badge is an interesting idea...i like it in terms of us making a feature of the speaker in terms of hierarchy e.g. if we give it a button style or even an border outline stroke (like in the full screen of the now playing page) then we indicate the speaker choice/current speaker as a prominent element of UI. I like the concept, but i'm not sure if a solid background, or gradient background to a button is the visual i would decide on. But agreed in principle re. some importants being given to the speaker

PR #1565, opened 2026-03-11
Party mode lyrics and karaoke

apophisnow:
Fixes multiple things with the Party Mode dashboard, mainly focusing on mobile view.

Adds lyrics/karaoke feature to dashboard.


src/components/LyricsViewer.vue
```diff
@@ -584,16 +450,16 @@ onMounted(() => {
   font-style: italic;
 }
 
-.lyrics-scroll-container {
-  height: 100%;
```
stvncode, line 588: You remove this but there is other place when this css is used (i see 3 places where you used :deep(.lyrics-scroll-container))

src/components/party/PartyModeQueueItem.vue
```diff
@@ -40,26 +37,28 @@
...
+      <div class="queue-item-spacer"></div>
+      <div class="queue-item-actions">
+        <Button
+          size="sm"
+          :disabled="boostDisabled || boosting"
+          class="boost-btn"
+          :style="{ '--btn-bg': boostBadgeColor }"
+          @click.stop="$emit('boost', item)"
+        >
+          <Spinner v-if="boosting" class="size-4" />
```
stvncode: Why different way for sizing, doesn't matter for this one but here's class and size just below. Can be complicated for new contributors and it's better for consistency as well

src/components/party/PartyModeQueueItem.vue
```diff
@@ -155,38 +158,45 @@ const badgeLabel = computed(() =>
...
 }
 
 .queue-item-actions .boost-btn {
   font-weight: 600;
   text-transform: none;
   letter-spacing: 0.5px;
-  color: white;
-  flex: 1;
+  background-color: var(--btn-bg) !important;
+  color: rgb(var(--v-theme-on-primary)) !important;
```
stvncode, line 197: I saw in a lot of place that you are using rgb before var etc, you don't need this, var is good enough.

src/components/party/PartyModeSearchBar.vue
```diff
@@ -1,33 +1,38 @@
...
-      enterkeyhint="search"
-      class="search-input"
-      @update:model-value="$emit('update:searchQuery', $event)"
-      @click:clear="$emit('clear')"
-      @keydown.enter="$emit('submit')"
-    />
+      <ArrowLeft :size="20" />
+    </Button>
+
+    <div class="search-input-wrapper">
```
stvncode: For Search Input, use InputGroup. You have other example in PlayerFilters for example: `
<InputGroup class="search-field">
      <InputGroupInput v-model="searchQuery" :placeholder="$t('search')" />
      <InputGroupAddon>
        <Search />
      </InputGroupAddon>
    </InputGroup>
`

src/components/party/PartyModeSearchBar.vue
```diff
@@ -73,12 +72,22 @@ defineProps<{
 }>();
 
 defineEmits<{
-  "update:searchQuery": [value: string];
+  "update:searchQuery": [value: string | number];
```
stvncode: update:searchQuery emit type widened to string | number while prop is string. Can pass numbers to string consumers

src/components/party/PartyModeTokensBadge.vue
```diff
@@ -25,6 +26,17 @@ const props = defineProps<{
   icon: string;
 }>();
 
+const iconComponent = computed(() => {
```
stvncode: Can you use match from ts-pattern here please for better types 🙏

stvncode: This will look like:
`match(props.icon)
.with("mdi-rocket-launch", () => Rocket)
.with("mdi-playlist-plus", () => ListPlus)
.otherwise(() => Rocket)
`

stvncode (commented):
Thanks a lot for all the change regarding the new ui-lib 🙏 
Some small comment on my own

src/components/party/PartyModeQueueItem.vue
apophisnow: AI slop in this case. What is the preferred pattern in MA?

src/components/party/PartyModeQueueItem.vue
apophisnow: I think that was a remanent from hardcoded values. will fix

src/components/party/PartyModeQueueItem.vue
stvncode: use the props before class, so here :size

src/components/LyricsViewer.vue
```diff
@@ -153,61 +237,80 @@ const hasLyrics = computed(() => {
...
-    // Handle milliseconds/hundredths part with more precision
-    if (match[3]) {
-      const decimalPart = match[3].replace(/[.:]/, ".");
-      milliseconds = Math.round(parseFloat(decimalPart) * 1000);
-    }
-
-    // Calculate total time in milliseconds for precision
-    const timeMs = minutes * 60 * 1000 + seconds * 1000 + milliseconds;
+// Parse a timestamp string "mm:ss.xx" into seconds
+const parseTimestamp = (
```
stvncode: Forgot that last time, can you put that in another file pls ? This does not need to be in a component, you can put that in a file in helpers and add some test in tests/helpers/xxx.test.ts

Same for: parseWordTimestamps and parseLrcLine

src/plugins/router.ts
```diff
@@ -13,32 +13,30 @@ const routes = [
...
     children: [
       {
         path: "",
         name: "party",
         component: () =>
           import(
-            /* webpackChunkName: "party" */ "@/views/PartyModeDashboardView.vue"
+            /* webpackChunkName: "party" */ "@/views/PartyDashboardView.vue"
           ),
         props: (route: { query: Record<string, any> }) => ({ ...route.query }),
```
stvncode: Why do we have an any here ?

src/plugins/router.ts
```diff
@@ -13,32 +13,30 @@ const routes = [
...
         beforeEnter: async (_to: any, _from: any, next: any) => {
```
stvncode: Same here

stvncode (commented):
Almost good. Sorry, i missed a few things last time

stvncode (approved):
Awesome 🙏

PR #1568, opened 2026-03-11
Add new custom icon and modify others

jozefKruszynski:
Adds agreed upon icon for Genres in different places, and changes the icon for show hide empty genres in the genre library toolbar


stvncode (commented):
I don't see where this is used ?

MarvinSchenkel:
Looks good to me, but can we move the svg into the `assets` folder?

jozefKruszynski:
> Looks good to me, but can we move the svg into the `assets` folder?

Removed it entirely as we don't actually need it. It's encoded in the Vue component

MarvinSchenkel (approved):
Looks good, thanks @jozefKruszynski

PR #1569, opened 2026-03-12
Fix toast in player settings

OzGav:
Requires: https://github.com/music-assistant/server/pull/3360
Satisifies: https://github.com/orgs/music-assistant/discussions/3288

Frontend support for per-player volume limits (min_volume / max_volume) set via the backend.

**Handles save errors gracefully** — The player settings dialog awaits the save call and stays open on failure, so users can correct invalid values (i.e. setting min > max) instead of being navigated away with no feedback.


stvncode (dismissed):
LGTM 😃

marcelveldt:
Why do we need frontend support for this ?
Can't we do this completely in the backend so it works with ALL api clients ?

OzGav:
The backend PR handles the actual enforcement so all API clients are covered but I found the following UX problems that I felt I needed to resolve in the front end 
	∙	Users could drag the slider to volumes the backend would reject, causing desync between the UI and actual playback volume
	∙	The slider thumb would snap back unexpectedly after the backend corrected the value
	∙	Saving invalid player config (e.g. min > max) would silently navigate away, leaving the user unaware the save failed

marcelveldt:
If we properly implement this in the backend, we will remap the volume completely within the limits.
This ensures its also compatible with all API clients such as HA, mobile apps etc.

In other words: We should not accept this to be implemented in the frontend at all

OzGav:
Fair enough. At some point you can have a look and see how far off the backend PR is…?

stvncode:
Can we close this then @OzGav ?

oleost:
> If we properly implement this in the backend, we will remap the volume completely within the limits. This ensures its also compatible with all API clients such as HA, mobile apps etc.
> 
> In other words: We should not accept this to be implemented in the frontend at all

Seems like something in the line I been trying to implement? 
Scale volume on output so everything else still manage volume like usual.

edit removed link

stvncode:
Sorry to ping you again, can we close this then @OzGav ?

OzGav:
@stvncode can we hold for a bit I do need to rework the backend PR but there may be one part of this, the config validation, that might still be useful?

stvncode:
@OzGav oh sure sorry i thought it was only be, no worries then 😃

src/views/settings/EditPlayer.vue
```diff
@@ -398,8 +398,16 @@ const enablePlayer = function () {
...
+  loading.value = true;
+  try {
+    await api.savePlayerConfig(props.playerId!, values);
+    router.back();
+  } catch {
+    // Error toast is already shown by the API layer (handleResultMessage).
+    // We just prevent navigation so the user can correct values.
+  } finally {
+    loading.value = false;
+  }
```
marcelveldt, line 410: the title of the PR should be updated as this change is no longer about the volume clamping

OzGav: Done

marcelveldt:
edited the PR title because this Pr has nothing to do anymore with strictly the min/max volume feature

marcelveldt (approved):

PR #1570, opened 2026-03-12
Party mode lyrics and karoke

apophisnow:
Builds on top of the UI fixes and adds lyric display and Karaoke mode!

Also a server PR required for this.


apophisnow:
Merged with party mode ui PR

PR #1581, opened 2026-03-14
LyricsViewer fixes/refactor

apophisnow:
Fixes a few things and simplifies the LyricsViewer a bit.
- Musical breaks use LRC instead of estimating based on large gaps.
- Add a musical break that starts 5 seconds before first line on songs with long intros.
- Only active musical breaks animates highlight (in case of multiple shown on screen)
- Removed active line scaling (which made reading distracting IMO even though it looked cool)


src/components/LyricsViewer.vue
```diff
@@ -336,84 +391,57 @@ watch(
   [() => props.mediaItem?.item_id, () => props.lyrics, () => props.lrcLyrics],
   () => {
     activeLyricIndex.value = -1;
-    musicalBreakLineIndex.value = -1;
-    isInMusicalBreak.value = false;
-    contentTranslateY.value = 0;
+    contentTranslateY.value = 99999;
```
stvncode, line 395: Where does this value comes from ?

src/components/LyricsViewer.vue
```diff
@@ -98,44 +105,48 @@ import { ScrollArea } from "@/components/ui/scroll-area";
...
+  highlightAhead: true,
 });
 
-// Core lyrics state
-const parsedLyrics = ref<Array<{ time: number; text: string }>>([]);
+// Core state
+const displayLines = ref<DisplayLine[]>([]);
 const loading = ref(false);
 const activeLyricIndex = ref(-1);
 const scrollAreaRef = ref<InstanceType<typeof ScrollArea> | null>(null);
```
stvncode: This is unused

src/components/LyricsViewer.vue
```diff
@@ -316,15 +373,13 @@ const findActiveLyricIndex = (positionMs: number): number => {
...
-// Calculate the translateY needed to place a line at the anchor point
-// (40% from top of the container).
 const computeTranslateY = (index: number) => {
   const el = lineRefs.get(index);
   const container = syncedContainerRef.value;
   if (!el || !container) return;
 
   const containerHeight = container.clientHeight;
-  const anchorY = containerHeight * 0.35;
+  const anchorY = containerHeight * (props.anticipation ? 0.5 : 0.35);
```
stvncode: Since anticipation can be undefined, can we have this `props.anticipation > 0` instead ?

stvncode (dismissed):
Really small comments, thanks for the pr 🙏

stvncode (approved):
Nice 🙏

PR #1504, opened 2026-02-28
Add support for multiple media types in playlist

fmunkes:
Backend PR: https://github.com/music-assistant/server/pull/3216

Adds support for different media types in playlist.
Added provider filtering at the necessary places. When a playlist is a created via add to playlist, then the playlist type is deduced from the media_type of the item. If a provider supports mixed types, we always create a playlist which supports all types.

The playlist create dialog within the PlaylistView depends on the provider Feature.
When a provider supports only a single media type or mixed, we only state that:
single_type
mixed_types
Otherwise, the user can choose:
multiple_non_mixed

PR #1582, opened 2026-03-15
Skip provider lookup when saving queue as playlist

chrisuthe:
## Summary
- The media types refactor (#1504) added provider feature detection to `CreatePlaylistDialog`, but the queue-to-playlist flow only passes a `queueId` (no `providerId`)
- `getProvider("")` returned `undefined`, triggering an error toast before the dialog could function
- Wrapped provider detection and media type UI in a `!queueId` guard since the backend `save_as_playlist` endpoint handles provider selection internally


chrisuthe:
@fmunkes  can you peak here since I'm changing your work

fmunkes:
I checked that PR out, and the playlist create/ add is still functional. But when I click "save queue as playlist" there is nothing happening. I pointed it at the latest beta. The web logs just give a 
``` 
Warning: Missing `Description` or `aria-describedby="undefined"` for DialogContent.
```

I'm not that good with the frontend yet, but if you want me to dig deeper, let me know. After all, I guess I broke your functionality here...😬 Very sorry!

stvncode:
When queueId is set, no <DialogDescription> is rendered, causing the aria-describedby warning i think.
Since there is a `<template v-if="!queueId">`, we need to handle the case when there is no queueId. Might want to add after the template close (line 49) this: `<template v-else>
          <DialogDescription class="sr-only">
            {{ $t("save_queue_as_playlist") }}
          </DialogDescription>
        </template>`

chrisuthe:
Good gravy, sorry guys, let me go back and fix this properly. I messed up my fix and tested a different version. Hold Please.

chrisuthe:
Hey @fmunkes fixed up my part here, can you make sure I didn't impact anything you were doing? Thank you!

src/layouts/default/CreatePlaylistDialog.vue
```diff
@@ -91,6 +96,7 @@ import router from "@/plugins/router";
 import { store } from "@/plugins/store";
 
 const showDialog = ref(false);
+const dialogKey = ref(0);
```
fmunkes, line 101: Found that in your commits
"Force Dialog remount via dynamic key to prevent the enter animation from stalling at opacity:0 when opened from a context menu"
At least for me that is not obvious. So I'd add a comment :)

fmunkes:
Other than that PR seems fine to me, gave it a test run too, all good!

src/layouts/default/CreatePlaylistDialog.vue
chrisuthe: Yea that is a super weird deal where it just never makes it any further than 0 when you use that from a context menu. Took a while to find that situation (You probably saw the 300 debug builds I did lol)

chrisuthe:
> Other than that PR seems fine to me, gave it a test run too, all good!

Thank you for looking it over!

marcelveldt:
@fmunkes @stvncode @chrisuthe 
I think this needs to be merged for 2.8 release ?

It has a few few merge conflicts. Can these be solved.

src/layouts/default/CreatePlaylistDialog.vue
```diff
@@ -91,6 +96,7 @@ import router from "@/plugins/router";
 import { store } from "@/plugins/store";
 
 const showDialog = ref(false);
```
fmunkes, line 98:
```suggestion
const showDialog = ref(false);
// force Dialog remount via dynamic key to prevent the enter animation from
// stalling at opacity:0 when opened from a context menu
```

fmunkes (dismissed):

fmunkes (approved):

chrisuthe:
Thanks @fmunkes I was on vacation, much appreciated getting those fixes in!

PR #1593, opened 2026-03-16
Hide sidebar navigation in frameless mode

OzGav:
The frameless URL parameter (?frameless=true) hides the footer/player bar but the sidebar menu was still visible after the navigation was reworked. This conditionally hides the AppSidebar and removes unnecessary bottom padding when in frameless mode.


stvncode (approved):
Thanks 🙏

PR #1595, opened 2026-03-16
Fix missing genre filter options with pagination addition

jozefKruszynski:


src/components/ItemsListing.vue
```diff
@@ -1316,18 +1316,29 @@ const loadGenreOptions = async () => {
...
-      undefined,
-      undefined,
-      100,
-      0,
-      "name",
-    );
+    const pageSize = 100;
+    const all: { label: string; value: number }[] = [];
+    let offset = 0;
+    let page: Awaited<ReturnType<typeof api.getLibraryGenres>>;
```
stvncode: Isn't the type just Genre[] ?

jozefKruszynski: Yes it is.

MarvinSchenkel (approved):

PR #1596, opened 2026-03-16
Add AI Radio Plugin to Frontend

swiftbird07:
As discussed in Discord here is the pull request for my AI Radio plugin.

> This is the PR for the frontend changes of AI Radio. For the backend / server PR see here: https://github.com/music-assistant/server/pull/3407

The plugin allows a user to define a 'station' and specify a MA playlist and the tool will generate the text and respective TTS mp3 files and insert them into a new playlist (or to a current queue) making it more like a radio like experience with weather reports, song intro and outros, news etc.

## Features 
- server adds new API commands for ai_radio if plugin is enabled
- AI-Radio frontend only shows "AI Radio" menu item if plugin is enabled
- There are two modes of operation: Playlist mode and Dynamic mode. In playlist mode the station is "pre-built" ahead of  time, while in Dynamic mode each section is generated "on the fly" for the current queue.
- Via the menu you can configure stations and sections and then start the radio
- Sections are used in station configuration and are basically re-usable prompts with placeholders and a few added config options like if you want to enable (or force) web search for a prompt
- Stations define how each selected section is used and with what guards; you can specify what happens at the start of the radio, between songs and at the end. For each of these you can configure specific flow rules that define what must happen or what can happen with a chance. E.g. you can define that between songs a song transition will be moderated and then with a 20% chance a weather broadcast is inserted as well
- If more than one section was selected it will be automatically merged together so it sound more like a continuous moderation
- Right-click menu "Run with AI Radio" for playlist to start AI Radio with the playlist pre-selected

## Notes 
- I tried to have the frontend plugin change as few lines of core code as possible (like the new party mode plugin) but if you think it could be even less then feel free to give feedback 

## Screenshots

Screenshot 2026-03-14 at 16 45 04
Screenshot 2026-03-14 at 16 46 57
Screenshot 2026-03-14 at 16 46 48


src/composables/ai-radio/useAiRadio.ts
```diff
@@ -0,0 +1,183 @@
...
+    );
+    toast.success(
+      mode === "playlist"
+        ? $t("providers.ai_radio.toast.playlist_starting")
+        : $t("providers.ai_radio.toast.live_starting"),
+    );
+    await loadStatus(true);
+    return result;
+  } catch (error) {
+    toast.error(
```
stvncode: Almost every action currently produces **two stacked error toasts**: the composable toasts and rethrows here, then the view catches the same error and toasts again (`runStart` → `run_start_failed`). Same pattern for `stopRun` + `stopSession`, and for save/delete of stations and sections in the editor composable. I'd pick one layer — e.g. toast in the view and let the composables throw silently.

src/composables/ai-radio/useAiRadio.ts
```diff
@@ -0,0 +1,183 @@
...
+    try {
+      return JSON.stringify(error);
+    } catch {
+      return String(error);
+    }
+  }
+  return String(error);
+};
+
+async function loadStations(silent = false): Promise<AIRadioStation[]> {
```
stvncode: The `silent` parameter is effectively dead code — nearly every call site passes `silent=true`, and success toasts for plain data loads ("Stations loaded") are noise UX anyway. Dropping the parameter and the load-success toasts would simplify both composables noticeably.

src/composables/ai-radio/useAiRadioEditor.ts
```diff
@@ -0,0 +1,335 @@
...
+
+const deepClone = <T>(value: T): T => {
+  return JSON.parse(JSON.stringify(value)) as T;
+};
+
+const sortByName = <T extends { name: string }>(items: T[]): T[] => {
+  return [...items].sort((a, b) => a.name.localeCompare(b.name));
+};
+
+async function loadStations(silent = false): Promise<AIRadioStation[]> {
```
stvncode: This duplicates `useAiRadio`: both composables keep their **own module-level `stations` ref** loaded from the same `ai_radio/stations/list` command, and `loadStations`, `sortByName` and the toast plumbing are copy-pasted between the two files (`errorMessage()` exists three times across the PR). Consider merging them, or having the run-side composable consume the editor's station list.

Also worth a comment either way: module-level refs are app-lifetime singletons, so drafts and selections survive leaving and re-entering the page.

src/composables/ai-radio/useAiRadioEditor.ts
```diff
@@ -0,0 +1,335 @@
...
+
+async function loadPlaylists(silent = false): Promise<Playlist[]> {
+  loadingPlaylists.value = true;
+  try {
+    // Load in pages to avoid truncating larger libraries.
+    const limit = 200;
+    let offset = 0;
+    let hasMore = true;
+    const allItems: Playlist[] = [];
+    while (hasMore && offset < 5000) {
```
stvncode: This silently caps at 5,000 playlists — a user with more will just not see some entries in the picker with no indication why. If the cap is needed, surface it (or log it) rather than truncating silently.

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
+<template>
```
stvncode, line 1: At 3,571 lines this is 1.6× the next-largest view in the repo (`Login.vue`) and is hard to review or maintain as a single file. It contains at least six natural components: Run tab, Sessions card, Station editor, Flow-rule editor, Section editor, Guided wizard, Tutorial dialog. The template also repeats near-identical blocks (player select ×3, playlist select ×2, placement select ×3) that would collapse into small shared components.

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
...
+      }
+    }
+
+    if (sections.value.length > 0 && !selectedEditorSectionId.value) {
+      await selectSectionForEdit(sections.value[0].id);
+    }
+
+    applyRouteOverrides();
+
+    refreshTimer = setInterval(() => {
```
stvncode: This polls `ai_radio/status` every 5s unconditionally — even with zero sessions, on the Stations/Sections tabs, and when the tab is hidden. The app already has a WS event bus (`api.subscribe`); if the backend can't emit session events yet, consider at least polling only while a session is `running`.

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
...
+                    id="station-target-provider"
+                    v-model="stationDraft.target_playlist_provider"
+                  />
+                </div>
+
+                <div class="space-y-2 md:col-span-2">
+                  <Label for="station-sections">{{
+                    $t("providers.ai_radio.fields.selected_sections")
+                  }}</Label>
+                  <select
```
stvncode: A `<select multiple>` with a ctrl-click hint is effectively unusable on mobile, which this app supports. The wizard's step 2 already uses the right pattern (a checkbox list) — reusing that here would fix it and remove the need for the hint text.

stvncode (commented):
Thanks for this — it's a substantial, well-typed feature and the plugin gating (menu/router/`enabledPlugins` refactor in `App.vue`) mirrors the existing `party` pattern nicely. I left inline comments on the code; a few cross-cutting points that don't anchor to a single line:

**First-run experience:** a new user gets an auto-opened tutorial modal, then lands on a Run tab with an empty station dropdown and two disabled buttons. Consider detecting the no-stations state and showing an inline empty state with a single "Create your first station" CTA into the wizard, with the tutorial content available inline rather than as a blocking modal.

**Mixed component systems:** the Run tab uses the shadcn `Select`, while the editors and wizard use raw native `<select>` elements, so the same screen shows two visual styles. Standardizing on the UI-kit components would tighten the look.

**Tests:** the PR adds ~4,700 lines with no tests, but `normalizeStationDraft`, `validateStationDraftLocal`, `buildGuidedSectionOrder` and `defaultGuidedPlacement` are pure functions that would be cheap to cover — and the placement heuristic already produced one bug (fixed in eb29f31), which is a good sign it deserves a spec.

**Branch is behind main:** the branch still has `yarn.lock` and predates the pnpm migration and other recent main changes — please rebase onto latest main. That should also shrink the translation-file diff considerably (see inline comments on `en.json`/`de.json`).

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
...
+                    min="1"
+                    step="1"
+                    :placeholder="
+                      $t('providers.ai_radio.placeholders.use_station_default')
+                    "
+                  />
+                </div>
+              </div>
+
+              <div class="grid gap-2 sm:grid-cols-2">
```
stvncode: UX: "Create playlist" and "Start live radio" have identical styling and both carry the Sparkles icon, and the difference between the two modes is only explained in the tutorial. A short caption under each button — or a mode selector plus a single Start button — would read better. Also, the override fields above (playtime cap, batch size) mostly apply to dynamic mode but sit in front of both buttons; grouping them under an "Advanced" disclosure would reduce first-run complexity.

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
...
+                    >
+                      <div class="flex flex-wrap items-center gap-2">
+                        <select
+                          :value="getFlowType(flowItem)"
+                          class="h-9 rounded-md border bg-background px-3 text-sm"
+                          @change="
+                            onFlowTypeChange(ruleIndex, flowIndex, $event)
+                          "
+                        >
+                          <option value="MUST">MUST</option>
```
stvncode: Raw untranslated backend jargon (`MUST` / `ALTERNATIVE` / `OPTIONAL`) in a user-facing dropdown. Humanized, translatable labels ("Always play", "Pick one of…", "Sometimes play") would make the flow editor — the most intimidating part of the feature — much more approachable. Same for the five text buttons per rule row (Up/Down/Add item/…): icon buttons or a drag handle would lighten it a lot.

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
...
+                  {{ $t("providers.ai_radio.station_editor.advanced") }}
+                </summary>
+                <div
+                  class="mt-5 grid gap-x-6 gap-y-5 md:grid-cols-2 md:gap-x-8 [&>div]:min-w-0"
+                >
+                  <div class="space-y-2">
+                    <Label for="station-id">{{
+                      $t("providers.ai_radio.fields.station_id")
+                    }}</Label>
+                    <Input id="station-id" v-model="stationDraft.id" />
```
stvncode: The station `id` is editable here even for an existing station — changing it and saving will create a duplicate rather than renaming (the backend keys on id). Suggest making the id read-only once the station exists, or explicitly handling rename.

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
...
+      ]),
+    );
+  }
+};
+
+const removeSelectedStation = async () => {
+  if (!selectedEditorStationId.value) {
+    return;
+  }
+  if (!window.confirm($t("providers.ai_radio.confirm.delete_station"))) {
```
stvncode: `window.confirm` — this is the only usage in the repo. Using the existing `Dialog` component would keep the delete confirmation consistent with the app's look and give you translated buttons.

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
...
+      ]),
+    );
+  }
+};
+
+const triggerStationImport = () => {
+  stationImportInput.value?.click();
+};
+
+const onStationImport = async (event: Event) => {
```
stvncode: Import silently takes only `stations[0]` from a multi-station file and ignores the rest — worth telling the user, or supporting multi-import. It also blindly casts the parsed JSON with no shape validation; a malformed file will surface as confusing downstream errors rather than a clear "invalid file" message (the `ai_radio/stations/validate` command could be reused here).

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
...
+  }
+  return String(error);
+};
+
+const closeTutorial = () => {
+  tutorialOpen.value = false;
+  localStorage.setItem(TUTORIAL_SEEN_STORAGE_KEY, "1");
+};
+
+const defaultGuidedPlacement = (section: AIRadioSection): AIRadioPlacement => {
```
stvncode: This heuristic (and `recommendedGuidedSectionIds`) matches English substrings only (`intro`, `outro`, `end`…) — it will misfire for German section names, which this PR itself supports. Fine as a best-effort default, but worth noting — and since this exact function already produced one bug (fixed in eb29f31), it's a prime candidate for a unit test.

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
...
+                    </div>
+                    <div class="mt-1 break-words">
+                      {{ sessionErrorMessage(session) }}
+                    </div>
+                  </div>
+                  <div v-if="session.status === 'running'" class="mt-3">
+                    <Button
+                      variant="outline"
+                      size="sm"
+                      :disabled="stoppingRun"
```
stvncode: `stoppingRun` is a single global flag — with two running sessions, clicking Stop on one shows "Stopping…" and disables the button on **both** rows. Tracking the session id being stopped would fix it.

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
...
+    .split(",")
+    .map((item) => item.trim())
+    .filter(Boolean);
+};
+
+onMounted(async () => {
+  try {
+    await Promise.all([refreshEditor(true), loadStatus(true)]);
+
+    if (!localStorage.getItem(TUTORIAL_SEEN_STORAGE_KEY)) {
```
stvncode: Auto-opening a blocking modal on first visit is a heavy first impression, and afterwards the user still lands on a Run tab with an empty station dropdown and two disabled buttons. Consider an inline empty state ("Create your first station" → wizard) instead, with the tutorial content reachable but not blocking.

src/views/AIRadioView.vue
```diff
@@ -0,0 +1,3571 @@
...
+  draft.id = "";
+  draft.name = "";
+  draft.source_playlist_id = "";
+  draft.source_playlist_provider = "library";
+  selectedEditorStationId.value = "";
+  stationDraft.value = draft;
+  activeTab.value = "stations";
+};
+
+const selectStationForEdit = async (stationId: string) => {
```
stvncode: No unsaved-changes protection: clicking another station in the list (or leaving the page) silently discards draft edits. Since prompts are the main content and can be long, a dirty-state check with a confirm would prevent lost work. Same applies to the section editor.

src/components/navigation/utils/getMenuItems.ts
```diff
@@ -53,6 +54,15 @@ export const getMenuItems = function () {
         isLibraryNode: false,
       });
     }
+    if (enabledMenuItemStr === "ai_radio") {
+      items.push({
+        label: "AI Radio",
```
stvncode: `AppSidebar` renders `t(item.label)`, and every other entry passes a translation key (`party_mode`, `discover`, …). A hardcoded `"AI Radio"` only works because vue-i18n falls back to the key string (with a missing-key warning). Suggest adding a top-level `ai_radio` key like the others.

src/plugins/router.ts
```diff
@@ -93,6 +93,32 @@ const routes: RouteRecordRaw[] = [
         component: () =>
           import(/* webpackChunkName: "discover" */ "@/views/HomeView.vue"),
       },
+      {
+        path: "/ai-radio",
+        name: "ai-radio",
+        component: () =>
+          import(/* webpackChunkName: "ai-radio" */ "@/views/AIRadioView.vue"),
+        beforeEnter: async () => {
```
stvncode, line 139: This guard awaits `INITIALIZED` with no timeout — if the connection never initializes, navigation deadlocks forever. Also the redirect to `discover` when the plugin is disabled is silent; a toast would help users following a stale link understand what happened.

src/components/SendspinPlayer.vue
```diff
@@ -242,12 +242,6 @@ onMounted(() => {
             playerState.value = state.playerState;
           },
           correctionMode: correctionMode.value,
-          onDelayCommand: (delayMs: number) => {
```
stvncode, line 323: This removal looks unrelated to AI Radio — it drops the `onDelayCommand` persistence of `sendspin_static_delay` that exists on main. Merge artifact? Should probably be reverted from this PR.

src/translations/en.json
```diff
@@ -1,9 +1,19 @@
 {
     "actions": "Actions",
+    "add": "Add",
     "add_library": "Add to library",
+    "add_alias": "Add alias",
+    "add_alias_failed": "Failed to add alias",
+    "alias_added_successfully": "Alias added successfully",
+    "add_genre": "Add genre",
```
stvncode: The `en.json` diff contains hundreds of lines of non-ai_radio churn (genre/alias/collection keys that appear manually copied from a newer main rather than arriving via merge). After rebasing onto latest main, the translation diff here should reduce to just the `providers.ai_radio.*` block.

src/translations/de.json
```diff
@@ -1395,9 +1395,6 @@
     "playlist_mix_not_allowed": "Dieser Anbieter unterstützt keine Mischung verschiedener Medientypen in Wiedergabelisten. Bitte wählen Sie den Medientyp aus, für den Sie eine Wiedergabeliste erstellen möchten:",
     "playlist_create_provider_error": "Anbieter nicht erreichbar.",
     "playlist_create_no_type_selected": "Kein Medientyp ausgewählt.",
-    "exclude_genre": "Genre ausschließen",
```
stvncode, line 1099: This **deletes three live keys** (`exclude_genre`, `genre_exclusions`, `remove_genre_exclusion`) that the genre UI on main still uses — German users would see raw keys there if this merges as-is. Should come back with the rebase.

src/layouts/default/ItemContextMenu.vue
```diff
@@ -430,6 +430,29 @@ export const getContextMenuItems = async function (
...
+        router.push({
+          name: "ai-radio",
+          query: {
+            source_playlist_id: firstItem.item_id,
+            source_playlist_provider: firstItem.provider,
+            source_playlist_name: firstItem.name,
+          },
+        });
+      },
+      icon: "mdi-radio-tower",
```
stvncode: Icon consistency: the context-menu entry uses `mdi-radio-tower` while the nav item and the view header use lucide `Sparkles`. Aligning the feature's iconography helps users connect the entry points.

src/plugins/api/interfaces.ts
```diff
@@ -1241,3 +1241,115 @@ export interface PartyConfig {
...
+  session_id: string;
+  station_id: string;
+  mode: AIRadioMode;
+  status: "running" | "completed" | "failed" | "stopped";
+  created_at: string;
+  started_at?: string;
+  ended_at?: string;
+  error?: string;
+  progress?: Record<string, unknown>;
+  result?: Record<string, unknown>;
```
stvncode: `result` is typed but never rendered anywhere — after a playlist run completes, the session card is a dead end and the user has to hunt for the generated playlist in their library. If `result` carries the playlist id, an "Open playlist" link on completed sessions would close the core loop of the feature.

stvncode (approved):

PR #1597, opened 2026-03-16
Add three way genre visibility toggle

jozefKruszynski:


src/components/ItemsListing.vue
```diff
@@ -906,15 +914,19 @@ const menuItems = computed(() => {
...
-        ? "tooltip.show_empty_genres"
-        : "tooltip.hide_empty_genres",
-      icon: params.value.hideEmptyFilter ? EyeClosed : Eye,
+      label:
+        hef === true
+          ? "tooltip.show_only_default_genres"
+          : hef === null
+            ? "tooltip.show_all_genres"
+            : "tooltip.hide_empty_genres",
+      icon: hef === true ? EyeClosed : hef === null ? Star : Eye,
```
stvncode: Not sure the Star is the best icon for that no ?

stvncode (commented):
LGTM, maybe the Star icon is questionnable but other than that it's good 👍

jozefKruszynski:
Agreed re. Star icon, could not think of a better icon for defaults, if you have an idea, I would be all ears

stvncode:
@jozefKruszynski, what do you think of Layers ? 
Screenshot 2026-03-17 at 11 19 11

or maybe LayoutList: 
Screenshot 2026-03-17 at 11 19 58

jozefKruszynski:
Layers might be too close to the music note layers, and layout list seems a bit odd for the use case.

jozefKruszynski:
Claude suggested either star or bookmark. I kinda like the star to be honest.

image

stvncode:
Yeah, so either go with bookmark or maybe bring-to-front idk:
Screenshot 2026-03-17 at 11 34 19

jozefKruszynski:
Layers 2?

https://github.com/user-attachments/assets/664279af-f668-4d0c-93aa-74edccd5b02a

EDIT: Bookmark looked wrong..

MarvinSchenkel:
I think layers is a good option as we partly use it for the genre icon itself so it is recognizable. @stvncode thoughts?

jozefKruszynski:
Actually, regular layers does look better than layers 2

https://github.com/user-attachments/assets/8dff379f-b8dc-465d-8955-3cbd711cc6e0

jozefKruszynski:
> I think layers is a good option as we partly use it for the genre icon itself so it is recognizable. @stvncode thoughts?

This was actually @stvncode 's first suggestion

MarvinSchenkel:
> > I think layers is a good option as we partly use it for the genre icon itself so it is recognizable. @stvncode thoughts?
> 
> This was actually @stvncode 's first suggestion

Ah I missed that.. In that case, I think this is a good option. Shall we use that one?

jozefKruszynski:
Already pushed 😉

Although this PR is reliant on the other on I opened that this branches off

MarvinSchenkel (dismissed):
Thanks @jozefKruszynski !

MarvinSchenkel (approved):

PR #1600, opened 2026-03-17
Improve title screen logic for intro and rewind

apophisnow:
Only show the lyrics title screen when the first LRC timestamp is at least 2s into the song, and re-show it (with lyrics scrolled back out of view) when the user rewinds before the first lyric.


MarvinSchenkel (approved):
Nice 🙏

PR #1606, opened 2026-03-19
Add settings view for background tasks

marcelveldt-traveling:
Adds a new settings view for the new Background Tasks feature.
Depends on https://github.com/music-assistant/server/pull/3426


stvncode (approved):
Damnnn ! I tested it and it's really nice ! And also using the new components, hats off to you <3

PR #1607, opened 2026-03-19
Updates for party mode feature

stvncode:
---                                                                                                                        
  Party Mode UI Improvements                                                                                                 
                                                                                                                             
  This PR enhances the party mode experience across both the guest view and the dashboard view, focusing on polish, usability, and visual consistency.                                                                                         
                                                                                                                           
  ---                                                                                                                        
  Guest View:                                                                                       
                                                                                                                             
  - Logo always visible: The Music Assistant logo is now displayed above the search bar at all times, adapting between light and dark mode (logo.svg / logo-dark.svg)                                                                                   
  - Search UX fixes:                                                                                                       
    - Search input now regains focus after a search completes                                                                
    - Queue is hidden while a search is in progress, only the loading spinner is shown                                      
    - Clearing search now reliably scrolls the queue back to the currently playing item             
  - Back arrow: Changed from invisible to visibly disabled (with reduced opacity) when there's nothing to go back to         
                                                                                                                             
  Guest Components:                                                                                                           
                                                                                                                             
  - PartySearchBar: Removed empty right-side toggle div; exposed focus() method; back arrow properly disabled instead of hidden                                                                                                                   
  - PartyQueueSection: Section header redesigned with a subtle label + divider line layout instead of a heavy title with borders                                                                                                                    
  - PartyQueueItem: Title/artist visually differentiated, hover state and pointer cursor
  only on upcoming (clickable) items                                                                                         
  - PartyResultItem: Artist rows now clickable to drill into tracks; consistent title/artist styling with queue items      
  - PartyTokensBadge: Redesigned as a split-pill badge, solid color left chip showing the count, soft tinted right side with
   the label                                                                                                                 
                                                                                                                             
  Dashboard View:                                                                           
                                                                                                                           
  - QR code: Removed the dark panel background; QR color now follows the theme (black in light mode, white in dark mode)
  Music Assistant logo placed inside the QR at the bottom, also theme-aware
  - Album art background: When album art background is active, all text, QR code, logo, and lyrics are forced white for readability                                                                                                                
  - Lyrics: Text color now follows both the theme and the album art background setting
  - QR availability: QR section is hidden (not just empty) when guest access is disabled                                     
  - PartyTrackCard: Added forceWhiteText prop for album art background override                                              
                                                                                                                             
  Assets                                                                                                                     
                                                                                                                             
  - Added logo-dark.svg for light mode usage (dark-colored text variant)                                                     
  - Moved logo.svg and logo.png into a dedicated assets/logo/ folder


MarvinSchenkel (dismissed):

src/views/PartyDashboardView.vue
```diff
@@ -263,6 +285,10 @@ const fetchLyrics = async () => {
 };
 
 const lyricsEnabled = computed(() => displayLyrics.value || karaokeMode.value);
+const lyricsTextColor = computed(() => {
+  if (useAlbumArtBackground.value && albumArtUrl.value) return "#FFFFFF";
+  return theme.current.value.dark ? "#FFFFFF" : "#000000";
```
apophisnow: I don't think we should change the look of the dashboard based on theme. It is a page that has a specific look and feel, not a general UI that needs to be responsive to user theme preferences. We could redesign it to be theme based, but I think that requires further discussion.

apophisnow (changes requested):

apophisnow (approved):

MarvinSchenkel (approved):

PR #1614, opened 2026-03-20
Use frameless mode for party dashboard instead of new tab

apophisnow:
Read the open_in_new_tab setting from party config to determine whether to open the dashboard in a new tab or navigate in-place with the frameless query param.


apophisnow:
Doesn't really solve anything. The dashboard really only makes sense as a dedicated tab that is view only not mixed with MA navigation etc.

PR #1615, opened 2026-03-21
Add multi-instance party provider support

apophisnow:
- Add instance_id and name to PartyConfig interface
- Route party dashboard by instanceId (/party/:instanceId)
- Update PartyDashboardView, PartyQR, PartyGuestView for per-instance APIs
- Rewrite usePartyConfig composable with per-instance caching and refetchAll
- Add extra_claims to JWTClaims, read party_instance from JWT
- Show context menu for instance selection when multiple party instances exist
- Update PlayerFullscreen to use badge colors from active player's party instance
- Update App.vue party detection for multiple instances
- Add select_party_instance translation key

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -237,20 +237,27 @@ const completeInitialization = async () => {
     console.debug("[App] Party guest - skipping full state fetch");
   }
 
-  // Check if party plugin is enabled
+  // Check party plugin instances
   try {
     const partyProviders = await api.getProviderConfigs(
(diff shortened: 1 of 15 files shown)


apophisnow:
Converting this to draft for now. We will wait on feedback from the current implementation, then decide if we want to implement this, or if some of this could be handled by working on the role/permission system.

apophisnow:
Need to explore integrating this functionality into core permission type logic.

PR #1619, opened 2026-03-22
Add 404 routes

stvncode:
If a user goes to an undefined url, there is nothing to handle it, here's a simple 404 page: Screenshot 2026-03-22 at 11 00 01


marcelveldt (approved):
Nice musical touch !

PR #1620, opened 2026-03-22
Global genre exclusion

jozefKruszynski:
Frontend changes for global genre exclusion enhancement.


src/components/genre/GenreActiveRow.vue
```diff
@@ -0,0 +1,104 @@
+<template>
+  <tr class="group border-b transition-colors hover:bg-muted/50">
+    <td class="pl-6 pr-2 py-3 align-middle">
+      <div class="flex size-8 items-center justify-center">
+        <img
+          v-if="thumbSrc"
+          :src="thumbSrc"
+          class="size-8 rounded object-cover"
```
stvncode: Can you add an alt="xxx" here pls ?

src/components/genre/GenreActiveRow.vue
```diff
@@ -0,0 +1,104 @@
...
+        />
+        <GenreIcon v-else class="size-4 text-muted-foreground" />
+      </div>
+    </td>
+    <td
```
stvncode: Why not use: https://www.shadcn-vue.com/docs/components/data-table for table ?

src/components/genre/GenreExcludedRow.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+  <tr class="group border-b transition-colors hover:bg-muted/50 opacity-50">
+    <td class="pl-6 pr-2 py-3 align-middle">
+      <div class="flex size-8 items-center justify-center">
+        <img
+          v-if="thumbSrc"
+          :src="thumbSrc"
+          class="size-8 rounded object-cover grayscale"
```
stvncode: Add alt="xxx" here too

src/components/genre/GenreTable.vue
```diff
@@ -0,0 +1,367 @@
...
+const scheduleTableReload = () => {
+  if (tableReloadTimer) clearTimeout(tableReloadTimer);
+  tableReloadTimer = setTimeout(async () => {
+    tableReloadTimer = null;
+    await loadData();
+    emit("data-changed");
+  }, 1500);
+};
+
+onMounted(() => {
```
stvncode, line 193: Having a onBeforeUnmount inside an onMounted is not the best practice. You might consider refacto to something like: 
`const unsub = ref<(() => void) | null>(null);
  onMounted(() => { unsub.value = api.subscribe_multi(...); });
  onBeforeUnmount(() => { unsub.value?.(); if (tableReloadTimer) clearTimeout(tableReloadTimer); });  `

stvncode: But it's not an obligation

src/components/genre/GenreTable.vue
```diff
@@ -0,0 +1,367 @@
...
+  allMediaCounts.value = null;
+  try {
+    const hideEmpty =
+      filter.value === "non_empty"
+        ? true
+        : filter.value === "default"
+          ? null
+          : false;
+    [allGenres.value, globalExclusions.value] = await Promise.all([
+      api.getLibraryGenres(
```
stvncode: I'm really not a fan of this. We don't know what we are doing. Would be better to have option and use brackets like: 
`api.getLibraryGenres({idk: undefined, xxx})`

src/components/genre/GenreExcludedRow.vue
```diff
@@ -0,0 +1,70 @@
...
+        <GenreIcon v-else class="size-4 text-muted-foreground" />
+      </div>
+    </td>
+    <td class="px-6 py-3 align-middle font-medium line-through">
+      {{
+        getGenreDisplayName(exclusion.name, exclusion.translation_key, t, te)
+      }}
+    </td>
+    <td
+      v-for="n in 6"
```
stvncode: This should be related to the number of column, if we add a column, this is not working anymore

src/components/genre/GenreTable.vue
```diff
@@ -0,0 +1,367 @@
...
+
+const countFor = (genreId: string, mediaType: string): string => {
+  if (countsLoading.value) return "…";
+  if (!allMediaCounts.value) return "—";
+  return (
+    (allMediaCounts.value[genreId] ?? {})[mediaType] ?? 0
+  ).toLocaleString();
+};
+
+const excludeGenre = async (itemId: string) => {
```
stvncode, line 151: ExcludeGenre removes from allGenres and then re-fetches globalExclusions from the server, while restoreGenre manually updates both allGenres and globalExclusions (without a server round-trip for the active list). Could we k one ? Either always optimistic-update both lists, or always re-fetch both after a mutation.

MarvinSchenkel (approved):

PR #1634, opened 2026-03-23
Update settings ui like settings

stvncode:
Screenshot 2026-03-23 at 17 03 08
Screenshot 2026-03-23 at 17 03 27


MarvinSchenkel (approved):
Beautiful <3

PR #1646, opened 2026-03-26
Fix widget rows reloading when toggling the player bar

MarvinSchenkel:
[Related Issue](https://github.com/music-assistant/support/issues/5141)

The reason is that we watch on the entire preference object of the currentUser, which is mutated when you expand/collapse the player bar, causing the Random albums/artists carousels to reload their content.

This PR narrows down the scope of the watch to the WidgetSettings only.

@stvncode Question before merging this. Do we need this watch at all?

### DIFF
--- a/src/components/HomeWidgetRows.vue
+++ b/src/components/HomeWidgetRows.vue
@@ -111,13 +111,13 @@ onMounted(() => {
 });
 
 watch(
-  () => store.currentUser?.preferences,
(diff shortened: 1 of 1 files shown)


stvncode:
@MarvinSchenkel, if store.currentUser is always guaranteed to be available by the time HomeWidgetRows mount, then no

stvncode (approved):

PR #1648, opened 2026-03-27
Shadcn migration

apophisnow:
Migrate the frontend to shadcn.

Things of note:
- Removes vuetify
- Removes MDI icons (there is currently a mapping file, we should probably rename and remove it eventually) iconMapping.ts
- Seems to work well so far, but this is a draft for a reason. I might have broken a ton of stuff.

TODO:
We need to make sure all UI elements resemble their original counterparts as much as possible. I have not tested many providers so we could have lots of lurking issues there. We should probably review the overall usage patterns as well and make sure it aligns with our desires. Watch out especially for missing or wrong icons, weird layout, messed up animations, etc...


marcelveldt:
I'll let @stvncode comment here but I think his plan was a more staged conversion to migrate component for component while also fixing the UI/UX while at it. But if we can do it instead in a full sweep conversion and then focus on improving components that is also fine maybe. Like I said, I leave the decision with @stvncode here.

stvncode:
I'll copy/paste our discussion on discord: "So, I don’t think this is a good idea to change everything in one time, that’s why I didn’t do it. First because you have to review everything and that we are sure the features stay the same and secondly, we need to make reusable components like DeleteConfirmModal etc. Same for icons. It would be better to have Eriol confirmation for every icons that we change because Lucide sometimes don’t have equivalent of mui icons."
But this is a good start. Will take this work when coming back full time in MA

stvncode:
I'm closing this regarding [this](https://github.com/orgs/music-assistant/projects/2/views/1?pane=issue&itemId=161290156). We should take example of this pr to start the work on the refacto

PR #1651, opened 2026-03-29
Improve FacetedFilter accessibility and clickability (#4883)

teancom:
This started with trying to take care of https://github.com/orgs/music-assistant/discussions/4883 as a way to dip my toe into working on the a11y of the front-end (in combination with https://github.com/music-assistant/desktop-app/pull/32). That took a few rounds of testing to make sure that mouse, keyboard-only, and screen-readers were all working well and then I figured I might as well take care of the whole file while I was there and testing.

- Make entire filter row clickable, not just the checkbox
- Use correct reka-ui prop (model-value) and emit (update:model-value)
- Stabilize v-for key to prevent focus loss for screen readers
- Add aria-labels to icon buttons, search input, badge count, and clear button

The manual test plan I used:

  Setup: Open any page with a FacetedFilter (e.g., Players or Providers settings pages that have filter dropdowns).

  1. Click anywhere in the row (mouse)
  - Open a filter dropdown
  - Click the label text (not the checkbox) — should toggle the filter AND the checkbox visual
  - Click empty space in the row — same result
  - Click the checkbox directly — should also work, no double-toggle

  2. VoiceOver (VO = Control+Option)
  - Open the filter dropdown
  - Navigate to a checkbox with VO+Arrow keys
  - Activate with VO+Space — should toggle the filter and checkbox
  - Focus should stay on that checkbox — you should NOT get kicked out to the page level
  - Toggle a second checkbox without needing Shift+Control+Option+Space to re-enter

  3. Aria labels
  - With VoiceOver on, navigate to:
    - The search input — should announce "Search [FilterName]" not just silence or the placeholder
    - An X button on a badge — should announce "Remove [OptionLabel] filter" not just "button"
    - The X button on the "N selected" badge — should announce "Clear all [FilterName] filters"
    - The "Clear filters" button at the bottom — should announce "Clear [FilterName] filters"
    - The small-screen badge count — should announce "N selected" not just the number

  4. Keyboard only (no VoiceOver)
  - Tab into the popover, use Space/Enter on checkboxes — should toggle correctly
  - Focus should remain stable after toggling


src/components/FacetedFilter.vue
```diff
@@ -42,23 +44,21 @@
...
       <div class="faceted-filter-content">
-        <Input v-model="search" :placeholder="title" class="mb-2 h-8" />
+        <Input v-model="search" :placeholder="title" :aria-label="`Search ${title}`" class="mb-2 h-8" />
         <div class="faceted-filter-list">
           <label
             v-for="option in filteredOptions"
-            :key="`${option.value}-${selectedSet.has(option.value)}`"
+            :key="option.value"
             class="faceted-filter-item"
+            @click.stop.prevent="toggle(option.value)"
```
stvncode: Instead of having a @click.stop.prevent, i would do something like: 
```
<div
    v-for="option in filteredOptions"                                                                                        
    :key="option.value"                                                                                                      
    class="faceted-filter-item"                                                                                              
    role="checkbox"                                                                                                          
    :aria-checked="selectedSet.has(option.value)"
    tabindex="0"
    @click="toggle(option.value)"
    @keydown.space.prevent="toggle(option.value)"                                                                            
  >
    <Checkbox                                                                                                                
      :model-value="selectedSet.has(option.value)"
      class="mr-2 pointer-events-none"
      tabindex="-1"
      aria-hidden="true"                                                                                                     
    />
    <span class="truncate">                                                                                                  
      {{ option.label }}
    </span
```

This way, we have one toggle source and no .stop modfiers. Need to be tested tho

teancom: After some extensive testing this *does* work. Thank you!

stvncode:
@teancom, I just tried to pr and I can't select anything when I use the add x provider filter for example

teancom:
@stvncode I had tested so many variations, I musta just forgot on the final round to try... clicking! VoiceOver and keyboard worked, which is, you know, a passing grade in most places. 😂 Fixed (with two lines, and I can't even blame the robot, that was all me choosing label instead of div before).

stvncode (approved):
Looks good now thanks 🙏

PR #1662, opened 2026-03-30
Add import playlist feature

chrisuthe:
Import m3u and m3u8 playlist files to create a new playlist using functionality in https://github.com/music-assistant/server/pull/3387. This will let users choose which libraries to search to match tracks.


marcelveldt (approved):
Cool. Nice work @chrisuthe !

PR #1666, opened 2026-04-03
Add favorite button to player bar

dmoo500:
Closes https://github.com/orgs/music-assistant/discussions/3830

## What

Adds a favorite (heart) button to the mini player bar at the bottom, using the existing `FavoriteButton` component.

The button:
- Appears on breakpoints `bp3` and larger (same as shuffle/repeat controls)
- Is only shown when a media item is currently in the queue
- Reuses `api.toggleFavorite()` — same behavior as the fullscreen player and list items
- Does not appear on mobile (floating player), where screen space is limited

### DIFF
--- a/src/layouts/default/PlayerOSD/Player.vue
+++ b/src/layouts/default/PlayerOSD/Player.vue
@@ -12,23 +12,30 @@
         />
       </div>
       <div class="mediacontrols-bottom-center">
-        <!-- player control buttons -->
-        <PlayerControls
-          :style="playIconStyle"
-          :visible-components="{
-            repeat: { isVisible: getBreakpointValue('bp3') },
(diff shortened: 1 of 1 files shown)


dmoo500:
Note: There is an existing `PlayerControlBtn/FavoriteBtn.vue` that appears to have been created for use in player controls but is currently not imported anywhere. This PR uses the simpler `FavoriteButton.vue` instead, which is consistent with how it's used in `InfoHeader`, `ListviewItem`, and `PanelviewItem`. `FavoriteBtn.vue` could be cleaned up in a separate PR, or repurposed if a more configurable variant is ever needed in `PlayerControls.vue`.

marcelveldt (approved):
Thanks, @dmoo500 !

PR #1667, opened 2026-04-03
Disable shuffle and repeat buttons for dynamic playlists

dmoo500:
## Summary

Shuffle and repeat have no meaningful effect on dynamic playlists — the provider controls track selection, not MA. The server already skips shuffle logic for dynamic playlists in [server#3527](https://github.com/music-assistant/server/pull/3527), but the buttons in the UI remained enabled regardless.

## Changes

- **\`interfaces.ts\`**: Added \`is_dynamic: boolean\` to the \`Playlist\` interface and \`enqueued_media_items: MediaItemType[]\` to the \`PlayerQueue\` interface — both fields exist in \`music-assistant-models v1.1.109\` but were missing from the frontend type definitions
- **\`ShuffleBtn.vue\`**: Disabled when \`enqueued_media_items\` contains a \`Playlist\` with \`is_dynamic=true\`
- **\`RepeatBtn.vue\`**: Same — disabled for dynamic playlists

## Related

- [server#3527](https://github.com/music-assistant/server/pull/3527) — queue controller: dynamic playlist auto-refill (server-side shuffle already skipped there)
- [server#3555](https://github.com/music-assistant/server/pull/3555) — Infinite Mix built-in playlists (use \`is_dynamic=True\`, benefit directly from this fix)
- [server#3433](https://github.com/music-assistant/server/pull/3433) — Apple Music provider: track-based stations mapped to \`Playlist(is_dynamic=True)\`
- [discussion#5129](https://github.com/orgs/music-assistant/discussions/5129) — original feature discussion

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerControlBtn/RepeatBtn.vue
+++ b/src/layouts/default/PlayerOSD/PlayerControlBtn/RepeatBtn.vue
@@ -4,7 +4,10 @@
     v-if="isVisible && playerQueue"
     v-bind="{ ...icon, ...$attrs }"
     :disabled="
-      !playerQueue.active || playerQueue.items == 0 || isLoading || isDynamic
+      !playerQueue.active ||
+      playerQueue.items == 0 ||
+      isLoading ||
+      isSingleDynamicPlaylist
     "
     :color="
       getValueFromSources(icon?.color, [
@@ -70,12 +73,12 @@ const isLoading = computed(() => {
   );
 });
 
-const isDynamic = computed(() => {
+const isSingleDynamicPlaylist = computed(() => {
+  const items = compProps.playerQueue?.enqueued_media_items;
   return (
-    compProps.playerQueue?.enqueued_media_items?.some(
-      (item) =>
-        item.media_type === MediaType.PLAYLIST && (item as Playlist).is_dynamic,
-    ) ?? false
+    items?.length === 1 &&
+    items[0].media_type === MediaType.PLAYLIST &&
+    (items[0] as Playlist).is_dynamic
   );
 });
 </script>
--- a/src/layouts/default/PlayerOSD/PlayerControlBtn/RepeatBtn.vue
+++ b/src/layouts/default/PlayerOSD/PlayerControlBtn/RepeatBtn.vue
@@ -3,7 +3,9 @@
   <Icon
     v-if="isVisible && playerQueue"
     v-bind="{ ...icon, ...$attrs }"
-    :disabled="!playerQueue.active || playerQueue.items == 0 || isLoading"
+    :disabled="
+      !playerQueue.active || playerQueue.items == 0 || isLoading || isDynamic
+    "
     :color="
       getValueFromSources(icon?.color, [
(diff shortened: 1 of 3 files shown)


marcelveldt:
This is not entirely true...
You can add multiple playlists to the queue.

For example add 2 normal playlists and also add one or more dynamic playlists (or media items in radio mode).
So in this case it stil makes sense to shuffle as you want to the total to be a mix.

A queue is not the same as a playlist - a queue is a combination of items. A result of the (expanded) media items you added to it.

src/layouts/default/PlayerOSD/PlayerControlBtn/RepeatBtn.vue
```diff
@@ -62,4 +69,13 @@ const isLoading = computed(() => {
     compProps.playerQueue?.extra_attributes?.play_action_in_progress === true
   );
 });
+
+const isDynamic = computed(() => {
+  return (
+    compProps.playerQueue?.enqueued_media_items?.some(
```
marcelveldt: see my previous reply. I think this only makes sense if the queue is ONLY playing one single dynamic playlist

dmoo500:
Good point, and you're right that a queue can contain a mix of dynamic and non-dynamic sources.

The current implementation already handles this: shuffle and repeat are only disabled when `enqueued_media_items.length === 1` **and** that single item is a `Playlist` with `is_dynamic=true`. So a mixed queue — e.g. two normal playlists plus a dynamic one — correctly leaves both buttons enabled.

The rationale for the single-dynamic-playlist case: [server#3527](https://github.com/music-assistant/server/pull/3527) already skips the shuffle step on the server side for dynamic playlists, and the tracks are randomly selected at the source on every refill anyway. Visually disabling the button just reflects what's actually happening. Same applies to repeat — auto-refill takes over, so repeat has no effect.

Happy to drop the feature entirely if you feel even the single-dynamic-playlist case doesn't warrant it.

marcelveldt:
In that case I have misread it - all good.

src/layouts/default/PlayerOSD/PlayerControlBtn/ShuffleBtn.vue
```diff
@@ -3,7 +3,9 @@
   <Icon
     v-if="isVisible && playerQueue"
     v-bind="{ ...icon, ...$attrs }"
-    :disabled="!playerQueue.active || playerQueue.items == 0 || isLoading"
+    :disabled="
+      !playerQueue.active || playerQueue.items == 0 || isLoading || isDynamic
```
marcelveldt:
```suggestion
      !playerQueue.active || playerQueue.items == 0 || isLoading || isSingleDynamicPlaylist
```

dmoo500: Good catch — the `:disabled` binding in the template was still referencing the old `isDynamic` name after the computed was already renamed to `isSingleDynamicPlaylist`. Fixed and pushed.

marcelveldt (approved):
Thanks, @dmoo500 !

PR #1669, opened 2026-04-04
Add track action menu to player bar

dmoo500:
Adds a ⋯ menu button in the player bar (between queue button and volume slider) that appears when a Track is playing.

**Actions:**
- Add to Playlist
- Show info (navigates to track detail page)
- Start Radio (only shown when the provider supports `SIMILAR_TRACKS`)

Relates to https://github.com/orgs/music-assistant/discussions/3830

### DIFF
--- a/src/layouts/default/PlayerOSD/Player.vue
+++ b/src/layouts/default/PlayerOSD/Player.vue
@@ -254,7 +254,6 @@ watch(
 }
 
 .mediacontrols-left {
-  flex: 1 1 0;
   min-width: 0;
   display: flex;
   align-items: center;
--- a/src/layouts/default/PlayerOSD/Player.vue
+++ b/src/layouts/default/PlayerOSD/Player.vue
@@ -52,6 +52,7 @@
             }"
             :volume="{
               isVisible: store.activePlayer != undefined,
+              volumeSize: getBreakpointValue('bp8') ? '150px' : '100px',
             }"
           />
         </div>
--- a/src/layouts/default/PlayerOSD/Player.vue
+++ b/src/layouts/default/PlayerOSD/Player.vue
@@ -10,16 +10,6 @@
           :color-palette="coverImageColorPalette"
           :primary-color="$vuetify.theme.current.dark ? '#fff' : '#000'"
         />
-        <!-- favorite button for current track -->
-        <FavoriteButton
-          v-if="
-            store.curQueueItem?.media_item &&
-            getBreakpointValue('bp7') &&
-            store.activePlayer?.powered !== false
-          "
-          :item="store.curQueueItem.media_item"
-          class="mediacontrols-favorite"
-        />
       </div>
       <div class="mediacontrols-bottom-center">
         <!-- player control buttons -->
@@ -134,7 +124,6 @@ import { getBreakpointValue } from "@/plugins/breakpoint";
 import { store } from "@/plugins/store";
 import vuetify from "@/plugins/vuetify";
 import { computed, ref, watch } from "vue";
-import FavoriteButton from "@/components/FavoriteButton.vue";
 import PlayerControls from "./PlayerControls.vue";
 import PlayerExtendedControls from "./PlayerExtendedControls.vue";
 import PlayerTimeline from "./PlayerTimeline.vue";
@@ -270,14 +259,11 @@ watch(
   display: flex;
   align-items: center;
   gap: 4px;
+  margin-inline-end: auto;
   > div {
     padding: 0px !important;
     min-width: 0;
   }
-}
-
-.mediacontrols-favorite {
-  flex-shrink: 0;
 }
 
 .mediacontrols-bottom-right {
--- a/src/layouts/default/PlayerOSD/Player.vue
+++ b/src/layouts/default/PlayerOSD/Player.vue
@@ -267,10 +267,8 @@ watch(
 }
 
 .mediacontrols-bottom-right {
-  flex: 1 1 0;
+  margin-inline-start: auto;
   min-width: 0;
-  display: flex;
-  justify-content: flex-end;
   > div {
     display: inline-flex;
     align-items: center;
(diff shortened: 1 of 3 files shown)


marcelveldt:
Can you add some screenshots of how this looks like ?
The player bar is already (very) cramped with controls so we try to minimize adding more. 
Also there is a UX discussion going on about the greater picture of improving the UX of the playerbar and fullscreen player

marcelveldt:
One idea could be to also pack the favoriting in this quick menu and another is that this should maybe only show up at bp7 and higher

dmoo500:
Here are 5 screenshots taken at the minimum desktop width (just before the mobile player kicks in, ~575px / bp3):

Player bar – favorite btn implemented but hidden by size (merged via #1666), menu closed (screenshot 1)
Screenshot 2026-04-04  18 33

Player bar – track playing, menu closed, favorite bun removed (screenshot 2)
Screenshot closed menu 2026-04-04 18 08

Player bar – track playing, menu open, favorite bun removed (screenshot 3)
Screenshot open menu 2026-04-04 18 08

Inspector view – no track selected,  favorite bin removed (screenshot 4)
Screenshot Sizes 2026-04-04 18 18

Inspector view – track playing,  favorite bin removed(screenshot 5)
Screenshot Sizes 2026-04-04 18 22


Observations after rebasing onto latest main (which includes #1666):

At the minimum desktop width (bp3, ~575px), the FavoriteButton (from #1666) is now hidden as expected (it's set to bp8 / 1300px). However, the Repeat button overlaps/clips with the SpeakerBtn button at this narrow width — both appear at bp3. This is the crowding issue you mentioned.

Two options to address this:

* **Option A**: Show the ⋯ menu only from bp7 (1100px) and higher — same as your suggestion. At that width there is enough room for all controls.
* **Option B**: Move Favorite into the ⋯ menu (as you suggested in the other comment) and keep the menu visible at bp3 but without a separate heart icon — net zero extra controls.

Happy to go with whichever direction you prefer. Let me prepare Option B - from my side it makes more sense

marcelveldt:
Yeah option B makes most sense!

dmoo500:
Thanks for the feedback! I've updated the PR with the following changes:

Implemented Option B — instead of keeping a standalone heart icon in the player bar, the Favorite action is now included as the first item in the ⋯ menu. This keeps the player bar compact and groups all track-related actions in one place:

* ♥ / ♡ Favorite (toggles, reflects current state)
* ➕ Add to Playlist
* ℹ️ Show Info
* 📡 Start Radio (only shown when provider supports SIMILAR_TRACKS)

The ⋯ button only appears when a Track is playing — not for radio streams or an empty queue. The queue button's padding adjusts dynamically so spacing stays consistent whether or not the menu is visible.

Screenshot 2026-04-04 20 16

marcelveldt:
I think the speaker button should stay next to the volume

dmoo500:
Found a CSS issue introduced by this PR itself: as I moved the ⋯ button to the possition between FullPage Player Btn and the volume slider, I accidentally changed `.mediacontrols-bottom-right` from `margin-inline-start: auto` to `flex: 1 1 0; display: flex; justify-content: flex-end` — reverted the class back to its original CSS.

The SpeakerBtn was always rendered immediately to the left of the FullPage Player Btn. The correct order from left to right is: SpeakerBtn → FullPage Player Btn → ⋯ menu (new!) → volume slider.

Screenshot 2026-04-04 21 03

marcelveldt:
Yeah, my issue is that the player button is now too far from the volume slider, hence I asked to move it right.
I think the most natural is to have the new contextmenu icon at the left as that is related to the track

dmoo500:
Moved the ⋯ menu to the leftmost position in the right-side controls, before the `SpeakerBtn`. The correct order from left to right is now: ⋯ menu → `SpeakerBtn` → Full Page Player Btn (QueueBtn) → volume slider. Also simplified the QueueBtn padding back to a static value since the dynamic padding logic is no longer needed.

screenshot 2026-04-04 21 43

dmoo500:
Setting back to draft while I work through a CSS layout issue — the ⋯ menu overlaps the center transport controls at narrow widths. Will re-request review once it's sorted.

dmoo500:
CSS layout issue resolved. The ⋯ menu no longer overlaps the center transport controls — fixed by reducing the volume slider width at narrower widths (`100px` below 1300px, `150px` above) so all right-side controls fit without overflow. Also equalized the spacing between all three right-side buttons to a consistent 15px.

marcelveldt (approved):
Nice work, thanks @dmoo500 !

OzGav:
Looking at the nightly the favourite selection doesn’t seem to be working correctly for mine. If a track page is visited, for example, the banner at the top has a heart icon to indicate favourite status. In the menu top right there is an item “Add to favourites” or “remove from favourites” which aligns with the indication in the banner. However, the player bar menu shows “Mark/unmark as favourite” or “remove from favourites” so this is inconsistent with the other menu and “mark/unmark as favourite” is being shown regardless of the favourite status set above and does not appear to be actually changing the favourite status of the item, although the player bar menu is changing.

OzGav:
Amendment to the above. Maybe the confusion I am experiencing is just down to the menu item. Why is it “mark/unmark as favourite” as it seems to be always marking as favourite when that is selected?

dmoo500:
Thanks for testing on the nightly! Both points are valid:

**Label issue:** The menu was using the wrong translation key (`tooltip.favorite` = "Mark/unmark as favourite") instead of the state-aware `favorites_add` / `favorites_remove`. Fixed in #1674.

**Favourite not reflecting elsewhere:** The toggle does work correctly — the menu label updates immediately after clicking. However, if you check the track detail page afterwards, the heart icon there may not update until the next server event, because it references a different object (from the library) than the one in the queue. This is pre-existing behaviour and not introduced by this PR.

PR #1670, opened 2026-04-04
Party duplicate prevention

apophisnow:
Adds feature to prevent duplicate track submissions by guests.
Defaults optional text on dashboard page to blank, and fixes a problem where it could not be set blank.
Fix auto selected player not being updated for guest UI.


marcelveldt (approved):
Thanks, @apophisnow !

PR #1671, opened 2026-04-05
Improve performance with GPU compositing and lyrics guard

apophisnow:
Use transform: scaleX() for the fill bar and translateX() for the thumb instead of modifying width/left properties, avoiding layout recalculations on every frame during playback.

Guard against lyrics raf loop when lyrics are not visible.


src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -1,49 +1,62 @@
 <template>
   <div style="width: auto; height: 24px">
     <div v-if="store.activePlayer" style="width: 100%">
-      <v-slider
```
stvncode, line 4: Since we removed the v-slider, we do not have anymore ARIA attributes and keyboard navigation

src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -247,6 +264,24 @@ const chapterTicks = computed(() => {
...
 
+const duration = computed(
+  () => store.activePlayer?.current_media?.duration || 1,
+);
+
+const progress = computed(() => {
+  return Math.min(Math.max(curTimeValue.value / duration.value, 0), 1);
+});
+
+const thumbX = computed(() => {
```
stvncode: ThumbX relies on non-reactive clientWidth. Thumb position probably won't update on container resize

src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -287,6 +350,73 @@ const chapterClicked = function (chaperPos: number) {
...
+  opacity: 0.5;
+}
+
+.timeline-rail {
+  position: absolute;
+  left: 0;
+  right: 0;
+  height: 4px;
+  border-radius: 2px;
+  background: rgba(255, 255, 255, 0.2);
```
stvncode: What about light mode ?

src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -247,6 +264,24 @@ const chapterTicks = computed(() => {
...
+});
+
+const thumbX = computed(() => {
+  if (!trackRef.value) return 0;
+  return progress.value * trackRef.value.clientWidth;
+});
+
+const fillColor = computed(() => {
+  if (props.color) return props.color;
+  return theme.current.value.colors.fg;
```
stvncode: Since we are trying to remove vuetify elements, can we not use this and this our them color instead ?

src/layouts/default/PlayerOSD/PlayerTimeline.vue
apophisnow: Added aria attributes and improved keyboard handling.

src/layouts/default/PlayerOSD/PlayerTimeline.vue
```diff
@@ -72,19 +90,21 @@ export interface Props {
   color?: string;
 }
 
-withDefaults(defineProps<Props>(), {
+const props = withDefaults(defineProps<Props>(), {
```
stvncode: Don't think it's being used

stvncode (approved):
Nice 🙏

PR #1672, opened 2026-04-05
Isolate queue elapsed time from main reactive queue objects

apophisnow:
QUEUE_TIME_UPDATED events were mutating elapsed_time directly on the reactive queue objects, triggering re-renders in every component reading activePlayerQueue (~158 references across 9 OSD files).

Introduce a separate api.queueElapsedTime reactive map that only subscribers to elapsed time will track. This prevents high-frequency time updates from dirtying the broad queue object and cascading unnecessary DOM commits and layerization.

Cuts down on CPU use by approximately half (combined with the other performance PR)
Screenshot 2026-04-04 at 10 18 46 PM
Screenshot 2026-04-04 at 10 22 37 PM

### DIFF
--- a/src/components/SendspinPlayer.vue
+++ b/src/components/SendspinPlayer.vue
@@ -317,7 +317,9 @@ onMounted(() => {
       const targetId = getTargetPlayerId();
       if (!targetId) return;
       const offset = evt.seekOffset || 10;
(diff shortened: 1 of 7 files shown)


stvncode (approved):
Really nice thanks ! 🙏

PR #1674, opened 2026-04-05
Extend player bar track menu to also show for radio stations

dmoo500:
## Summary

Extends the `⋯` action menu in the player bar to be visible when a **radio station** is playing, not just when a Track is playing.

### Changes

- Menu guard changed from `currentTrack` to `currentItem` (Track or Radio)
- Added `currentRadio` computed (resolves `MediaType.RADIO` from the current queue item)
- Added `currentItem` computed (`currentTrack ?? currentRadio`)
- **Favorite** and **Show Info** work for both Track and Radio
- **Add to Playlist** and **Start Radio** remain Track-only (guarded with `v-if="currentTrack"`)
- `onShowInfo` now routes to the correctly typed detail page (`radio` route for stations)
- `onToggleFavorite` now calls `api.toggleFavorite(currentItem)` which accepts `MediaItemType`

### Why

When playing a radio stream, there was no way to favorite the station from the player bar — you had to navigate away to the station's detail page. This extends the existing menu to cover that case with minimal changes.

Note: "Start Radio" (Radio Mode) for a Radio station is intentionally not supported — `radio_mode_base_tracks` raises `NotImplementedError` for Radio items on the server side.


src/layouts/default/PlayerOSD/PlayerControlBtn/PlayerTrackMenu.vue
```diff
@@ -84,23 +93,26 @@ const radioModeSupported = computed(() => {
 });
 
 const onAddToPlaylist = () => {
```
stvncode, line 95: Revert to currentTrack to match the Track-only intent:                                                
  
  ```const onAddToPlaylist = () => {                                                                                            
    if (!currentTrack.value) return;                                                                                         
    eventbus.emit("playlistdialog", { items: [currentTrack.value] });
  }; ```

stvncode (commented):
Add to Playlist should remain Track-only but we don't have any v-if="currentTrack"  for the list-item line 25.

src/layouts/default/PlayerOSD/PlayerControlBtn/PlayerTrackMenu.vue
dmoo500: This is intentional — `MediaType.RADIO` is already supported by the playlist dialog on the server side (`ItemContextMenu.vue` line 682 explicitly lists `MediaType.RADIO` for "Add to Playlist"). Keeping currentItem here means a radio station playing in the queue can be added to a playlist directly from the player bar, without navigating to the station detail page. OzGav confirmed this use case in a comment on PR #1669.

dmoo500: "Add to Playlist" on the `<v-list-item>` was intentionally left enabled for radio items — tested on the radio detail view and it works correctly: the playlist dialog accepts `Radio` items and the server-side handler supports `MediaType.RADIO`. The `v-if="currentTrack"` guard on the surrounding block already hides "Start Radio" (radio mode) for stations, which is the Track-only action.

stvncode (approved):
Thanks 🙏

PR #1681, opened 2026-04-06
Replace deprecated unload event with pagehide

apophisnow:


src/plugins/web_player.ts
```diff
@@ -109,7 +109,7 @@ bc.onmessage = (event) => {
 };
 
 // Called on close
```
stvncode: pagehide fires on bfcache entry (event.persisted === true), which means the cleanup code runs even when the page may be restored. This could leave a restored tab with a dead BroadcastChannel listener and prematurely released playback control.
We could do soemthing like this: 
```
window.addEventListener("pagehide", function (event) {
    if (event.persisted) return;                                                               
    bc.onmessage = null;                                                                                                     
    if (isPlaybackMode(webPlayer.tabMode) && webPlayer.player_id) {                                                          
      bc.postMessage(BC_MSG.CONTROL_AVAILABLE);
    }                                                                                                                        
  }); 
```

stvncode (approved):
Thanks ! 🙏

PR #1682, opened 2026-04-06
Use swiper virtual slides in Carousel

apophisnow:
Refactor Carousel to accept items prop with scoped slot and enable Swiper's Virtual module so only visible + nearby slides are rendered.


stvncode (approved):
Nice!

PR #1683, opened 2026-04-06
Add virtual scrolling for panel/grid views using TanStack Virtual

apophisnow:


src/components/ItemsListing.vue
```diff
@@ -376,6 +396,40 @@ const initialDataReceived = ref(false);
 const tempHide = ref(false);
 const genreOptions = ref<{ label: string; value: number }[]>([]);
 
+// virtual scroll setup for panel views
+const scrollContainerRef = ref<HTMLElement | null>(null);
```
stvncode, line 405: Could we use this ref as a ```ref<HTMLElement | null>(null)``` instead of a ```document.querySelector('.content-section')``` ?

src/components/ItemsListing.vue
```diff
@@ -376,6 +396,40 @@ const initialDataReceived = ref(false);
...
+  const items = pagedItems.value;
+  const rows: MediaItemType[][] = [];
+  for (let i = 0; i < items.length; i += cols) {
+    rows.push(items.slice(i, i + cols));
+  }
+  return rows;
+});
+
+const panelRowHeight = computed(() => {
+  return viewMode.value === "panel_compact" ? 180 : 260;
```
stvncode, line 423: Why use harcoded value here ? Could we not match the actual rendered row heights ? If they don't match, the virtualizer will produce layout jumps as it measures real elements

src/components/ItemsListing.vue
```diff
@@ -1372,11 +1444,17 @@ const loadGenreOptions = async () => {
 
 let _unsubscribeMediaEvents: (() => void) | undefined;
 onBeforeUnmount(() => {
+  scrollContainerRef.value?.removeEventListener("scroll", checkInfiniteScroll);
```
stvncode, line 1456: Can you add { passive: true } to both addEventListener and removeEventListener pls ?It's recommended for scroll listeners to avoid jank

src/components/ItemsListing.vue
```diff
@@ -64,67 +64,85 @@
...
+            height: `${virtualRow.size}px`,
+            transform: `translateY(${virtualRow.start}px)`,
+          }"
+        >
+          <v-row>
+            <v-col
+              v-for="item in panelRows[virtualRow.index]"
+              :key="item.uri"
+              cols="12"
+              :class="`col-${panelViewItemResponsive($vuetify.display.width)}`"
```
stvncode: This is called inside the v-for template via :class binding. Since this is already computed via columnCount, maybe we can use columnCount.value here instead of re-calling the function for every column in every virtual row

src/components/ItemsListing.vue
stvncode: Could be nice to use TanStack Virtual's measureElement

stvncode (commented):
Love tanstack virtualizer! 
On top of that, can you add aria-rowcount/aria-rowindex for screen reader support on the virtualized list pls ?

stvncode:
@apophisnow, sorry for the late answer... FInally have time to review old prs. I just switch everything to pnpm and there is conflicts here. Would you prefer to create a new pr or to update this one ?

OzGav:
Marking this as draft until the comments and conflicts are resolved

PR #1685, opened 2026-04-07
Add edit radio station feature

OzGav:
DEPENDS ON: https://github.com/music-assistant/server/pull/3602

Allows editing of custom (builtin provider) radios, tracks, and playlists from the item details page context menu. Reuses the existing AddManualLink dialog in edit mode.


src/components/AddManualLink.vue
```diff
@@ -131,25 +151,63 @@ const fetchItemDetails = () => {
...
+            },
+          ],
+        };
+      } else {
+        updatedRadio.metadata = {
+          ...updatedRadio.metadata,
+          images: [],
+        };
+      }
+      await api.updateRadio(
```
marcelveldt: why aren't we using the update_item endpoint ?
updateRadio could be a helper method but it doesn't belong in the api class then

OzGav: The updateRadio method has been removed from the API class. The component now calls api.sendCommand("music/radios/update", {...}) directly.

P.S. I now have to add frontend support for tracks and playlists...

marcelveldt (approved):
Thanks, @OzGav !

PR #1694, opened 2026-04-11
Add artist track list sort by album sort name

OzGav:
Satsfies https://github.com/orgs/music-assistant/discussions/5260#discussioncomment-16501258

We already have alphabetical and by sort name in the album section so this just enables the sort name option in the tracks list


marcelveldt (approved):
Thanks, @OzGav !

PR #1699, opened 2026-04-12
Pass sort order to server for play from here

philjackson:
Threads the current sortBy state through the UI so that "Play Playlist From Here", "Play Album From Here", and the playlist header play button all respect the user's chosen sort order.

- API client: adds sort_by parameter to playMedia
- ItemsListing: exposes sortBy via defineExpose
- PlaylistDetails: reads sort state and passes to InfoHeader
- InfoHeader: forwards sortBy to handlePlayBtnClick
- handlePlayBtnClick: always uses play-from-here for tracks in playlists/albums (removes stale queue-state guard), passes sortBy on all playMedia paths
- Context menu: passes sortBy in play-from-here actions
- Includes 6 frontend tests for sortBy pass-through

Ref: music-assistant/support#4838
Partner PR: https://github.com/music-assistant/server/pull/3663


marcelveldt:
Like discussed in the backend PR, at some point we should get rid of this frontend-side sorting for these lists and just use server-side paging and sorting everywhere (like the library views) so we can also use that for the playback.

But that is a big refactor/change which can be done later.
I think your approach is completely fine as an intermediate step.
Nice work!

marcelveldt (approved):

PR #1710, opened 2026-04-13
Reload queue items after reset in fullscreen player

dmoo500:
## Problem

When the queue is refilled (e.g. for a dynamic playlist/radio station), a `QUEUE_ITEMS_UPDATED` event triggers `resetItems()`, which clears `queueItems` to `[]`. After the 100ms hide animation, `tempHide` is set back to `false` but no reload is triggered — leaving the queue panel empty until the user scrolls or toggles the view.

## Fix

After clearing items in `resetItems()`, call `loadNextPage()` if the fullscreen player and queue panel are currently visible.


marcelveldt (approved):
Thanks, @dmoo500 !

PR #1711, opened 2026-04-13
Hide player in more places

OzGav:
Follow up to https://github.com/music-assistant/frontend/pull/1686

I missed something the first time.

- If a hidden player is currently part of a group, it still shows up in that group's volume control so you can see it's playing and adjust its volume.
- If a hidden player isn't in a group, it's left out of the lists for adding it to one — both the sync picker in the volume control and the "synchronize with" option in the player menu.


src/components/VolumeControl.vue
```diff
@@ -201,6 +201,9 @@ const getChildPlayers = function (player: Player) {
       // skip unavailable players
       if (!syncPlayer.available) continue;
 
+      // skip hidden players
+      if (syncPlayer.hide_in_ui) continue;
```
marcelveldt, line 207: I believe this is incorrect.
the syncPlayer should still show up if it is grouped/groupable

The note on the setting also mentions that the player will still show up as a sync child

OzGav: Even if they have have set hide this player in the UI
. That doesn’t seem right to me?

marcelveldt: That would cause weird artifacts - what if a player is joined to another one or part of a group and you have no idea that its there, participating in the group ?

The filter would need to be a bit more subtle;
- Either just only hide it in the main players list (and equivalents such as the "play on" dropdown etc
- Or only hide it if not participating in the group

Why else would you want to hide the player ? That setting has always been there just to hide the player in the main overview, not hide it everywhere (or prevent access).

If you dont want to see and interact with the player, why not just disable the player?

OzGav: Dropped this so hidden sync children stay visible. Also add a description to the setting in the server so it is clear what this setting does.  https://github.com/music-assistant/server/pull/3792

OzGav: I realised that hidden players were still showing in the list of players available to manually sync, so I've reapplied `hide_in_ui` only there — current group members stay visible. This is essentially your "only hide it if not participating in the group" option from above.

stvncode (approved):

MarvinSchenkel:
This PR breaks Hue lights from showing up in the frontend. I will have a look if I can come up with a fix for this. @maximmaxim345 fyi

PR #1725, opened 2026-04-22
Fix favorite heart state resetting on queue updates for radio and provider items

dmoo500:
## Problem

When a radio station or provider track (not yet in the library) was playing and the user favorited it, the heart button in the fullscreen player would briefly light up and then reset to empty. This happened on every subsequent queue update from the server.

**Root cause:** `store.curQueueItem` is derived from `activePlayerQueue.current_item`, which the server pushes as a complete replacement object on every queue event. When the item was enqueued with `favorite: false`, all server-side queue updates kept sending that original value — overwriting any optimistic local change made by `toggleFavorite`.

Library tracks that were already favorited *before* being enqueued were not affected because their `favorite: true` was baked in at enqueue time.

## Fix

Introduce a local `currentItemFavorite` ref that lives independently of the server-pushed `media_item` object:

- **Initialised** from `media_item.favorite` whenever the queue item changes (watch on `queue_item_id`)
- **Optimistically updated** to `true`/`false` in `onHeartBtnClick` — survives server queue refreshes because it is not part of the replaced object
- **Synced from server** via a `MEDIA_ITEM_UPDATED` subscription: when the server confirms the favorite change (URI match), `currentItemFavorite` is updated to the server's value

The template heart fill and the "already favorited" branch in `onHeartBtnClick` now read `currentItemFavorite` instead of `store.curQueueItem?.media_item?.favorite`.


stvncode (approved):
Sorry for the late answer on this one... LGTM thanks ✅

PR #1727, opened 2026-04-22
Prune stale provider ids from stored listing filters

OzGav:
I couldn't test this but I got the original reporter to do so https://github.com/music-assistant/support/issues/5368#issuecomment-4297950504


OzGav:
@stvncode would be good to get this in for 2.9 as well.

stvncode:
Pruning against loaded providers (api.providers) can permanently delete saved filters for providers that are merely failing to load (auth/network errors at boot), not removed. Prune against provider configs instead, or verify/document that configured-but-broken providers always remain in api.providers with available: false.

OzGav:
fixed. You're right that `api.providers` only holds loaded instances, so a provider that's disabled or still starting up isn't in there and would've had its saved filter wrongly deleted.
  
I've switched the prune to run against provider configs (`getProviderConfigs()`) instead. A genuinely removed provider has no config and gets pruned. A configured-but-broken/loading one keeps its config, so its filter is preserved. Added a guard to skip pruning if the config fetch fails or returns empty.

Also stopped the runtime watch in `ItemsListing.vue` from persisting its prune. It now only updates the live query for an instant refresh, while persisted cleanup is handled centrally by the config-based prune. That way a temporarily-unavailable provider keeps its saved filter and gets it back when it returns

stvncode (approved):
Thanks for the changes, lgtm now ✅

PR #1728, opened 2026-04-23
Perf/gpu timeline slider

apophisnow:
This adds a progress-bar component that uses gpu compositing, which is good for the constantly updating timeline slider to reduce cpu usage.

### DIFF
--- a/src/components/ui/progress-bar/ProgressBar.vue
+++ b/src/components/ui/progress-bar/ProgressBar.vue
@@ -1,0 +1,266 @@
+<template>
+  <div
+    ref="trackRef"
+    class="progress-bar-track"
+    :class="{ 'progress-bar-track--disabled': disabled }"
+    role="slider"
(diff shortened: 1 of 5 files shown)


MarvinSchenkel:
Could you please make sure that all the paddings etc are exactly the same. Just fire up two tabs, one pointing to :3000 and one pointing to :8095. Both the home screen and the fullscreen mode show a lot of padding differences. 

It's hard to show in a screenshot but when you switch tabs side by side, you will quickly notice it.

apophisnow:
> Could you please make sure that all the paddings etc are exactly the same. Just fire up two tabs, one pointing to :3000 and one pointing to :8095. Both the home screen and the fullscreen mode show a lot of padding differences.
> 
> It's hard to show in a screenshot but when you switch tabs side by side, you will quickly notice it.

Yeah I see what you mean. I also found a bug where if the element redraws it won't show the current position until an update.

PR #1755, opened 2026-05-07
Add Lyrics offset functionality

OzGav:
Satisfies https://github.com/orgs/music-assistant/discussions/4864 

Adds a button to be able to adjust the sync timing of the lyrics. The button only appears when tht elyrics tab is selected. The arrow icon then shows and hides the adjustment controls and value.

image

### DIFF
--- a/src/components/LyricsViewer.vue
+++ b/src/components/LyricsViewer.vue
@@ -97,11 +97,40 @@
         </div>
       </div>
     </ScrollArea>
+
+    <!-- Synced lyrics offset slider (proof of concept) -->
+    <div
+      v-if="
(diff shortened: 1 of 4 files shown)


stvncode (approved):
This is really nice ! I took the liberty to refactor button and icon to shadcn-vue and lucide. Also, it would be nice at some point that we create a single component for the lyrics part, the component is 2k lines now 😢

OzGav:
Thanks Steven for the improvement! We will see in the beta if people say that even AirPlay needs the adjustment functionality. As for the refactor I agree, will add it to the to do list!

PR #1758, opened 2026-05-09
Remove border-radius from sidebar header logo

remon1496:
6px sidebar-header-logo cut of the MA header logo. Removing the border-radius resolves this.


stvncode (approved):
Thanks !

PR #1765, opened 2026-05-11
Add trending_stations translation key for TuneIn recommendations

dmoo500:
Adds the `trending_stations` translation key under `recommendations` in `en.json`, required by the companion server PR (https://github.com/music-assistant/server/pull/3865) that adds `translation_key="trending_stations"` to the TuneIn recommendations folder.


OzGav (approved):
Thanks!

PR #1768, opened 2026-05-13
Toggle homescreen edit mode label in user menu

dmoo500:
## Problem

The "Edit Home screen" menu item in the user dropdown always showed the same label regardless of whether the homescreen was currently in edit mode or not. The translation key `homescreen_edit_disable` ("Leave Home screen edit mode") already existed but was never used.

## Solution

Track the homescreen edit mode state in the global store (`homescreenEditMode: boolean`) so that `NavUser.vue` can reactively switch the menu item label between:
- `homescreen_edit_enable` → "Edit Home screen" (when not in edit mode)
- `homescreen_edit_disable` → "Leave Home screen edit mode" (when in edit mode)

Both translation keys were already present in all language files — no new translations needed.

### DIFF
--- a/src/components/navigation/NavUser.vue
+++ b/src/components/navigation/NavUser.vue
@@ -17,11 +17,11 @@ import {
 import { authManager } from "@/plugins/auth";
 import { eventbus } from "@/plugins/eventbus";
 import { store } from "@/plugins/store";
+import { LogOut, MoreVertical, Pencil, Settings } from "lucide-vue-next";
 import { computed } from "vue";
+import { useRouter } from "vue-router";
 
 const isEditMode = computed(() => store.homescreenEditMode);
-import { LogOut, MoreVertical, Pencil, Settings } from "lucide-vue-next";
-import { useRouter } from "vue-router";
 
 const router = useRouter();
 const { isMobile, setOpenMobile } = useSidebar();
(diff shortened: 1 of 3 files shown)


stvncode (commented):
That's nice thanks ! One more thing, could you add a condition for the dropdown content width to enhance it if it's in edit mode to have the "Leave Home screen edit mode" in one line pls ? It would need something like: editMode ? min-w-64 : min-w-56

src/components/navigation/NavUser.vue
```diff
@@ -17,6 +17,9 @@ import {
 import { authManager } from "@/plugins/auth";
 import { eventbus } from "@/plugins/eventbus";
 import { store } from "@/plugins/store";
+import { computed } from "vue";
+
+const isEditMode = computed(() => store.homescreenEditMode);
```
stvncode, line 24: Can you put this const below the last import please ?

dmoo500:
Added the conditional width as suggested — `min-w-64` when in edit mode, `min-w-56` otherwise. Also fixed the resulting Prettier formatting violations.

> **Note:** The `min-w-64` (256 px) was validated against the English string "Leave Home screen edit mode". Other languages may require a larger minimum width when they add translations for `homescreen_edit_disable`.


edit mode - leave

stvncode (commented):
Sorry, I forgot to add this comment...

src/components/navigation/NavUser.vue
dmoo500: done

stvncode (approved):
Thanks a lot ! 🙏

PR #1760, opened 2026-05-10
Show similar artists section on artist detail view

dmoo500:
## Summary

Adds a **Similar Artists** section to the artist detail view, shown below the tracks listing when at least one loaded provider supports `ProviderFeature.SIMILAR_ARTISTS`.

### Changes

- **`ArtistDetails.vue`** — new collapsed `ItemsListing` section (panel layout, no filter controls) that calls `getSimilarArtists` and is conditionally shown via `hasSimilarArtistsProvider`
- **`api/index.ts`** — new `getSimilarArtists(item_id, provider, limit)` method wrapping the `music/artists/similar_artists` API command
- **`api/interfaces.ts`** — added `SIMILAR_ARTISTS = "similar_artists"` to the `ProviderFeature` enum
- **`translations/en.json`** — added `"similar_artists": "Similar artists"` key

### Dependencies

The required `music/artists/similar_artists` API endpoint and `MusicProvider.get_similar_artists()` base method were merged into `dev` via server PR [#3811](https://github.com/music-assistant/server/pull/3811). No additional server PR is needed. The section is hidden when no provider declares `SIMILAR_ARTISTS` support, so this is safe to merge independently.

### Screenshot
artist view with similar artists

PR #1776, opened 2026-05-15
Show similar tracks section on TrackDetails view

dmoo500:
Adds a 'Similar tracks' section to the TrackDetails view, following the same pattern as the similar artists section in ArtistDetails (#1760).

## Changes
- Added `getSimilarTracks()` API method in `api/index.ts` (calls existing `music/tracks/similar_tracks` server endpoint)
- Added `"similar_tracks"` translation key to `en.json`
- Added Similar Tracks `ItemsListing` section to `TrackDetails.vue`, only shown when at least one provider supports `SIMILAR_TRACKS`
- The original track is filtered out from the results by comparing `provider_mappings`

## Screenshot
track view with similar tracks

### DIFF
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -57,6 +57,7 @@
     "on": "On",
     "other_versions": "Other versions",
     "party_mode": "Party",
+    "similar_tracks": "Similar tracks",
     "perform_action": "Perform action on {0} item(s)",
     "play": "Play",
     "play_album_from": "Play Album from here",
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -57,7 +57,6 @@
     "on": "On",
     "other_versions": "Other versions",
     "party_mode": "Party",
-    "similar_tracks": "Similar tracks",
     "perform_action": "Perform action on {0} item(s)",
     "play": "Play",
     "play_album_from": "Play Album from here",
@@ -752,6 +751,7 @@
         "missing_players_hint": "Missing players? {0}",
         "add_player_providers": "Add more player providers"
     },
+    "similar_tracks": "Similar tracks",
     "show_info": "Show info",
     "show_select_boxes": "Show selection boxes",
     "shuffle": "Shuffle",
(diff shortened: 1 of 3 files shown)


OzGav:
The artists are in a row whereas this is in a list. Isn’t that a bit inconsistent?

dmoo500:
The default view follows the standard MA pattern — tracks default to list, artists to panel — consistent with the rest of the app. The user can toggle via the header button.

Good spot though: without a unique itemtype, the view preference is shared with other listings of the same type (e.g. genre detail tracks). The existing codebase pattern for sub-listings in detail views is to use a distinct itemtype (like "artisttracks", "albumtracks" etc.)  I'll update both PRs to use itemtype="similartracks" / itemtype="similarartists" accordingly.

src/translations/en.json
```diff
@@ -57,6 +57,7 @@
     "on": "On",
     "other_versions": "Other versions",
     "party_mode": "Party",
+    "similar_tracks": "Similar tracks",
```
stvncode: Can you move this in alphabetical order please ?

src/views/TrackDetails.vue
```diff
@@ -122,4 +140,27 @@ const loadTrackAlbums = async function (params: LoadDataParams) {
...
   );
 };
+
+const hasSimilarTracksProvider = computed(() =>
+  Object.values(api.providers).some((p) =>
+    (p.supported_features as unknown as string[]).includes(
+      ProviderFeature.SIMILAR_TRACKS,
+    ),
+  ),
+);
```
stvncode, line 148:
```suggestion
const hasSimilarTracksProvider = computed(() =>
  Object.values(api.providers).some((p) =>
    p.supported_features .includes(ProviderFeature.SIMILAR_TRACKS),
  ),
);
```

stvncode (commented):
Same comments as similar artists pr 👍

src/translations/en.json
dmoo500: Done,

src/views/TrackDetails.vue
dmoo500: removed the unnecessary cast

stvncode (approved):
Thanks!

PR #1777, opened 2026-05-16
A11Y - Improve spoken search clear button label

bartbunting:
## Summary

Improves the accessible name for the search field clear button.

Previously, VoiceOver announced the clear button as something like “Clear type here to search”. This changes the field’s programmatic label so the generated clear button is announced as “Clear search”, while keeping the visible label as “Type to search”.

## Files changed

- `src/views/Search.vue`

## Testing

- `git diff --check`
- `yarn build`
- `yarn lint`
- `yarn test:run`

I also tested this with VoiceOver.

## AI usage

I used OpenAI Codex as an assistive tool while preparing this PR. I reviewed and tested the final change and am responsible for the submitted code.

### DIFF
--- a/src/views/Search.vue
+++ b/src/views/Search.vue
@@ -6,12 +6,14 @@
         v-model="store.globalSearchTerm"
         clearable
         prepend-inner-icon="mdi-magnify"
-        :label="$t('type_to_search')"
+        :label="$t('search')"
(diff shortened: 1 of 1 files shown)


stvncode (approved):
Thanks !

PR #1778, opened 2026-05-16
A11Y - Support Escape back navigation

bartbunting:
## Summary

Adds Escape-key handling for detail and settings pages so VoiceOver’s scrub/back gesture can navigate back from those views.  This is important for ease of navigation when using VoiceOver as without it the user needs to manualy locate the back button with their finger which is very inefficient.  

The handler avoids taking over Escape when a dialog, menu, overlay, player menu, or editable control is active, so existing modal/menu/input behavior should continue to take priority.

## Files changed

- `src/components/InfoHeader.vue`
- `src/views/settings/Settings.vue`

## Testing

- `git diff --check`
- `yarn build`
- `yarn lint`
- `yarn test:run`

I also tested this with VoiceOver on detail and settings pages.

## AI usage

I used OpenAI Codex as an assistive tool while preparing this PR. I reviewed and tested the final change and am responsible for the submitted code.

### DIFF
--- a/src/components/InfoHeader.vue
+++ b/src/components/InfoHeader.vue
@@ -441,7 +441,7 @@ import { eventbus } from "@/plugins/eventbus";
 import { store } from "@/plugins/store";
 import { IconHeart, IconHeartFilled } from "@tabler/icons-vue";
 import { ArrowLeft, Merge, Trash2 } from "lucide-vue-next";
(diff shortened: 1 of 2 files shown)


OzGav:
Marking this as draft until the comments and conflicts are resolved

PR #1779, opened 2026-05-17
Add Lucide icon picker with custom MA device icons for player settings

dmoo500:
## Summary

Replaces the MDI-only icon picker with a searchable [Lucide](https://lucide.dev) icon picker. Adds custom MA device icons and a unified `PlayerIcon` component used consistently throughout the UI.

---

## Icon picker

- Searchable dropdown across ~1 900 Lucide icons
- Default view shows a curated suggestion list (device icons first, then room/area/media icons)
- Ranked search: exact word match → prefix → word-prefix → substring; suggested icons are boosted within each rank
- Icon preview shown inline in the input field after selection
- Existing `mdi-*` values continue to work (backward compat via `PlayerIcon`)

Icon Picker

---

## Custom MA device icons

Four new device-specific icons ship in `src/components/ma-icons/`:

| Name | Device |
|---|---|
| `homepod-mini` | Apple HomePod mini |
| `sonos` | Sonos speaker |
| `mac` | Mac / MacBook |
| `apple-tv` | Apple TV |

Custom icons implement the same `size` prop interface as Lucide icons and use `currentColor` — they work identically everywhere a Lucide icon is used.

Adding a new icon in future: create `src/components/ma-icons/<name>.ts`, add it to `index.ts` registry. One file, ~10 lines.

---

## PlayerIcon component

New `PlayerIcon.vue` replaces all direct `<v-icon :icon="player.icon">` usages:

- Renders MDI icons (`mdi-*`) via `<v-icon>` for backward compat
- Renders Lucide / MA icons via `<component :is>`  
- Falls back to the Lucide `Speaker` icon when no icon is set

Player setting list icon

Playerview wo filledQueue

FullPlayerView wo filledQueue


OzGav:
Satisfies https://github.com/orgs/music-assistant/projects/2/views/1?filterQuery=mdi&pane=issue&itemId=62864526

stvncode:
So we are slowly switching everything to lucide icons. @MarvinSchenkel, do you think this make sense or it would be preferable to directly update it with an icon picker with lucide icons ?

MarvinSchenkel:
> So we are slowly switching everything to lucide icons. @MarvinSchenkel, do you think this make sense or it would be preferable to directly update it with an icon picker with lucide icons ?

If the effort is low the change the icons in the picker to lucide icons, I would suggest to go for lucide 👍

stvncode:
@MarvinSchenkel hmmm, i don't think it's low coz we have to do a new component (even if i think there is already examples online so that's ok) but the main problem is that everyone who's already using custom mdi icons will not have it anymore. So either we chose a default icon  if they were using a mdi icon and we let them modify with lucide for others or we stay with mdi for now (but we will modify it at some point)

marcelveldt:
We want to switch to lucide icons anyways and we also want to have some custom player specific icons, like an outline of a homepod mini etc. So maybe we should start working on that ?

Then backend can already set a good default based on the player type/model.

For instance, I would also love to see some icons for specific areas in your house, such as a kitchen etc.
A bit how the Hue app does this. It has nice icons for specific light fixtures but also area based. We would want the same for players/speakers and areas.

Idea: what if we support both for now so it remains backwards compatible ?
player icon has explicit user-set value of an mdi icon (the icon always starts with mdi-) --> use md icon
so the icon picker will show our selection of lucide icons (and maybe also md icons for now?) but can also still handle an existing mdi value.

alternative is that we do a one time migration in the backend to migrate well-known mdi icons to our curated list and reset the ones we cant match to default

dmoo500:
@marcelveldt This PR has been updated to address your feedback:

**What is implemented:**

- **Lucide-only picker** — MDI icons are no longer shown in the picker UI. Existing `mdi-*` values continue to render correctly via a new `PlayerIcon` component, so no migration is needed (your "support both" approach).
- **Custom MA device icons** — Four new icons in `src/components/ma-icons/`: `homepod-mini`, `sonos`, `mac`, `apple-tv`. The shapes were traced from the [MA Figma design file](https://www.figma.com/design/0EJf99xReC4253VWNzARWn/Player-drawer-and-selection-and-interactions?node-id=612-5677&t=ezaj6YBHn0GY7CQP-0).
- **`PlayerIcon` component** — Unified rendering throughout the UI. Falls back to Lucide `Speaker` when no icon is configured.
- **Suggestion list** — The picker shows MA device icons first, then curated Lucide icons for rooms/areas (sofa, bed-double, tree, car, door-open, sun, flower-2, utensils, …) — similar to the Hue approach you described.

src/components/IconPicker.vue
```diff
@@ -0,0 +1,112 @@
...
+        :is="getLucideIcon(modelValue)"
+        v-if="modelValue && getLucideIcon(modelValue)"
+        :size="20"
+        class="mr-2"
+      />
+    </template>
+
+    <!-- icon list items -->
+    <template #item="{ item, props: itemProps }">
+      <v-list-item v-bind="itemProps" :title="item.raw.value">
```
stvncode, line 37: Vuetify component

src/components/IconPicker.vue
```diff
@@ -0,0 +1,112 @@
+<template>
+  <v-autocomplete
```
stvncode, line 2: Vuetify component here

src/helpers/icon.ts
```diff
@@ -0,0 +1,89 @@
+import * as LucideIcons from "lucide-vue-next";
```
stvncode: Could we lazy load the icons so it doesn't land in the main bundle ?

src/components/PlayerIcon.vue
```diff
@@ -0,0 +1,22 @@
+<template>
```
stvncode, line 1: Extend PlayerIcon to handle the "grouped player" case (mdi-speaker-multiple) so PlayerCard.vue, PlayerFullscreen.vue (two spots) and PlayerTrackDetails.vue stop repeating the same v-if/v-else ladder

src/components/PlayerIcon.vue
```diff
@@ -0,0 +1,22 @@
...
+  <v-icon
+    v-if="isMdiIcon(icon)"
+    :icon="icon ?? undefined"
```
stvncode, line 10: is ?? undefined needed ?

src/components/IconPicker.vue
```diff
@@ -0,0 +1,112 @@
...
+  modelValue: string | null | undefined;
+  label?: string;
+  placeholder?: string;
+  disabled?: boolean;
+}>();
+
+const emit = defineEmits<{
+  (e: "update:modelValue", value: string | null): void;
+}>();
+
```
stvncode, line 77:
```suggestion
const emit = defineEmits<{
  (e: "update:modelValue", value: string | null): void;
  (e: "click:clear"): void;
}>();

```

  So today, when the user clicks the ✕ on the icon field:
  1. v-autocomplete fires click:clear
  2. IconPicker catches it, emits update:modelValue: null
  3. Parent runs onUpdateValue(null) which substitutes default_value rather than clearing

stvncode:
For the vuetify component, maybe you have no choice for the v-autocomplete since it's for the settings elements that didn't get refactored yet. You can let it like this because using a shadcn autocomplete will defer from other settings

src/components/PlayerIcon.vue
dmoo500: Even though the v-if="isMdiIcon(icon)" guard makes null impossible at runtime, Vue's template type checker still sees icon as string | null | undefined in the binding — and Vuetify's :icon prop doesn't accept null (only string | undefined). So ?? undefined is required to satisfy the TypeScript compiler.

src/helpers/icon.ts
dmoo500: done

src/components/PlayerIcon.vue
dmoo500: done

src/components/IconPicker.vue
dmoo500: done

src/components/PlayerIcon.vue
```diff
@@ -0,0 +1,29 @@
...
+import { Speaker } from "lucide-vue-next";
+import { computed } from "vue";
+
+const props = defineProps<{
+  icon?: string | null;
+  size?: number | string;
+  grouped?: boolean;
+}>();
+
+const lucideIcon = computed(() => getLucideIcon(props.icon));
```
stvncode, line 32: getLucideIcon depends on a non-reactive module variable, so PlayerIcon's computed shows the Speaker fallback for Lucide-icon players on first paint and never self-corrects. Making the loaded module a shallowRef might resolve the issue

yarn.lock
```diff
@@ -1783,7 +1783,7 @@
 
 "@mdi/js@7.4.47":
   version "7.4.47"
-  resolved "https://registry.npmjs.org/@mdi/js/-/js-7.4.47.tgz"
+  resolved "https://registry.yarnpkg.com/@mdi/js/-/js-7.4.47.tgz#7d8a4edc9631bffeed80d1ec784f9beae559a76a"
```
stvncode: Why this was changed ?

src/components/PlayerIcon.vue
```diff
@@ -0,0 +1,29 @@
+<template>
```
stvncode, line 1: This comp should have inheritAttrs: false imo

stvncode (commented):
Regarding the dynamic import, the goal is not really met. So the clean move is to import it synchronously, which both removes the hazard and simplifies the code. I'll try to improve it later

yarn.lock
dmoo500: sorry - fixed

src/components/PlayerIcon.vue
dmoo500: Good catch. I removed the non-reactive dynamic path, so Lucide icons now resolve reliably on first render.

src/components/PlayerIcon.vue
dmoo500: Agreed. I set inheritAttrs: false and keep explicit $attrs forwarding in the render branches.

marcelveldt:
@stvncode, @dmoo500 what is the status of this PR ?
Would be awesome to get this in and see these custom nice icons for your players everywhere

dmoo500:
I hope I've implemented this correctly.

One more question/thought: Would it make sense to create a separate icon repository later?

stvncode:
Hey, sorry, i was busy elsewhere. Can i let you rebase the pr so i can do a last review pls ?

src/helpers/icon.ts
```diff
@@ -0,0 +1,92 @@
...
+import * as LucideIcons from "@lucide/vue";
+
+/** All custom MA icon names, shown first in search results. */
+export const MA_ICON_NAMES: readonly string[] = Object.keys(
+  MaIcons.registry,
+).sort();
+
+function kebabToPascal(name: string): string {
+  return name.replace(/(^|-)([a-z])/g, (_, __, c: string) => c.toUpperCase());
+}
```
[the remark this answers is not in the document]
dmoo500: Fixed

src/helpers/icon.ts
```diff
@@ -0,0 +1,92 @@
...
+
+function pascalToKebab(name: string): string {
+  return name.replace(/([a-z])([A-Z])/g, "$1-$2").toLowerCase();
+}
```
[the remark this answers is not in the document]
dmoo500: Fixed

src/helpers/icon.ts
```diff
@@ -0,0 +1,92 @@
...
+  name: string | null | undefined,
+): Component | undefined {
+  if (!name || name.startsWith("mdi-")) return undefined;
+  // Resolve alias first, then look up in registry
+  const canonical = MaIcons.aliases[name] ?? name;
+  if (MaIcons.registry[canonical]) return MaIcons.registry[canonical];
+  // Fall through to Lucide
+  const comp = lucideModule[kebabToPascal(canonical)];
+  return typeof comp === "function" ? (comp as Component) : undefined;
+}
```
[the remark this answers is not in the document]
dmoo500: Fixed

src/helpers/icon.ts
```diff
@@ -0,0 +1,92 @@
...
+      return (
+        typeof val === "function" &&
+        /^[A-Z]/.test(key) && // icon exports are PascalCase
+        !key.endsWith("Icon") && // skip *Icon suffix aliases
+        !key.startsWith("Lucide") // skip Lucide* prefix aliases
+      );
+    })
+    .map(pascalToKebab)
+    .sort();
+}
```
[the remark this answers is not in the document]
dmoo500: Fixed

src/components/IconPicker.vue
```diff
@@ -0,0 +1,120 @@
+<template>
+  <v-autocomplete
+    v-model:search="searchQuery"
+    :model-value="modelValue"
+    :items="filteredItems"
+    :label="label"
```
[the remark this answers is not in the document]
dmoo500: Won't fix (by design)

As noted by @stvncode in earlier review: "For the vuetify component, maybe you have no choice for the v-autocomplete since it's for the settings elements that didn't get refactored yet. You can let it like this because using a shadcn autocomplete will defer from other settings."

This component is used exclusively in player settings (ConfigEntryField.vue) which still uses Vuetify. Mixing shadcn-vue here would create UI inconsistency. Once player settings are refactored to shadcn-vue, this picker can be migrated as well.

src/helpers/icon.ts
```diff
@@ -0,0 +1,92 @@
...
+
+type LucideModule = Record<string, unknown>;
+const lucideModule = LucideIcons as unknown as LucideModule;
+
+/** Returns the Vue component for a Lucide or MA kebab-case icon name, or undefined for MDI names. */
+export function getLucideIcon(
+  name: string | null | undefined,
+): Component | undefined {
+  if (!name || name.startsWith("mdi-")) return undefined;
```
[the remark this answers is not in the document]
dmoo500: Fixed / Added

src/components/IconPicker.vue
```diff
@@ -0,0 +1,120 @@
...
+    } else {
+      continue;
+    }
+
+    // Suggested icons sort before non-suggested within the same rank
+    if (suggestedSet.has(name)) rank -= 0.5;
+
+    ranked.push({ rank, name });
+    if (ranked.length >= MAX_ITEMS * 2) break;
+  }
```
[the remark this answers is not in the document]
dmoo500: Won't fix - the early break is a deliberate performance optimization. With 200 matches scanned before breaking, search quality remains high for typical queries. The theoretical edge case (better matches appearing after 200 partial matches) is unlikely given alphabetical distribution of icon names.

stvncode:
Few stuff that i saw while testing and that i'll fix with the rebase:

- Old icon just have a no data available, it's weird:
Screenshot 2026-06-25 at 16 10 45

- Some icons are missing: 
Screenshot 2026-06-25 at 16 11 09

stvncode (approved):
Really nice job, i just fixed a few small stuff and now it works really well 🙏

OzGav:
Noticed this. Here is a screenshot of an existing player and the associated dropdown

image

then I type in speaker and the option speaker-loud isn't shown?

image

PR #1780, opened 2026-05-17
Add sidebar shortcuts for playlists, artists, albums, tracks, radios, podcasts and audiobooks

dmoo500:
## Summary

Adds the ability to pin sidebar shortcuts for all supported media types: playlists, artists, albums, tracks, radios, podcasts, audiobooks and genres.

Related to https://github.com/orgs/music-assistant/projects/2/views/1?pane=issue&itemId=22640416

## Changes

- **`useShortcuts.ts`** (new composable): Central state for pinned sidebar shortcuts. Exports a `ShortcutItem` union type (`Playlist | Artist | Album | Track | Radio | Podcast | Audiobook | Genre`) and standalone helpers `isShortcutPinned`, `pinShortcutStandalone`, `unpinShortcutStandaloneItem`.

- **`NavShortcuts.vue`** (new component): Renders pinned shortcuts in the sidebar with thumbnail, name and media type subtitle. Includes a context menu button (matching ListView style).

- **`ItemContextMenu.vue`**: Added pin/unpin actions for all 8 supported media types (library items only).

- **`AppSidebar.vue`**: Registered `NavShortcuts`.

- **`en.json`**: Added `shortcut.add_to`, `shortcut.remove_from` translation keys.

- **`SidebarContent.vue`**: Scrollbar support for the collapsed sidebar (see below).

- **`userPreferences.ts`**: Extracted `setUserPreference` as a standalone helper.

## Storage

Pinned shortcuts are stored as a list of URIs in `user.preferences["sidebar.shortcuts"]` and persisted server-side via `api.updateUser`. Shortcuts are per-user and survive browser/app restarts.

## Collapsed sidebar scrollbar

When many shortcuts are pinned, the collapsed sidebar (icon mode) can overflow vertically and a scrollbar appears. A scrollbar inside the fixed 3rem-wide icon sidebar would normally clip the thumbnail images.

**`useSidebarScrollbarGutter.ts`** solves this with a `ResizeObserver`:
- When vertical overflow is detected in collapsed mode, `--sidebar-width-icon` is temporarily widened to `calc(3rem + 6px)` and `scrollbar-gutter: stable` is set — giving the scrollbar its own dedicated 6px gutter so it never overlaps content.
- When there is no overflow (or the sidebar expands), the variable is removed and the sidebar returns to its default width.
- The scrollbar is styled to be minimal: 4px wide on WebKit, `scrollbar-width: thin` on Firefox.

### Shortcuts in the Sidebar
Shortcuts in Sidebar (mouseover shows context menu)

### Shortcuts in the Sidebar (collapsed)
Shortcuts in Sidebar - collapsed2

context menu - playlist - pin
context menu - playlist - unpin


MarvinSchenkel:
First off, amazing stuff picking this up. This has been on my 'rainy day to do list' forever. I will have a play with it and hopefully @stvncode can assess the code part of this PR. Would be amazing if we can include this in the 2.9 release :-).

src/translations/en.json
```diff
@@ -93,6 +93,9 @@
     "export_playlist": "Export playlist",
     "remove_library": "Remove from library",
     "remove_playlist": "Remove from playlist",
+    "pin_to_sidebar": "Pin to sidebar",
```
MarvinSchenkel: What do we think of "Add to shortcuts"?

src/layouts/default/AddShortcutDialog.vue
```diff
@@ -0,0 +1,175 @@
...
+  type ShortcutItem,
+} from "@/composables/useShortcuts";
+import { eventbus } from "@/plugins/eventbus";
+import { $t } from "@/plugins/i18n";
+import { MediaType } from "@/plugins/api/interfaces";
+import { store } from "@/plugins/store";
+import { Pin, Search } from "lucide-vue-next";
+import { computed, onBeforeUnmount, onMounted, ref, watch } from "vue";
+
+const FILTER_TYPES = [
```
MarvinSchenkel: Should we add FOLDER here as well? That will allow people to add Recommendations to shortcuts as well. I think Genre is also a good candidate here

MarvinSchenkel:
Just had a play with it. Everything seems to work pretty well 👍 . What I would like to see functionally:
- Right click on desktop is not captured and does not open the context menu as with its library counterpart
- I would like the Play option at the top of the context menu as well, so people can quickly play the items in their shortcuts

dmoo500:
Thanks for the feedback @MarvinSchenkel! Here's what was addressed in the latest commit:

**Addressed feedback:**
- ✅ Right-click on desktop now opens the context menu (same as the library counterpart)
- ✅ Play option added at the top of the context menu
- ✅ Genre added as a supported shortcut type (shown in the add dialog and pinnable via context menu)
- ✅ Translation keys renamed and grouped under a `shortcut` namespace (`shortcut.add`, `shortcut.add_to`, `shortcut.remove_from`)

**Question about FOLDER:**
Regarding your suggestion to add `FOLDER` support — this would allow users to pin e.g. Apple Music recommendation folders as shortcuts. It's technically feasible but there are a few things to consider:

1. Folders navigate to `/browse?path=...` instead of a detail page, so the shortcut URL handling would need a special case
2. Folders are never `library` items, so the current pin/unpin guard (`provider === 'library'`) would need to be relaxed for folders
3. There's no way to list all available folders in the "Add shortcut" dialog — users could only add them via right-click while browsing

The main concern is **path stability**: if the `path` of a recommendation folder can change over time (similar to the rotating station IDs we fixed on the server side), shortcuts would silently break. Do you know whether recommendation folder paths are stable? If yes, we can add FOLDER support.

src/translations/en.json
Erioldoesdesign: mmm 'Shortcuts' is better than 'sidebar' 

Another common term that could be more familiar is 'bookmark' like when you bookmark a website. Book feels weird to use here though given these are albums and not books.

Erioldoesdesign:
Ah i see shortcut was changed already so disregard the bookmark comment i made! 

also i agree with Steven re. playing from the pins

Would it also be feasible to right click and unpin the shortcutted albums?

dmoo500:
Everything is implemented (see my previous comment), including right-clicking the shortcut to display the context menu. This also displays the "Remove from Shortcuts" menu item.

Only the FOLDER shortcut is not yet implemented.

MarvinSchenkel:
Hm I think I should be a bit more clear here. I don't think these are folders per se, but I noticed that for example the Apple Music stations do not show the 'Add to shortcut' options. The reason for my previous comment was that I think it would be great if we could pin all items that are shown on the 'Discover' page.

dmoo500:
Hi @MarvinSchenkel, it’s clear to me now what was meant by FOLDER. I added support for non-library shortcuts, and I also added composable tests.

MarvinSchenkel:
I did a demo of this during our weekly catch-up and we think it's a good idea to remove the entire '(+) add shortcut' button and the screen with search. Two reasons:
1. You are already interacting with the media item when you want to add something to the shortcut
2. Keeping it simple. Allowing users to do 1 thing in many ways is generally considered confusing.

What do you think of this?

dmoo500:
That makes sense. The dialog is an additional component that isn't really needed — users can already find any media item via the existing search / global search and pin it from there via the context menu. It also keeps the codebase smaller.

dmoo500:
Done — removed the Add Shortcut button and dialog. Shortcuts can now only be added via the context menu.

src/components/navigation/NavShortcuts.vue
```diff
@@ -0,0 +1,227 @@
...
+  [MediaType.ALBUM]: "albums",
+  [MediaType.TRACK]: "tracks",
+  [MediaType.RADIO]: "radios",
+  [MediaType.PODCAST]: "podcasts",
+  [MediaType.AUDIOBOOK]: "audiobooks",
+  [MediaType.GENRE]: "genres",
+};
+
+const MEDIA_TYPE_FALLBACK_ICON: Partial<Record<MediaType, string>> = {
+  [MediaType.PLAYLIST]: "mdi-playlist-music",
```
stvncode: Can you use lucide icons instead of mdi pls ?

src/components/navigation/NavShortcuts.vue
```diff
@@ -0,0 +1,227 @@
...
+    true,
+  );
+};
+</script>
+
+<template>
+  <template v-if="pinnedItems.length > 0 || !isCollapsed">
+    <div
+      :class="[
+        'my-1',
```
stvncode: Those 4 lines can just be one

src/components/navigation/NavShortcuts.vue
```diff
@@ -0,0 +1,227 @@
...
+                  'shortcut-thumb',
+                  isCollapsed ? 'shortcut-thumb--collapsed' : '',
+                ]"
+              />
+              <span v-if="!isCollapsed" class="shortcut-label">
+                <span class="shortcut-name">{{ item.name }}</span>
+                <span class="shortcut-type">{{ t(item.media_type) }}</span>
+              </span>
+            </SidebarMenuButton>
+            <v-btn
```
stvncode: Button instead of v-btn

src/components/navigation/NavShortcuts.vue
```diff
@@ -0,0 +1,227 @@
...
+              <img
+                v-if="getImageThumbForItem(item)"
+                :src="getImageThumbForItem(item, undefined, 64)"
+                :class="[
+                  'shortcut-thumb',
+                  isCollapsed ? 'shortcut-thumb--collapsed' : '',
+                ]"
+                :alt="item.name"
+              />
+              <v-icon
```
stvncode: Update since we use icon from lucide

src/composables/useShortcuts.ts
```diff
@@ -0,0 +1,320 @@
...
+
+    for (const uri of uris) {
+      try {
+        const item = await api.getItemByUri(uri);
+        if (item && SUPPORTED_TYPES.has(item.media_type)) {
+          results.push(item as ShortcutItem);
+          validUris.push(uri);
+        }
+      } catch {
+        // Item no longer exists — will be pruned below
```
stvncode: This is not good. Say a user has 5 pinned shortcuts. They open the app on a flaky train Wi-Fi. Two of the requests fail because of the connection. The code then writes the "valid" list (3 URIs) back to the server, permanently deleting the 2 that just had a network hiccup

src/composables/useShortcuts.ts
```diff
@@ -0,0 +1,320 @@
...
+      }
+    }
+
+    resolvedItems.value = results;
+
+    // Prune stale URIs from preferences
+    if (validUris.length !== uris.length) {
+      await setPreference(PREF_KEY, validUris);
```
stvncode: While that slow loop is running, the user can right-click an item and pin it via the context menu. That adds a 6th URI to preferences. Then the prune at the bottom runs and writes back the old list of 5, silently overwriting the new pin

src/composables/useShortcuts.ts
```diff
@@ -0,0 +1,320 @@
...
+
+async function _setPinnedUris(uris: string[]): Promise<void> {
+  if (!store.currentUser) return;
+  if (!store.currentUser.preferences) store.currentUser.preferences = {};
+  const updated = { ...store.currentUser.preferences, [PREF_KEY]: uris };
+  store.currentUser.preferences = updated;
+  await api.updateUser(store.currentUser.user_id, { preferences: updated });
+}
+
+export function isShortcutPinned(uri: string): boolean {
```
stvncode, line 157: You have two different function to see if a shortcut is pinned or not, this one and `isPinned`. Pick one comparison function and use it everywhere  (probably isShortcutPinned). Two functions named almost identically should never have different semantics

src/components/navigation/NavShortcuts.vue
```diff
@@ -0,0 +1,227 @@
...
+    event.clientX,
+    event.clientY,
+    true,
+    true,
+  );
+};
+</script>
+
+<template>
+  <template v-if="pinnedItems.length > 0 || !isCollapsed">
```
stvncode: I think it should && instead of || no ? Or maybe we can do pinnedItems.length > 0 and only show the image of the album/tracks (whatever tpye is it) when it's collapsed ?

src/components/navigation/AppSidebar.vue
```diff
@@ -68,6 +69,7 @@ onUnmounted(() => {
     </SidebarHeader>
     <SidebarContent>
       <NavMain :items="navItems" />
+      <NavShortcuts />
```
stvncode, line 72: Imagine you have 100 shortcuts, on every page load, you'll fetch a lot of data. Could we have a virtual scroll (@tanstack/vue-virtual)

src/components/ui/sidebar/SidebarContent.vue
```diff
@@ -1,23 +1,93 @@
...
+
+watch(state, updateScrollbarWidth);
+
+onMounted(() => {
+  updateScrollbarWidth();
+
+  resizeObserver = new ResizeObserver(updateScrollbarWidth);
+  if (el.value) resizeObserver.observe(el.value);
+
+  mutationObserver = new MutationObserver(updateScrollbarWidth);
```
stvncode: MutationObserver with { childList: true, subtree: true } fires on every DOM mutation anywhere inside the sidebar content. WOuld be better to wrap the callback so it runs at most once per animation frame

src/composables/useShortcuts.ts
```diff
@@ -0,0 +1,320 @@
...
+
+// ---------------------------------------------------------------------------
+// Standalone helpers — usable outside Vue component setup (e.g. context menus)
+// ---------------------------------------------------------------------------
+
+function _getPinnedUris(): string[] {
+  return (store.currentUser?.preferences?.[PREF_KEY] as string[]) ?? [];
+}
+
+async function _setPinnedUris(uris: string[]): Promise<void> {
```
stvncode: For this, either promote setPreference (and getPreference if needed) to standalone exports in userPreferences.ts and call them from useShortcuts.ts OR at minimum, make _setPinnedUris match setPreference's behaviour (deep-clone, try/catch, warn-on-no-user) so the two implementations don't drift.

src/composables/useShortcuts.ts
```diff
@@ -0,0 +1,320 @@
+import { computed, onMounted, onUnmounted, ref, watch } from "vue";
+import { api } from "@/plugins/api";
+import type {
+  Album,
+  Artist,
+  Audiobook,
+  EventMessage,
+  Genre,
+  ItemMapping,
+  MediaItemType,
```
stvncode: Unused

src/composables/useShortcuts.ts
```diff
@@ -0,0 +1,320 @@
...
+  });
+
+  onUnmounted(() => {
+    _unsubscribe?.();
+  });
+
+  return {
+    pinnedItems: computed(() => resolvedItems.value),
+    isPinned,
+    pinItem,
```
stvncode, line 367: Why export both if they are only use inside the components ?

src/components/ui/sidebar/SidebarContent.vue
```diff
@@ -1,23 +1,93 @@
...
+  if (!el.value) return;
+  const sidebarEl = el.value.closest<HTMLElement>("[data-slot=sidebar]");
+  if (!sidebarEl) return;
+
+  const overflows =
+    state.value === "collapsed" &&
+    el.value.scrollHeight > el.value.clientHeight;
+
+  if (overflows) {
+    sidebarEl.style.setProperty("--sidebar-width-icon", "calc(3rem + 6px)");
```
stvncode: Where does that magic number cam from ? Might be better to compute the sidebar-icon-width offset from the current value rather than hardcoding 3rem + 6px

src/components/navigation/NavShortcuts.vue
```diff
@@ -0,0 +1,227 @@
...
+                isCollapsed ? 'shortcut-button-collapsed' : 'shortcut-button',
+                isActive(getItemUrl(item))
+                  ? 'no-underline font-bold'
+                  : 'no-underline font-medium',
+              ]"
+              @click="handleClick"
+              @contextmenu.prevent="openContextMenu($event, item)"
+            >
+              <img
+                v-if="getImageThumbForItem(item)"
```
stvncode: Can you compute this please (getImageThumbForItem) so we don't do two calls like: 
`const thumb = computed(() => getImageThumbForItem(props.item, undefined, 64));
.
.
.
 `

src/composables/useShortcuts.ts
```diff
@@ -0,0 +1,320 @@
...
+      resolvedItems.value = [];
+      return;
+    }
+
+    const results: ShortcutItem[] = [];
+    const validUris: string[] = [];
+
+    for (const uri of uris) {
+      try {
+        const item = await api.getItemByUri(uri);
```
stvncode: Await inside a for loop is sequential. The next iteration doesn't start until the previous await resolves (it's really slow if you have a lot of shortcuts).
Promise.all should be better:
`const settled = await Promise.all(
      uris.map(async (uri) => {
        try {
          const item = await api.getItemByUri(uri);
          if (item && SUPPORTED_TYPES.has(item.media_type)) {
            return { uri, item: item as ShortcutItem };
          }
          return null;
        } catch {
          return null;
        }
      }),
    );`

src/composables/useShortcuts.ts
```diff
@@ -0,0 +1,320 @@
...
+    const toRemove = currentUris.filter((uri) => !newUris.includes(uri));
+    if (toRemove.length > 0) {
+      resolvedItems.value = resolvedItems.value.filter(
+        (p) => !toRemove.includes(p.uri),
+      );
+    }
+
+    // Fetch and add newly pinned items not yet resolved
+    const toAdd = newUris.filter((uri) => !currentUris.includes(uri));
+    for (const uri of toAdd) {
```
stvncode: Same here. Maybe we can use Promise.allSttled which never rejects: 
`  const settled = await Promise.allSettled(uris.map(uri => api.getItemByUri(uri)));
  settled.forEach((result, i) => {
    if (result.status === "fulfilled") { ... }
    else { /* prune uris[i] */ }
  });`

src/components/ui/sidebar/SidebarContent.vue
```diff
@@ -1,23 +1,93 @@
...
+import { useSidebar } from "./utils";
 
 const props = defineProps<{
   class?: HTMLAttributes["class"];
 }>();
+
+const el = ref<HTMLElement | null>(null);
+const { state } = useSidebar();
+
+let resizeObserver: ResizeObserver | null = null;
```
stvncode: Why having this logic here ? Would it not be better to be inside NavShorcuts ?

src/components/navigation/NavShortcuts.vue
dmoo500: Done — Lucide components (ListMusic, Mic2, Disc3, Music, Radio, Podcast, BookAudio, Tag, EllipsisVertical)

src/components/navigation/NavShortcuts.vue
dmoo500: done

src/components/navigation/NavShortcuts.vue
dmoo500: Done

src/components/navigation/NavShortcuts.vue
dmoo500: Done

src/components/navigation/AppSidebar.vue
dmoo500: Virtual scroll would only help with DOM rendering, not with the underlying API fetches — Promise.allSettled would still fire one request per URI regardless of what's visible.

To address the perceived loading delay I've added skeleton placeholders (SidebarMenuSkeleton) that appear immediately while the API calls are in flight, so the sidebar is never blank. I've also added MAX_SHORTCUTS = 50 as a hard cap — the "Add to shortcuts" menu item becomes disabled once the limit is reached, which guards against both the fetch and render concern.

src/composables/useShortcuts.ts
dmoo500: Done

src/components/navigation/NavShortcuts.vue
dmoo500: Done — changed to `v-if="pinnedItems.length > 0"`.

src/components/ui/sidebar/SidebarContent.vue
dmoo500: Done

src/components/navigation/NavShortcuts.vue
dmoo500: Done

src/composables/useShortcuts.ts
dmoo500: Done

src/composables/useShortcuts.ts
dmoo500: Done

src/components/ui/sidebar/SidebarContent.vue
dmoo500: Done

src/composables/useShortcuts.ts
dmoo500: Done — the prune step now uses `pinnedUris.value` (the reactive ref, read at prune time) instead of the stale `uris` snapshot captured at the start of the async fetch.

src/composables/useShortcuts.ts
dmoo500: Done — `isPinned` (returned by the composable) now also uses `isSameShortcutUri` so both paths have identical matching semantics. The two exported standalones serve different purposes: `isShortcutPinned` matches by URI string, `isShortcutPinnedItem` matches by item identity including provider_mappings — so both are needed.

src/components/ui/sidebar/SidebarContent.vue
dmoo500: Done — the MutationObserver callback now goes through an `requestAnimationFrame` debounce, so `updateScrollbarWidth` is called at most once per frame regardless of how many mutations fire.

src/composables/useShortcuts.ts
dmoo500: Done — extracted `setUserPreference` as a standalone export from `userPreferences.ts`. `_setPinnedUris` is removed; all standalone helpers now call `setUserPreference` directly.

src/composables/useShortcuts.ts
dmoo500: Done — removed.

src/composables/useShortcuts.ts
dmoo500: Done — `isShortcutPinned` (URI-based) is needed by the test suite and by context menus that only have a URI. `isShortcutPinnedItem` (item-based) is needed when provider_mappings must be checked. The redundant internal `isShortcutPinned` that duplicated `isPinned` was removed.

src/components/navigation/NavShortcuts.vue
```diff
@@ -0,0 +1,263 @@
...
+const { navEl } = useSidebarScrollbarGutter(pinnedItems);
+</script>
+
+<template>
+  <template v-if="pinnedItems.length > 0 || isLoading">
+    <div ref="navEl"></div>
+    <div
+      class="my-1 h-px shrink-0 bg-sidebar-border"
+      :class="isCollapsed ? 'mx-1' : 'mx-3'"
+    ></div>
```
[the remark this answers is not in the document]
dmoo500: Fixed.

src/composables/useShortcuts.ts
```diff
@@ -0,0 +1,370 @@
...
+
+  async function unpinItem(uri: string) {
+    // Remove immediately; watch won't re-remove (already absent)
+    resolvedItems.value = resolvedItems.value.filter(
+      (p) => !isSameShortcutUri(uri, p),
+    );
+    await setPreference(
+      PREF_KEY,
+      pinnedUris.value.filter((u) => u !== uri),
+    );
```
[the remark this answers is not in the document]
dmoo500: Fixed.

src/composables/useShortcuts.ts
```diff
@@ -0,0 +1,370 @@
...
+
+export type ShortcutItem =
+  | Playlist
+  | Artist
+  | Album
+  | Track
+  | Radio
+  | Podcast
+  | Audiobook
+  | Genre;
```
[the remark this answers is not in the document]
dmoo500: Intentional — genre items from the library view can be pinned too. PR description updated.

src/translations/en.json
```diff
@@ -93,6 +93,10 @@
     "export_playlist": "Export playlist",
     "remove_library": "Remove from library",
     "remove_playlist": "Remove from playlist",
+    "shortcut": {
+        "add_to": "Add to shortcuts",
+        "remove_from": "Remove from shortcuts"
+    },
```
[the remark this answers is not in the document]
dmoo500: Fixed in PR description.

marcelveldt (approved):
LGTM, thanks @dmoo500 !

PR #1782, opened 2026-05-19
Use server-derived color palette via `MediaItemPalette`

maximmaxim345:
Drop the `colorthief` dependency and the local palette extraction in `utils.ts`. PlayerCard, PartyDashboardView, PlayerOSD and the player fullscreen view now read `current_media.palette` from `PlayerMedia` and convert it to the existing `ImageColorPalette` shape through the new `paletteFromServer` helper.

Depends on:
- https://github.com/music-assistant/server/pull/3915

### DIFF
--- a/src/components/PlayerCard.vue
+++ b/src/components/PlayerCard.vue
@@ -233,20 +233,17 @@
 </template>
 
 <script setup lang="ts">
-import {
-  imgCoverDark,
-  imgCoverLight,
-} from "@/components/QualityDetailsBtn.vue";
 import { Button } from "@/components/ui/button";
 import VolumeControl from "@/components/VolumeControl.vue";
 import { useActiveSource } from "@/composables/activeSource";
 import { getPlayerMenuItems } from "@/helpers/player_menu_items";
 import {
-  getColorPalette,
+  EMPTY_COLOR_PALETTE,
   getMediaImageUrl,
   getPlayerName,
   ImageColorPalette,
   isBuiltinPlayer,
+  paletteFromServer,
 } from "@/helpers/utils";
 import api from "@/plugins/api";
 import {
@@ -259,7 +256,6 @@ import {
 import { getBreakpointValue } from "@/plugins/breakpoint";
 import { eventbus } from "@/plugins/eventbus";
 import { store } from "@/plugins/store";
-import vuetify from "@/plugins/vuetify";
 import { webPlayer } from "@/plugins/web_player";
 import { MoreVertical, Pause, Play, Power, Speaker } from "lucide-vue-next";
 import { computed, ref, toRef, watch } from "vue";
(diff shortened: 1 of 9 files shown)


src/components/PlayerCard.vue
```diff
@@ -318,29 +314,14 @@ const canPlayPause = computed(() => {
...
-});
-
-// utility feature to extract the dominant colors from the cover image
-// we use this color palette to colorize the playerbar/OSD
-const img = new Image();
-img.src = vuetify.theme.current.value.dark ? imgCoverDark : imgCoverLight;
-img.crossOrigin = "Anonymous";
-img.addEventListener("load", function () {
-  coverImageColorPalette.value = getColorPalette(img);
+  ...EMPTY_COLOR_PALETTE,
```
stvncode: You can just do this: `const coverImageColorPalette = ref<ImageColorPalette>(EMPTY_COLOR_PALETTE);`

src/plugins/api/interfaces.ts
```diff
@@ -895,13 +895,23 @@ export interface DeviceInfo {
   identifiers: Record<IdentifierType, string>;
 }
 
+export interface MediaItemPalette {
+  background_dark?: [number, number, number] | null;
```
stvncode, line 899: You only use on_dark and on_light, is the others properties needed ?

src/helpers/utils.ts
```diff
@@ -425,11 +425,30 @@ export const numberRange = function (start: number, end: number): number[] {
...
+  lightColor: "",
+  darkColor: "",
+};
+
+export function paletteFromServer(
+  palette: MediaItemPalette | null | undefined,
+): ImageColorPalette {
+  if (!palette) return { ...EMPTY_COLOR_PALETTE };
+  return {
+    lightColor: _rgbTupleToHex(palette.on_dark),
```
stvncode, line 446: Why do we have lightColor that depends on dark palette and same for the opposite ?

src/components/PlayerCard.vue
```diff
@@ -318,29 +314,14 @@ const canPlayPause = computed(() => {
...
 });
 
+// Use the server-derived palette from the now-playing media.
 watch(
```
stvncode: On top of that, instead of having a ref + watch, you can just do: `const coverImageColorPalette = computed<ImageColorPalette>(() =>
    paletteFromServer(compProps.player.current_media?.palette),
  );`

stvncode (commented):
Nice !! There are some unsued function now in utils that we should remove as long as their tests: getContrastingTextColor, getContrastRatio and lightenColor.
We should think about at some point to do the same for the party dashboard view that was doing the same logic as the playfullscreen view

stvncode (approved):
LGTM

PR #1783, opened 2026-05-19
Add audio-analysis page with coverage information to system-> settings

chrisuthe:
Adds a self-contained page to settings-> system-> Audio Analysis displaying per-provider analyzed/pending coverage bar, stale & version columns, plus a read-only background-scan status line.

image

image

Depends on https://github.com/music-assistant/server/pull/3851 + https://github.com/music-assistant/models/pull/231


OzGav:
Screenshot?

OzGav:
Nice. Not sure how easy it would be to add my AcoustID provider. We would have to know how many tracks are missing a musicbrainz recording ID first….

MarvinSchenkel:
I think it is a bit weird to show this by default at the top of the screen. Let's make a button next to 'add provider' that just opens a new breadcrumb settings > audio analysis > coverage (or something) that shows this overview?

chrisuthe:
@OzGav  and @MarvinSchenkel  Updated to new location per discussion. Updated to use new model cleanly

chrisuthe:
@OzGav  Your AA Provider should show up by default, no work on your part.

marcelveldt (approved):
Thanks, @chrisuthe !

OzGav:
Hi Chris, This looks great. I am just updating the docs. What does the stale number mean? How is the version number useful?

chrisuthe:
> Hi Chris, This looks great. I am just updating the docs. What does the stale number mean? How is the version number useful?

Stale == Tracks that have been analyzed by an older version of a provider. The provider has a "Version" field specifically to be able to update/adjust the analysis, if that happens (Say, we change how we do downbeat detection) - it should be "v2" of that analysis. That means any tracks with v1 analysis are "Stale" - if there is a better phrase or if you don't think that is useful in the front end feel free to get rid of it @OzGav , we use it under the covers to know we still need to do v2 analysis.

OzGav:
OK thanks. My general thoughts would be to show info to the user that they want to know? So maybe they need to know that unless it is already reflected in the bar above?

PR #1807, opened 2026-05-26
Format ArtistDetails similar artists API call

dmoo500:
Small formatting-only change in `ArtistDetails.vue` to keep the `api.getSimilarArtists` call on one line.

This isolates the recurring local format diff into a dedicated PR so it no longer keeps reappearing in unrelated branches.


stvncode:
Already done here: https://github.com/music-assistant/frontend/pull/1808/

PR #1809, opened 2026-05-26
Fix player selection on clearing queue

stvncode:


MarvinSchenkel:
I am not too sure about grabbing the latest selected speaker though.. Just thinking about accidental baby wake ups

stvncode:
@MarvinSchenkel, what you recommand ?

PR #1810, opened 2026-05-27
Derive library membership from in_library flag

OzGav:
I noticed that a single track saved from Spotify was inconsistently showing in-library status for the album and artist across the frontend. The artist's page (and search) treated it as already in the library and hid "Add to library", yet the artist never appeared under Library >> Artists.

I found that the frontend treated an item as "in your library" whenever it had a library database row, but that isn't right as the backend also creates library rows for an item's relatives (e.g. the artist and album behind a saved track). Those support rows were showing up as already-in-library.

This PR fixes this by determining in-library from the per-provider `in_library` flag, in one shared helper that every surface goes through. That is, the add/remove-to-library control (context menu and detail header) and the in-library badge on tiles, the detail header, and search results.


src/plugins/api/helpers.ts
```diff
@@ -74,6 +74,51 @@ export function requireServerVersion(minVersion: string): boolean {
...
+  if (!item) return false;
+  // favoriting forces an item into the library, so favorite implies membership
+  if ("favorite" in item && item.favorite === true) return true;
+  if ("provider_mappings" in item && Array.isArray(item.provider_mappings)) {
+    return item.provider_mappings.some((pm) => !!pm.in_library);
+  }
+  return false;
+};
+
+export const isItemFavorite = function (
```
stvncode: This is no used right ?

src/components/PanelviewItemCompact.vue
```diff
@@ -174,6 +171,16 @@ const compProps = withDefaults(defineProps<Props>(), {
 });
 
 const { t, te } = useI18n();
+
+// playlists always surface their source provider icon; everything else shows
+// the bookshelf when in library and the source provider icon otherwise.
+const providerIconDomain = computed(() => {
```
stvncode: Can we move this to a file to not have a duplicaiton here and in Listviewitem.vue pls ?

stvncode (commented):
One thing, the remove from playlist doesn't clear favorite, so favorited items won't update optimistically given favorite implies membership or maybe this id one in the be ?

src/plugins/api/helpers.ts
OzGav: Yeah sorry I thought I was going to use it but then forgot to remove it when I didn't

src/components/PanelviewItemCompact.vue
OzGav: Yes! I pulled the rule into a shared `getListItemProviderIconDomain()` in `plugins/api/helpers.ts`, so `ListviewItem.vue` and `PanelviewItemCompact.vue` both call that instead of each carrying their own copy.

OzGav:
> One thing, the remove from playlist doesn't clear favorite, so favorited items won't update optimistically given favorite implies membership or maybe this id one in the be ?

Favourites are a subset of `in_library` so it must be cleared when `in_library` is set to false so it makes sense to set it optimistically here as well. Done

stvncode (commented):
Lgtm. We can add unit test at some point for the helper function but will do that in another pr 👍

stvncode (approved):

PR #1811, opened 2026-05-27
fix(theme): fix dark-mode rendering

teancom:
Ran into an issue where the web UI wasn't properly rendering when in dark mode in Safari in macOS, but was able to reproduce in Chrome as well. Black text on a dark background, etc. No backend changes needed.

Three changes:

1. vuetify.ts: give Vuetify a private stylesheetId ("mass-vuetify-theme") instead of the well-known default. Anything injecting CSS into the page (browser extensions, userscripts, other libraries on the same page) can target the default "vuetify-theme-stylesheet" by name and overwrite the runtime CSS custom properties Vuetify generates. We discovered this via a Safari extension injecting Vuetify-2-era theme CSS under that id, but the vulnerability is generic across any browser with installable extensions.

2. style.css: hardcode shadcn sidebar tokens in :root and .dark instead of indirecting through rgb(var(--v-theme-panel)). The original was incorrect by CSS scope rules: --v-theme-panel is only declared inside .v-application; at :root the indirection resolves to whatever Vuetify put at the root level (the defaultTheme: "dark" value). That meant light-mode rendered the sidebar with dark colors. The hardcoded values mirror vuetify.ts's panel colors and are commented to that effect. Also adds color-scheme: light/dark at :root / .dark for correct native form-control and scrollbar defaults.

3. App.vue: switch setTheme from `theme.global.name.value = X` to `theme.change(X)` (4 call sites). The former goes through a Vuetify 3.x deprecation proxy that logs a console warning on assignment; theme.change is the recommended API. Functionally equivalent.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -91,22 +91,22 @@ const setTheme = function () {
 
   if (themePref == "dark") {
     // forced dark mode
-    theme.global.name.value = "dark";
+    theme.change("dark");
     themeValue = "dark";
   } else if (themePref == "light") {
     // forced light mode
-    theme.global.name.value = "light";
+    theme.change("light");
     themeValue = "light";
   } else if (
     window.matchMedia &&
     window.matchMedia("(prefers-color-scheme: dark)").matches
   ) {
     // dark mode is enabled in browser
-    theme.global.name.value = "dark";
+    theme.change("dark");
     themeValue = "dark";
   } else {
     // light mode is enabled in browser
-    theme.global.name.value = "light";
+    theme.change("light");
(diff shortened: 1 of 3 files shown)


teancom:
~~FYI, testing this on a local DEV server instance won't work until https://github.com/music-assistant/home-assistant-addon/pull/22   is merged and deployed.~~

https://github.com/music-assistant/home-assistant-addon/pull/23 was merged and this is now testable.

stvncode (approved):
Lgtm and sorry for the long response time 😃

PR #1693, opened 2026-04-10
Smart playlist UI

dmoo500:
## Smart Playlist UI

Frontend companion to the Smart Playlist plugin (music-assistant/server#3630).

### Features

- **Create Smart Playlist dialog**: name, mode (dynamic/fixed), track count, full rules form
- **Edit Smart Playlist dialog**: loads existing rules, saves changes
- **SmartPlaylistRulesForm** shared component used by both dialogs
- Rules: Genre (multi-select), Artist (search), Album (search), Favorites toggle, Min. popularity slider, Year range, Logic AND/OR
- **Seed track picker**: Popover search filtered to providers with `SIMILAR_TRACKS` feature; selected track shown as Badge; restores on re-open via `api.getItemByUri`
- **Seed artist picker**: Popover search filtered to providers with `SIMILAR_ARTISTS` feature; selected artist shown as Badge; mutually exclusive with seed track
- **Exclusion filters**: exclude specific artists, albums or track URIs from results; managed as badge lists with remove buttons
- **Dedup hours**: optional number input to skip recently played tracks (last N hours)
- Artist and Album pickers are disabled when a seed track or seed artist is active (incompatible on server side)
- Hint text shown under Artist/Album when seed overrides them
- `seed_track_name` / `seed_artist_name` stored alongside URI for human-readable playlist description
- `@keydown.stop` on all inputs inside dialogs (reka-ui Dialog intercepts keydown otherwise)
- Year inputs use uncontrolled DOM refs to avoid re-render interference

### Dependencies

This PR depends on music-assistant/server#3860 (`feature/similar-artists-controller`) and music-assistant/server#3630 (`feature/smart-playlist-plugin`). The seed artist picker uses the `SIMILAR_ARTISTS` provider feature flag introduced in #3860.

PR #1813, opened 2026-05-27
Fix Smart Playlist seed picker dropping all results when only plugin providers supply SIMILAR_TRACKS

chrisuthe:
## Summary

The seed-track / seed-artist pickers in the Smart Playlist dialog (added in #1693) filtered the global search results to tracks/artists whose `provider_mappings` included a provider declaring `SIMILAR_TRACKS` / `SIMILAR_ARTISTS`. Plugin-tier providers (e.g. `sonic_similarity`, `lastfm_recommendations`) never appear in a music track's `provider_mappings` — they're cross-provider engines, not music providers — so when a plugin was the only source of similar tracks the filter dropped every result and the picker appeared empty.

The server's [`mass.music.tracks.similar_tracks()`](https://github.com/music-assistant/server/blob/dev/music_assistant/controllers/media/tracks.py) already falls back to `METADATA` + `PLUGIN` providers that declare `SIMILAR_TRACKS` (and the same for similar_artists), so any track returned by `music/search` is a valid seed — the picker's per-result filter was encoding a stricter rule than the dispatcher actually enforces.

## Change

- `useSmartPlaylistSeedItems.ts`: drop the per-result `provider_mappings` filter on both the track and artist seed searches; keep the up-front "is any provider capable?" gate so the picker still returns `[]` when no provider supplies the feature at all.
- `selectSeedTrack` / `selectSeedArtist` are unchanged — they prefer a mapping owned by a SIMILAR_TRACKS-capable music provider (e.g. Spotify) and fall back via `buildItemUri` to a `library://` URI otherwise, which the server resolves correctly through the plugin fallback.

## Test plan

- [x] `useSmartPlaylistSeedItems.test.ts`: existing tests still pass (URI build, provider-id collection)
- [x] New regression test: search returns library/streaming tracks whose mappings only point at `library` / `filesystem_local` when a plugin provides `SIMILAR_TRACKS`
- [x] New gate test: search returns `[]` without calling `api.search` when no provider declares `SIMILAR_TRACKS`
- [x] Manual: with only `sonic_similarity` installed (no Spotify / Last.fm), open *Create Smart Playlist → Similar to track → Pick a track*, type a library track name → expect results
- [ ] Manual: with Spotify installed, same flow → expect Spotify + library results
- [ ] Manual: with no SIMILAR_TRACKS-capable provider at all, picker remains empty (existing behaviour)

### DIFF
--- a/src/composables/useSmartPlaylistSeedItems.ts
+++ b/src/composables/useSmartPlaylistSeedItems.ts
@@ -37,20 +37,16 @@ export function useSmartPlaylistSeedItems() {
       .map((p) => p.instance_id),
   );
 
+  // Don't filter results by provider_mappings: the server's similar_*
+  // dispatchers fall back to plugin-tier providers, which never appear there.
   setupDebouncedSearch({
     query: seedTrackSearch,
     results: seedTrackResults,
     isSearching: isSeedSearching,
     searchFn: async (q) => {
-      const providerIds = _similarTrackProviderIds.value;
-      if (providerIds.length === 0) return [];
-
+      if (_similarTrackProviderIds.value.length === 0) return [];
       const result = await api.search(q, [MediaType.TRACK], 20);
-      return result.tracks.filter((t) =>
-        t.provider_mappings?.some((m) =>
-          providerIds.includes(m.provider_instance),
-        ),
-      );
+      return result.tracks;
     },
   });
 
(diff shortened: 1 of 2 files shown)


chrisuthe:
Ping @dmoo500  - A plugin that exposed similar_tracks didn't exist when you wrote this, but now with sonic similarity you can use the similar to with your local library too.

stvncode (approved):
Lgtm ! 😃

PR #1819, opened 2026-05-28
Keep shortcuts in sync after delete and update events

dmoo500:
## Summary
- move shortcut delete cleanup to a global app-level subscription so stale pins are removed even when sidebar components are unmounted (mobile case)
- keep media item updates in `useShortcuts` so visible shortcut metadata (name/image) stays live

## Validation
- `yarn lint` (passes, existing repo warnings only)
- `yarn test:run` (all tests pass)

## Mobile view context
- On mobile, the sidebar/nav subtree can be unmounted while the app continues running.
- If a media item is deleted during that time, a listener that lives inside the sidebar composable can miss the MEDIA_ITEM_DELETED event.
- The stale shortcut URI then remains in user preferences and later triggers repeated "item not found" behavior when shortcuts are resolved.

## Why move delete listener to App.vue
- App.vue is always mounted for the active frontend session, so the delete cleanup listener is always active.
- This guarantees stale shortcut cleanup regardless of whether the sidebar is currently rendered.
- We keep MEDIA_ITEM_UPDATED handling in useShortcuts for local/live UI refresh (name/image) when shortcuts are visible.
- This split keeps correctness for deletion and preserves efficient UI updates for metadata changes.

### DIFF
--- a/src/composables/useShortcuts.ts
+++ b/src/composables/useShortcuts.ts
@@ -212,6 +212,31 @@ export async function pinShortcutStandalone(
   await setUserPreference(PREF_KEY, [...uris, maUri]);
 }
 
+let _globalShortcutsSyncInitialized = false;
+
+async function removePinnedUriIfPresent(uri: string): Promise<void> {
+  const currentUris = _getPinnedUris();
+  const nextUris = currentUris.filter((u) => !isSameShortcutUri(u, uri));
+  if (nextUris.length !== currentUris.length) {
+    await setUserPreference(PREF_KEY, nextUris);
+  }
+}
+
+export function initGlobalShortcutsSync(): void {
+  if (_globalShortcutsSyncInitialized) return;
+  _globalShortcutsSyncInitialized = true;
+
+  api.subscribe_multi(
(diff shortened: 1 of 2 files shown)


src/composables/useShortcuts.ts
```diff
@@ -212,6 +212,31 @@ export async function pinShortcutStandalone(
...
+  if (nextUris.length !== currentUris.length) {
+    await setUserPreference(PREF_KEY, nextUris);
+  }
+}
+
+export function initGlobalShortcutsSync(): void {
+  if (_globalShortcutsSyncInitialized) return;
+  _globalShortcutsSyncInitialized = true;
+
+  api.subscribe_multi(
```
stvncode: initGlobalShortcutsSync subscribes to MEDIA_ITEM_UPDATED but never handles it, we should either drop it or do something when we update ?
It only does something fro MEDIA_ITEM_DELETED

dmoo500: done

dmoo500:
@stvncode : I think the wrong label was set. It's more of a bug fix for MobileView – so that deleting a MediaItem within MobileView works without an error message appearing afterwards.

stvncode (approved):
Thanks 🙏

PR #1820, opened 2026-05-29
Fix smart playlist operator label after field switch

dmoo500:
## Summary
Fixes an operator label mismatch in the smart playlist rule editor when switching rule fields.

## Repro
- Change a rule field from `Year` to `Album` (or another non-year field).
- The operator label could remain `between` instead of switching back to `is`.
- The same could happen when switching back and forth between `Album` and `Year`.

## Fix
- Remount the operator select when the rule field changes by adding a field-based key.
- This ensures the displayed operator label always matches the currently selected field.

## Screenshot
Before and after the change from album to year

issue_switch from genre

issue_switch_to_year
You can see that the `is` is not set correctly – this PR fixes that.

### DIFF
--- a/src/components/smart_playlist/SmartPlaylistRuleRow.vue
+++ b/src/components/smart_playlist/SmartPlaylistRuleRow.vue
@@ -28,6 +28,7 @@
 
     <Select
       v-if="rule.field !== 'favorite'"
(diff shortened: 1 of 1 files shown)


stvncode (approved):
Thanks 🙏

PR #1822, opened 2026-05-29
Skip getSmartPlaylistRules call for non-smart playlists

dmoo500:

`PlaylistDetails` previously called `getSmartPlaylistRules` unconditionally for every playlist, relying on a try/catch to discard the result for non-smart playlists. This caused an unnecessary API round-trip on every playlist detail view.

Fix: Check `provider_mappings` for `provider_domain === "smart_playlist"` before making the call. Only smart playlists get the rules fetched; all others skip the call entirely.

Additionally, the previous code only fetched `smartRules` when `provider === "library"`, which meant the edit button and rule overview were missing when navigating to a smart playlist from the Discover/Recommendations page. This is fixed by removing that guard (the rules API accepts both library DB IDs and provider UUIDs).

- Companion server PR: https://github.com/music-assistant/server/pull/4025

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/src/views/PlaylistDetails.vue
+++ b/src/views/PlaylistDetails.vue
@@ -112,7 +112,10 @@ const listingRef = ref<InstanceType<typeof ItemsListing>>();
 
 const loadItemDetails = async function () {
   itemDetails.value = await api.getPlaylist(props.itemId, props.provider);
-  if (props.provider === "library") {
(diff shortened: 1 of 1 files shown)


MarvinSchenkel (approved):
Thanks @dmoo500

PR #1823, opened 2026-05-30
Refactor settings Edit config with shadcn comps

stvncode:
Screenshot 2026-05-30 at 10 21 58
Screenshot 2026-05-30 at 10 22 16

### DIFF
--- a/src/components/ui/alert/Alert.vue
+++ b/src/components/ui/alert/Alert.vue
@@ -1,0 +1,21 @@
+<script setup lang="ts">
+import type { HTMLAttributes } from "vue";
+import type { AlertVariants } from ".";
+import { cn } from "@/lib/utils";
+import { alertVariants } from ".";
+
+const props = defineProps<{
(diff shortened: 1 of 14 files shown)


marcelveldt:
Closing in favor of an incremental, piece-by-piece refactor on current `main`.

The direction here (shadcn + `@tanstack/vue-form` + `zod`) was right — `main` has since adopted those exact libraries, so we're building on the same foundation. But this branch predates the i18n server-migration (it re-translates server-provided strings on the frontend, which is now resolved server-side), uses the old icon package (`lucide-vue-next` vs `@lucide/vue`), and conflicts with several config changes that have landed since. Rather than rebase the whole 14-file change at once, we're redoing it in small, reviewable PRs.

Salvaged from here for the later steps: the `config-entry.ts` zod/validation helpers, the `ConfigEntryRow` split, and the shadcn `alert` component.

Thanks @stvncode — this was a useful blueprint. 🙏

First step just landed: #2008

PR #1848, opened 2026-06-02
Show Smart Playlist provider in playlists provider filter

dmoo500:
## Problem

The provider filter dropdown in the Playlists library view (`ItemsListing.vue`) was restricted to `ProviderType.MUSIC` providers only. The Smart Playlist provider is a `PluginProvider` (`type = PLUGIN`) and was therefore never shown in the filter, even though it creates playlists that appear in the same view.

## Solution

Special-case the Smart Playlist provider domain in the `musicProviders` computed property so it is included when `itemtype === "playlists"`.

## Why not use `ProviderFeature.LIBRARY_PLAYLISTS`?

The clean solution would be to add `ProviderFeature.LIBRARY_PLAYLISTS` to the Smart Playlist provider's `SUPPORTED_FEATURES` and remove the `ProviderType.MUSIC` restriction for that feature. However, `LIBRARY_PLAYLISTS` currently implies library sync behaviour on the server side — adding it to a Plugin provider would trigger unwanted sync logic.

A more generic solution should be considered in v2.10.0, since Smart Playlist may not be the only Plugin provider that manages playlists.

## Screenshot

provider filter smart_playlist

### DIFF
--- a/src/components/ItemsListing.vue
+++ b/src/components/ItemsListing.vue
@@ -250,6 +250,7 @@ import {
   type MediaItemType,
   type Track,
 } from "@/plugins/api/interfaces";
+import { SMART_PLAYLIST_PROVIDER_DOMAIN } from "@/components/smart_playlist/constants";
 import { eventbus } from "@/plugins/eventbus";
 import { store } from "@/plugins/store";
 import {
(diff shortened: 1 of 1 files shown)


stvncode (approved):
Thanks 🙏

PR #1851, opened 2026-06-03
Show catalog providers in library provider filter

OzGav:
The provider filter on library views only listed providers that declared the LIBRARY_<type> (sync) feature for the current mediatype. Catalog providers without a syncable library (e.g. Phish.in) were hidden even though their items can be favorited into the library and the backend filters library items by provider mapping regardless of sync support.

Broaden the per-itemtype feature map to also include the ARTIST_ALBUMS / ARTIST_TOPALBUMS / ARTIST_TRACKS / ARTIST_TOPTRACKS catalog features so these providers appear in the filter where they can contribute items, while keeping mediatype-specific providers (e.g. radio-only) out of unrelated filters. Also add the missing ARTIST_TRACKS / ARTIST_TOPALBUMS members to the ProviderFeature enum.

### DIFF
--- a/src/components/ItemsListing.vue
+++ b/src/components/ItemsListing.vue
@@ -819,20 +819,39 @@ const isPlayActionInProgress = computed(() => {
 });
 
 const musicProviders = computed(() => {
-  // Map itemtype to required ProviderFeature(s)
-  const featureMap: Record<string, ProviderFeature | ProviderFeature[]> = {
-    artists: ProviderFeature.LIBRARY_ARTISTS,
-    albums: ProviderFeature.LIBRARY_ALBUMS,
-    tracks: ProviderFeature.LIBRARY_TRACKS,
+  // Map itemtype to the ProviderFeatures that mark a provider as a possible
+  // source of that mediatype. This is intentionally broader than the
+  // LIBRARY_* (sync) features: a provider without a syncable library can still
+  // contribute items to the library by being browsed/searched and favorited
+  // (e.g. catalog providers exposing an artist's albums/tracks).
+  const featureMap: Record<string, ProviderFeature[]> = {
+    artists: [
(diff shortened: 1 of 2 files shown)


src/components/ItemsListing.vue
```diff
@@ -819,20 +819,39 @@ const isPlayActionInProgress = computed(() => {
...
-  const featureMap: Record<string, ProviderFeature | ProviderFeature[]> = {
-    artists: ProviderFeature.LIBRARY_ARTISTS,
-    albums: ProviderFeature.LIBRARY_ALBUMS,
-    tracks: ProviderFeature.LIBRARY_TRACKS,
+  // Map itemtype to the ProviderFeatures that mark a provider as a possible
+  // source of that mediatype. This is intentionally broader than the
+  // LIBRARY_* (sync) features: a provider without a syncable library can still
+  // contribute items to the library by being browsed/searched and favorited
+  // (e.g. catalog providers exposing an artist's albums/tracks).
+  const featureMap: Record<string, ProviderFeature[]> = {
```
stvncode, line 827: Are we not forgetting about smart playlist as well ?

OzGav: Smart playlists aren't affected by this change. The smart playlist view renders `DynamicPlaylistSample`, not `ItemsListing`, so this provider filter never runs there. 

Also the smart_playlist provider is type: "plugin", so it's already excluded by the existing provider.type !== ProviderType.MUSIC guard. 

Separately, smart playlists showing in Library Playlists but not being filterable by provider is a pre-existing trait of plugin-typed providers, independent of this PR

stvncode: Got it. Thanks for thee explanation 🙏

stvncode (approved):
LGTM ✅

PR #1858, opened 2026-06-04
Fix artist tiles using fanart instead of thumb in square views

OzGav:
`itemArtwork()` became fanart-first globally in "Fanart for top picks", which leaked into every square `EditorialMediaCard `tile (artist/album grids, search, genre, home rows) once those views adopted the discover-style card. Wide fanart is correct only for the Top Picks hero row.

Make fanart preference opt-in (default thumb-only) and have only the wide `EditorialHeroCard` request it.

### DIFF
--- a/src/components/discover/EditorialHeroCard.vue
+++ b/src/components/discover/EditorialHeroCard.vue
@@ -66,7 +66,9 @@ interface Props {
 const props = defineProps<Props>();
 const { t } = useI18n();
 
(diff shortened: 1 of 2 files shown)


stvncode:
Sorry @OzGav, I didn't see that you already fixed it.. I did it in [here](https://github.com/music-assistant/frontend/pull/1859) with the fix for the provider icon. Is that ok if I close this one ?

PR #1860, opened 2026-06-04
Always show lights and visualisers in the group list

OzGav:
A possible solution to the issue that visualisers are marked as HIDE IN UI when they are created. However, they need to appear in the group list so they can be joined to the actual music players. 

This idea is that this type of player (and light) will always be seen in the group list (assuming there is a compatible protocol to group via)

The PlayerType enum aligns with https://github.com/music-assistant/models/blob/eb1ac336d7b673622b084041ba9ea486f77dfb7d/music_assistant_models/enums.py#L403

### DIFF
--- a/src/components/VolumeControl.vue
+++ b/src/components/VolumeControl.vue
@@ -138,7 +138,11 @@
 
 <script setup lang="ts">
 import Button from "@/components/Button.vue";
-import { getPlayerName, truncateString } from "@/helpers/utils";
+import {
+  getPlayerName,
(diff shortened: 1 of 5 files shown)


stvncode (approved):
Can't really test it since i don't have hue lights but the code is perfect 🚀

OzGav:
What do you think @MarvinSchenkel

MarvinSchenkel:
Works great again, thanks @OzGav

PR #1861, opened 2026-06-04
Lower smart playlist dedup_hours max to 2160h (90 days)

MarvinSchenkel:

Lowers the maximum value of the smart playlist "do not repeat within X hours" (`dedup_hours`) field from **8760h to 2160h (90 days)** to match the server-side validation.

The server now bounds `dedup_hours` to the playlog retention window (90 days), since the dedup is computed from the playlog. Without this change the UI would accept values up to 8760h that the backend rejects.

Two spots in `SmartPlaylistDedupField.vue` are updated: the `NumberField` `:max` attribute and the clamp in `onChange()`.

## Companion PR

Server: music-assistant/server#4082


stvncode (approved):
LGTM ✅

PR #1862, opened 2026-06-04
Unify list row layout and refine play affordances

MarvinSchenkel:
## Summary

Brings the tracks, playlist and other list views in line with the new album-overview row style ([#1828](https://github.com/music-assistant/frontend/pull/1828)) and reworks how the play button is presented, with a deliberate goal of **never triggering playback by accident** (important for MA, where playback can target speakers in other rooms).

## Changes

- **Uniform row styling** — the album title font and right-aligned duration now apply to every list row; the track number stays exclusive to the album view.
- **Desktop (hover-capable):** a blue, discover-style play button reveals on hover — it replaces the album art on normal rows, and swaps in for the track number on album rows.
- **Touch devices (phones *and* tablets):** an explicit play button sits on the right, just before the ⋮ menu. The album art / track number stay clean on the left, so tapping them navigates (never plays). Gated on `(hover: none)` rather than viewport width, so tablets (e.g. iPad) keep an explicit button instead of an unreachable hover affordance.
- **Playlist tracks** now show the favourite heart on the right, for consistency with the other views.

## Testing

- `eslint` and `vue-tsc --noEmit` clean (pre-existing `require-default-prop` warnings only).
- Full unit suite passes (`vitest`, 176 tests).
- Manually checked desktop (hover reveal on album / tracks / playlist / artists), phone and iPad widths.


stvncode:
@MarvinSchenkel, we can maybe close this and put in in the discord thread for taking it again when tackling the subject for next release ?

MarvinSchenkel:
I put it indraft for now

stvncode (approved):
LGTM ✅

anaelseraphinux-netizen (approved):

PR #1863, opened 2026-06-04
Auto-Refresh the AA Status Page 

chrisuthe:
Automatically refresh the data for this page every 5 seconds (when there are pending candidates only) so people who are waiting to see numbers tick over as their library is scanned can see that it's working.  This does this and tests it.

I would prefer to base this on an event being emitted from the AA system but those don't exist- I plan to add them in a later version.  I did this based on what the Genre page does (polls the genre scanner status unconditionally ever 30s)


MarvinSchenkel:
As discussed on discord, I think we should leave this out for now. I think this should be event driven, but @stvncode will probably know this better than I do

PR #1864, opened 2026-06-04
Fix genre display in smart playlist rule picker

dmoo500:

The Smart Playlist genre rule picker was showing raw lowercase genre names from the database (e.g. `rock`, `synthwave`) instead of the properly translated/capitalised display names shown everywhere else in the UI (e.g. `Rock`, `Synthwave`).

**Root cause:** `useSmartPlaylistGenres.ts` was using `g.name` directly (the raw DB value), while the rest of the frontend uses `getGenreDisplayName(name, translation_key, t, te)` which resolves the i18n translation key and falls back to the raw name when no translation exists.

**Fix:** `genreOptions` and `genreName()` in `useSmartPlaylistGenres.ts` now call `getGenreDisplayName()`, matching the pattern used by `ListviewItem.vue` and other components. No server-side changes required.

The test for `useSmartPlaylistGenres` was updated with mocks for `vue-i18n` and `@/helpers/utils` to avoid pulling in the router/auth import chain in the test environment.

- related issue <link to issue>

### DIFF
--- a/src/composables/useSmartPlaylistGenres.ts
+++ b/src/composables/useSmartPlaylistGenres.ts
@@ -1,9 +1,12 @@
 import { computed, onMounted, ref } from "vue";
+import { useI18n } from "vue-i18n";
 import api from "@/plugins/api";
 import type { Genre } from "@/plugins/api/interfaces";
+import { getGenreDisplayName } from "@/helpers/utils";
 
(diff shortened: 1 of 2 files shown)


stvncode (approved):
LGTM ✅

PR #1865, opened 2026-06-05
Add explicit_only filter to smart playlist rules

dmoo500:

Adds UI for the `explicit` filter in smart playlist rules with clear, user-friendly options:

- **Allowed** (default/no filter) — shows all tracks regardless of explicit flag
- **Explicit only** — only show tracks marked as explicit
- **Not allowed** (default when adding filter) — exclude explicit tracks

**UI improvements:**
- Three-state dropdown with clear labels
- Default selection is "Not allowed" when user adds the filter
- Best-effort info message shown for both "From my library" and "Similar to..." modes
- Uses tri-state field: `null` (allowed), `true` (explicit only), `false` (not allowed)

**Related:**
- Companion server PR: music-assistant/server#4095  
- Addresses feedback from @marcelveldt (renamed from `explicit_only` to `explicit`)

#### Screenshots
discover_explicit_filter

library_explicit_filter


OzGav:
This could be contentious as while I am a fan Marcel has pointed out in the past that this will only work for some streaming providers that send the metadata. Thus you can seelct the option but still get explicit songs which obviously wont be what the user expects.

dmoo500:
We could also show the "best effort" warning banner for rule-based filters — see screenshot. This makes it clear to users that filter results depend on provider metadata availability, not just in seed mode.

show_warning_on_rules
Note: the view will look slightly different once PR #1866 (field icons) is merged.

OzGav:
I haven't following this closely enough. Is that screenshot for in-library only?

dmoo500:
This is a suggestion for the "From my library" section, showing how the "Best Effort" warning banner could look with the filters (and the explicit_only filter).

This banner is already displayed in the "Discover" filters. The explicit_only filter could also be implemented there.

OzGav:
OK that sounds OK. The point I was going to make is favourite and explicit are only available for in_library items. (Probably genre, album type and year as well but I would have to confirm)

dmoo500:
I've double-checked my current development:

#### "Discover" post filters (with existing "Best Effort" warning banner)
* Genre
* Album type
* Year
* Explicit (new)

#### "From my library" filters (without the "Best Effort" warning banner)
* Genre
* Album
* Favorite
* Artists
* Album type
* Year
* Explicit (new)

If it turns out that the "Explicit" option in "Discover" isn't working, I can remove it when I make this pull request "ready for review".

OzGav:
My point is you would still need the banner on in-library for explicit because many (most?) streaming providers don't send that flag and for local files there isn't a tag for it. The best you could probably do is a checkbox with something like "skip tracks with explicit lyrics when identified"

marcelveldt (approved):
LGTM
Thanks, @dmoo500 !

OzGav:
@dmoo500 please PR the docs and ensure a disclaimer is added about this being best effort.

dmoo500:
Please also approve https://github.com/music-assistant/server/pull/4095

PR #1830, opened 2026-05-31
Add link to background analysis Concurrency Setting

chrisuthe:
## What

Adds a card to **Settings → System → Audio Analysis**, positioned **above** the existing coverage frame (`<AudioAnalysisCoverage />`), that links to the **Streams** core configuration where the `background_scan_concurrency` value can be edited.

The setting stays in the generic core-config editor (it is *not* hidden on the backend), so this page simply points to it rather than duplicating the control.

## Changes

- **New component** `src/components/AudioAnalysisConcurrency.vue` — a `Card` (title + description) with a button that routes to `/settings/editcore/streams` via the existing `editcore/:domain` route, using the shadcn `Button as-child` + `RouterLink` pattern.
- **`src/views/settings/AudioAnalysis.vue`** — renders the card above `<AudioAnalysisCoverage />`.
- **`src/translations/en.json`** — adds the `audio_analysis_concurrency` block (title, description, and `open_settings` label).

## Testing

- `yarn lint` → 0 errors (3 pre-existing warnings in unrelated files).
- `vue-tsc --noEmit` → clean.
- Full test suite → 171/171 passing.

image

PR #1872, opened 2026-06-05
Fix erroneous underline on Audio Analysis concurrency link

chrisuthe:
Removes the erroneous underline on the "Open settings" button added in #1830 (Settings → System → Audio Analysis).

Adds the standard `no-underline` Tailwind utility to the `RouterLink` — the same approach used in `About.vue` and the navigation components (`NavMain.vue`, `NavShortcuts.vue`).

```vue
<RouterLink to="/settings/editcore/streams" class="no-underline">
```

image


stvncode (approved):
Thanks 🙏

PR #1876, opened 2026-06-07
Add built-in playlists for favorites and random tracks

OzGav:
Added strings for built-in playlists with labels and descriptions. This is not dependent on the backend change although a PR has been opened here https://github.com/music-assistant/server/pull/4122


stvncode (approved):
LGTM ✅

PR #1889, opened 2026-06-08
Add translation key for now playing badge

MarvinSchenkel:
## What

The "Now playing" badge in `NowPlayingBadge.vue` had its text and `aria-label` hardcoded in English. This replaces both with a new `now_playing` translation key so the badge is localizable.

## Changes

- `NowPlayingBadge.vue`: use `$t("now_playing")` for the visible text and the `aria-label` (same key for both).
- `translations/en.json`: add `now_playing` source key. Other locales will be populated via Lokalise.

## Testing

- `now_playing` badge renders on the now-playing list item (e.g. radio list) — confirm it shows "Now playing" in English and the localized string under a non-English locale.
- Lint + full unit test suite pass via pre-commit hooks.

### DIFF
--- a/src/components/NowPlayingBadge.vue
+++ b/src/components/NowPlayingBadge.vue
@@ -4,9 +4,9 @@
     v-if="showBadge"
     :style="badgeStyle"
     class="now-playing-badge"
-    aria-label="Now playing"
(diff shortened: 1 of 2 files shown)


stvncode (approved):
🚀

PR #1923, opened 2026-06-15
Use server-provided genre descriptions

marcelveldt:
Genre descriptions were hardcoded in the frontend (`genre_descriptions.*`, translated per-locale here), unlike every other media item whose description comes from the server. They are now authored and localized server-side and delivered on a Genre's `metadata.description`, so the frontend can simply read that.

- Read `metadata.description` for genres (drop the genre-specific branch in `InfoHeader.vue`)
- Remove the now-dead `getGenreDescription`/`genreKeyFromName` helpers
- Remove the `genre_descriptions.*` keys from all locale files

Depends on music-assistant/server#4227 and music-assistant/models#252.


src/components/InfoHeader.vue
```diff
@@ -626,14 +623,9 @@ const playButtonClick = function (forceMenu = false) {
 const rawDescription = computed(() => {
   if (!compProps.item) return "";
   if (compProps.item.metadata && compProps.item.metadata.description) {
+    // genre descriptions are now provided (and localized) by the server in
+    // metadata.description, the same as every other media item.
```
marcelveldt:
```suggestion
```

PR #1924, opened 2026-06-15
Use ExplicitIcon for explicit content field in smart playlist rules

dmoo500:
Updates the smart playlist rule field icon mapping to use the dedicated `ExplicitIcon` component for the explicit content field instead of the generic `Ban` icon.

This provides better visual consistency with the explicit content markers shown elsewhere in the UI.

**Changes:**
- Import `ExplicitIcon` component in `fieldIcon.ts`
- Map "explicit" field to `ExplicitIcon` in the field icon matcher


marcelveldt (approved):
Thanks, @dmoo500 !

PR #1934, opened 2026-06-19
Keep shuffle/repeat player buttons enabled with most queues

OzGav:
Satisfies: https://github.com/orgs/music-assistant/discussions/5691

The player-bar shuffle and repeat buttons greyed out whenever the queue was empty, inactive or mid play-action, while the now-playing menu kept the equivalent options available. Drop those conditions so the buttons sre the same as the menu, leaving only the queue-type guards (single dynamic playlist, infinite stream) that genuinely make the controls meaningless.

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerControlBtn/RepeatBtn.vue
+++ b/src/layouts/default/PlayerOSD/PlayerControlBtn/RepeatBtn.vue
@@ -3,13 +3,7 @@
   <Icon
     v-if="isVisible && playerQueue"
     v-bind="{ ...icon, ...$attrs }"
-    :disabled="
-      !playerQueue.active ||
-      playerQueue.items == 0 ||
-      isLoading ||
-      isSingleDynamicPlaylist ||
-      isInfiniteStream
-    "
+    :disabled="isSingleDynamicPlaylist || isInfiniteStream"
     :color="
       getValueFromSources(icon?.color, [
         [playerQueue.repeat_mode == RepeatMode.OFF, undefined],
@@ -65,12 +59,6 @@ const compProps = withDefaults(defineProps<Props>(), {
   isVisible: true,
   icon: undefined,
   size: 20,
-});
-
-const isLoading = computed(() => {
-  return (
-    compProps.playerQueue?.extra_attributes?.play_action_in_progress === true
-  );
 });
 
 const isSingleDynamicPlaylist = computed(() =>
--- a/src/layouts/default/PlayerOSD/PlayerControlBtn/ShuffleBtn.vue
+++ b/src/layouts/default/PlayerOSD/PlayerControlBtn/ShuffleBtn.vue
@@ -3,13 +3,7 @@
   <Icon
     v-if="isVisible && playerQueue"
     v-bind="{ ...icon, ...$attrs }"
-    :disabled="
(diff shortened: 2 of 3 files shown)


src/layouts/default/PlayerOSD/PlayerControlBtn/RepeatBtn.vue
```diff
@@ -3,13 +3,7 @@
...
     v-if="isVisible && playerQueue"
     v-bind="{ ...icon, ...$attrs }"
-    :disabled="
-      !playerQueue.active ||
-      playerQueue.items == 0 ||
-      isLoading ||
-      isSingleDynamicPlaylist ||
-      isInfiniteStream
-    "
+    :disabled="isSingleDynamicPlaylist || isInfiniteStream"
```
marcelveldt, line 6:
```suggestion
    :disabled=" isLoading || isSingleDynamicPlaylist || isInfiniteStream"
```

src/layouts/default/PlayerOSD/PlayerControlBtn/RepeatBtn.vue
```diff
@@ -67,12 +61,6 @@ const compProps = withDefaults(defineProps<Props>(), {
   size: 20,
 });
 
-const isLoading = computed(() => {
-  return (
-    compProps.playerQueue?.extra_attributes?.play_action_in_progress === true
-  );
-});
-
```
marcelveldt, line 75: dont drop this

src/layouts/default/PlayerOSD/PlayerControlBtn/ShuffleBtn.vue
```diff
@@ -3,13 +3,7 @@
...
     v-if="isVisible && playerQueue"
     v-bind="{ ...icon, ...$attrs }"
-    :disabled="
-      !playerQueue.active ||
-      playerQueue.items == 0 ||
-      isLoading ||
-      isSingleDynamicPlaylist ||
-      isInfiniteStream
-    "
+    :disabled="isSingleDynamicPlaylist || isInfiniteStream"
```
marcelveldt, line 6:
```suggestion
    :disabled=" isLoading || isSingleDynamicPlaylist || isInfiniteStream"
```

src/layouts/default/PlayerOSD/PlayerControlBtn/ShuffleBtn.vue
```diff
@@ -55,12 +49,6 @@ const compProps = withDefaults(defineProps<Props>(), {
   size: 20,
 });
 
-const isLoading = computed(() => {
-  return (
-    compProps.playerQueue?.extra_attributes?.play_action_in_progress === true
-  );
-});
-
```
marcelveldt, line 63: why did you drop this ?

OzGav: The isLoading condition, as I understand it, is to indicate a playback transition is happening and to stop a user from pressing another button that could interfere with the transition. I didn't think these two buttons have that concern and this was supported by the fact that the identical menu items were never gated in this way?

marcelveldt:
superseded by https://github.com/music-assistant/frontend/pull/1945

PR #1936, opened 2026-06-19
Show selected DSP preset in settings

zednotdead:
Related: https://github.com/orgs/music-assistant/discussions/5681

The DSP settings toolbar now shows the preset selected for the configured player using the server-persisted preset identity.

- Resolve the current name through the shared DSP preset registry and update it reactively
- Clear or safely fall back after manual changes, deletions, unresolved presets, player switches, and server events
- Keep the configured player preset separate from the effective playback output shown in stream details


src/views/settings/EditPlayerDsp.vue
```diff
@@ -454,8 +493,19 @@ watch(
       updatedFromServer = false;
       return;
     }
+
     if (oldVal === null) return; // We haven't changed anything yet
-    if (newVal) debouncedSave(newVal);
+    if (newVal) {
+      debouncedSave(newVal);
+
+      if (updatedFromLoadPreset) {
```
zednotdead: I'm not really happy with the way this section of the code is structured, it's not very clear. I welcome feedback on how it could be improved.

src/views/settings/EditPlayerDsp.vue
```diff
@@ -354,6 +385,8 @@ const loadPreset = async (preset: DSPConfigPreset) => {
   if (!preset || !preset.config) return;
 
   selectedStage.value = "input";
+  changeSelectedPreset(preset);
+  updatedFromLoadPreset = true;
```
zednotdead: This could be collapsed into
```suggestion
  changeSelectedPreset(preset, true);
```
with `updatedFromLoadPreset` being updated in that function, and having a toggle as an optional parameter.

I went with doing it like this, as it's a one-off functionality and fits with the rest of the code (like on [this line](https://github.com/music-assistant/frontend/pull/1936/changes#diff-71c92d99145d439aa2745e787fb2412d47bbef59b6edfbd9e317467979fa8fc8L270))

remon1496:
Thank you for your work on my suggestion @zednotdead! It's a really nice start. Would love the altered way of implementation. This will also (likely?) allow the mobile companion app to permanently checkmark the current active profile (currently, after selecting it, the checkmark also isn't persistent).

Schermafbeelding 2026-06-19 om 19 58 28 normaal

zednotdead:
> This will also (likely?) allow the mobile companion app to permanently checkmark the current active profile (currently, after selecting it, the checkmark also isn't persistent).

For that I think we'd need backend-side implementation, the mobile app is in a completely separate repo

remon1496:
> For that I think we'd need backend-side implementation, the mobile app is in a completely separate repo.

Indeed, I feel like it should be a backend implementation. :)

src/views/settings/EditPlayerDsp.vue
stvncode: It's ok the way you did it 👍

src/views/settings/EditPlayerDsp.vue
```diff
@@ -354,6 +385,8 @@ const loadPreset = async (preset: DSPConfigPreset) => {
...
   // Deep copy the preset config to avoid reference issues
   dsp.value = preset.config;
```
stvncode: Here you could do something like: `dsp.value = JSON.parse(JSON.stringify(preset.config));`

src/views/settings/EditPlayerDsp.vue
```diff
@@ -370,6 +403,7 @@ const savePreset = async () => {
     await api.saveDSPPreset(preset);
```
stvncode, line 491:
```suggestion
    const saved = await api.saveDSPPreset(preset)
```

src/views/settings/EditPlayerDsp.vue
```diff
@@ -370,6 +403,7 @@ const savePreset = async () => {
...
     newPresetName.value = "";
     showSavePresetDialog.value = false;
+    changeSelectedPreset(preset);
```
stvncode:
```suggestion
    changeSelectedPreset(saved);
```

marcelveldt:
Backend now has support to remember the active preset and I updated this PR to use that new backend logic.

marcelveldt (approved):
Thanks, @zednotdead !

PR #1938, opened 2026-06-19
Localize frontend-injected config entry categories

marcelveldt:
Config entry category headings come pre-translated from the server, but entries that the frontend injects on its own (the DSP shortcut, the multi-device DSP notes and the player options link) showed their raw category key (e.g. `dsp`, `options`) as the section heading. Only server-provided entries get their `category_label` resolved server-side, so the injected ones had no localized heading.

## Changes

- Resolve `category_label` for frontend-injected config entries from the `settings.category.*` translations in `EditPlayer.vue`, falling back to the raw category key when a translation is missing.
- Add the missing `settings.category.dsp` translation key.

### DIFF
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -150,6 +150,7 @@
             "web_player": "Web Player",
             "protocol_settings": "Output Protocol(s)",
             "options": "Player options",
+            "dsp": "DSP",
             "preferences": "Preferences",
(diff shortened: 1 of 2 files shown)


jazz05mine-bit (commented):
I need some assistance/guidance #502) 755-9013

PR #1940, opened 2026-06-20
Fix rectangle art album not well displayed in the now playing bottom (#66)

joperafe:
This PR addresses the issue reported in [_Rectangle art album or not well displayed in the now playing bottom
 #66_](https://github.com/music-assistant/backlog/issues/66)

### What changed
Minimal changes in PlayerTrackDetails.vue, the wrapper div around the album art `v-image` had no explicit dimensions, making the height grow unconstrained for non square images.
This fix adds width and height of 100% to that wrapper div, making it fill the 60/64px available.

### Before

Mobile / Desktop  
mobile before  
desktop before

---

### After

Mobile / Desktop  
mobile after  
desktop after

### Testing

Tested on Samsung S21 (Chrome)

Image placeholder from the big wide web: `https://images.unsplash.com/photo-1558572687-df2e1e142ea3?w=500&auto=format&fit=crop&q=60&ixlib=rb-4.1.0&ixid=M3wxMjA3fDB8MHxzZWFyY2h8MjB8fDE2JTNBOXxlbnwwfHwwfHx8MA%3D%3D`


src/layouts/default/PlayerOSD/PlayerTrackDetails.vue
```diff
@@ -21,6 +21,7 @@
             store.activePlayer?.powered != false &&
             store.activePlayer?.current_media?.image_url
           "
+          style="width: 100%; height: 100%"
```
kissu: Let's maybe use Tailwind utility classes for it instead? 🤗

joperafe: You're absolutely right! Just updated.
Thank you for the heads-up 🍻

stvncode:
@joperafe This is nice thanks. One thing that I saw after testing, we lost the border radius for the image, can you add it back pls ?

joperafe:
> @joperafe This is nice thanks. One thing that I saw after testing, we lost the border radius for the image, can you add it back pls ?

Sure thing! I'll have a look tomorrow and update.

Thank you!

joperafe:
> @joperafe This is nice thanks. One thing that I saw after testing, we lost the border radius for the image, can you add it back pls ?

After checking things I've come to the conclusion that the border radius is, technically, still there, but it's applied to the wrapper element generated by `v-img` rather than to the actual `` element.
The wrapper element of `v-img` receives the styles and expands to the full dimensions defined by `media-thumb` (60/64px). Since this particular image ratio doesn't fit the full area of `media-thumb`, the border-radius is not noticeable.

For the time being, my suggestion would be to add:
**`background-color: rgba(var(--v-theme-on-surface), 0.08);`**

Results in:
image

This is already used in `PlayerCard`, so we would keep consistency across the displayed media thumbnails.

One other, more involved option would be to introduce a shared component, `PlayerMediaThumb`, so that we have a single source of truth for artwork rendering.

 
 Let me know what you think of these findings. I'm happy to go with either approach, and I'm also open to other suggestions.

kissu:
Maybe we could work around the aspect ratio or have it cover/fill the background?
It is kinda fair that something very tall but narrow is not looking rounded. 😄

Moreover, not sure what looks the best:
- fitting the image better with its ratio, something [like this](https://codepen.io/kaicheng928/full/abBPerQ)
- having the whole thing with a nearby color as background (kinda what you are proposing and what we already have on the player with a simple gradient?)

CleanShot 2026-06-24 at 15 12 03@2x


Btw not sure if `v-theme-on-surface` is a Vuetify thing but we're trying to move away from it, hence maybe prefer something using tailwind/vanillaCSS over Vuetify's theme variables. 🤗
Not sure what is the overall vision when it comes down to design details on this one. 🤔

joperafe:
> Maybe we could work around the aspect ratio or have it cover/fill the background? It is kinda fair that something very tall but narrow is not looking rounded. 😄
> 
> Moreover, not sure what looks the best:
> 
> * fitting the image better
> * having the whole thing with a nearby color as background (kinda what you are proposing?)
> 
> Btw not sure if `v-theme-on-surface` is a Vuetify thing but we're trying to move away from it, hence maybe prefer something using tailwind/vanillaCSS over Vuetify's theme variables. 🤗 Not sure what is the overall vision when it comes down to design details on this one. 🤔

I feel you and I've checked both but I'm not so keen on the result, as we're either stretching or cutting the image.
Cover:
image
image

Fill:
image

If we go with any of the options above, I'd like to apply it consistently across the app (I saw `PlayerCard` using this background-color pattern, which is why I included the full-page screenshot for comparison).

Adding a background color (`bg-foreground/8` 😉) would probably be the lowest effort approach and still gets us to a reasonably good result without changing how artwork is currently rendered elsewhere.

kissu:
Honestly, any solution above is fine with me and is quite subjective I'd say.
Maintainers should have the last word haha. 😋

Not a huge difference between each other and I don't think that some cover would "fail" either way anyway. 🙌🏻
PS: also both of your images above are exactly the same to me, am I blind? 😂

joperafe:
> Honestly, any solution above is fine with me and is quite subjective I'd say. Maintainers should have the last word haha. 😋
> 
> Not a huge difference between each other and I don't think that some cover would "fail" either way anyway. 🙌🏻 PS: also both of your images above are exactly the same to me, am I blind? 😂

Totally agree!

I've updated the screenshots (not sure myself if those where the same image 😄) zoomed in since the differences are minimal.

Thank you for checking!

stvncode (approved):
Thanks @joperafe 🙏

PR #1950, opened 2026-06-22
Upgrade vitest and vitest coverage

stvncode:
Heads up: I tried bumping vitest to 4.x but it can't be installed under our current package manager. Since vitest@4.1.0, vite is declared in both dependencies and peerDependencies, and yarn 1.x's linker can't handle that. Yarn install dies with:

  `error Invariant Violation: could not find a copy of vite to link in node_modules/vitest/node_modules`

This is a known, unfixed limitation of yarn classic (confirmed upstream as a yarn 1.22.x bug, not a vitest one), and there's no resolutions/lockfile
  workaround:

  - https://github.com/vitest-dev/vitest/issues/9859
  - https://github.com/vitest-dev/vitest/issues/10523
  - https://github.com/vitest-dev/vitest/pull/9862 (attempted fix, didn't resolve yarn 1.x)

  So with yarn 1 we're stuck on vitest@3.x.

  The clean way out would be migrating to pnpm. It's something I've wanted to move us to for a while anyway: better install performance, stricter/more correct dependency resolution, and it unblocks the vitest 4 upgrade. The migration would touch the packageManager field, swap the lockfile, and update the CI workflows.

@marcelveldt, is that something we'd be open to?


stvncode:
Will do the switch to pnpm before doing that

PR #1952, opened 2026-06-22
Genre media type filtering

jozefKruszynski:
This PR adds a MediaType filter to the genre library view.
This uses existing backend features, but is the first part of a series of PRs aimed at adding 2 extra genre taxonomies to Music Assistant.


stvncode (approved):
LGTM ✅

PR #1958, opened 2026-06-23
Upgrade eslint + add oxlint

stvncode:
- Updating eslint from v8 to v10
- Implement oxlint that is on top and faster than eslint
- Fix the format that were formatting after the pr

Trade off is that we format now and lint and do a pnpm test before every commit but at least when you push everything is clean (still can be bypass by a force push tho)


eslint.config.js
```diff
@@ -0,0 +1,50 @@
...
+    rules: {
+      "vue/multi-word-component-names": "off",
+      "vue/no-v-text-v-html-on-component": "off",
+      "vue/no-v-html": "off",
+      "vue/attribute-hyphenation": ["error", "always", { ignore: ["onLoad"] }],
+      "no-unused-vars": "off",
+      "@typescript-eslint/no-non-null-assertion": "off",
+      "@typescript-eslint/no-unused-vars": "off",
+      "@typescript-eslint/ban-ts-comment": "off",
+      "@typescript-eslint/no-explicit-any": "off",
```
kissu: We still allow for some `any`? 😄
I thought that we were trying to move away from having too many of those in the codebase: https://github.com/music-assistant/frontend/pull/1535

.oxlintrc.json
```diff
@@ -0,0 +1,10 @@
+{
```
kissu, line 1: Unless I'm mistaken, it looks like oxlint is not offloading anything from ESLint as of right now because the plugins are not enabled.

We should probably add the following here
```js
"plugins": ["vue", "typescript", "oxc", "unicorn"],
```
that way, no gap in terms of linting. 🤗

Running `oxlint --vue-plugin` over `src/` brought those two
> src/views/PartyDashboardView.vue:638  vue/return-in-computed-property
> src/layouts/default/PlayerOSD/PlayerTimeline.vue:200  vue/no-side-effects-in-computed-properties

Hence the plugins key sounds like it might be mandatory for proper rules parity.

package.json
```diff
@@ -7,7 +7,12 @@
         "dev": "vite",
         "build": "vue-tsc --noEmit && vite build",
         "preview": "vite preview",
-        "lint": "eslint . --fix --ignore-path .gitignore",
+        "lint": "oxlint && eslint . --fix && pnpm format",
+        "lint:check": "oxlint && eslint . && pnpm format:check",
+        "lint:oxlint": "oxlint",
+        "lint:eslint": "eslint .",
+        "format": "prettier --write \"**/*.{js,cjs,ts,vue}\"",
+        "format:check": "prettier --check \"**/*.{js,cjs,ts,vue}\"",
```
kissu, line 15: I was curious if running the formatter on every file was relevant here or if we could maybe keep it only to the staged files maybe? Especially because we do have husky already. 👍🏻

```json
"lint-staged": {
  "*.{js,cjs,ts,vue}": [
    "oxlint",
    "eslint --fix",
    "prettier --write"
  ]
}
```

Moreover, we don't want to lint other files like JSON, CSS, yaml, md?

kissu: Moreover, kinda on the same note but I noticed that we do have a `/.pre-commit-config.yaml` file too?

**Isn't** that clashing/duplicate with Husky? Is it something we do use in a CI pipeline maybe?

kissu:
Having the ESlint flat config is indeed a huge addition that will make things easier down the road! 👍🏻

---

From what I understand of this PR, the main goal here is to have:
- oxlint run the most common basic JS/TS linting part, so that it speeds up the whole process as a whole 🏎️
- ESlint is still taking care of the rest (vue/TS-speciifc rules) to keep parity
- Prettier is then formatting all of it one last time while avoiding any conflicts so that it doesn't run **inside** of ESlint since it's slow + not recommended

I hope I got the initial idea of the PR right?

Given that assumption, I do have a few propositions from safest with no tradeoffs to more debatable ideas:
1. 🟢 we could use [dprint](https://dprint.dev/plugins/markup_fmt/) to replace Prettier entirely, safe and faster replacement with Vue support
2. 🟠 the linting situation is a bit of a mess indeed, not sure if we'll have huge gains overall adding 1 extra tool for small benefits (or are they meaningful for our use case? 🤗), hence just keeping the oldie ESlint
3. 🔴 we could potentially wipe the entire ESlint and live our life with `oxlint` + `oxfmt` (or `oxlint` + `dprint`), explanation below

This is the [current state of oxlint + oxfmt](https://oxc.rs/compatibility.html)

CleanShot 2026-06-24 at 09 09 31@2x

oxlint does not cover the following rules:
- 🟢 vue2, not really relevant for us
- 🟢 template validity rules:`valid-v-*`, `no-parsing-error` etc...but those are caught by the compiler during the build step anyway
- 🟢 style rules: `attributes-order`, `v-on-style` etc...not critical from a functional POV + could be suggest as a skill
- 🔴 actual bug guard rails: `no-mutating-props`, `require-v-for-key`, `no-ref-as-operand`, `no-unused-components`, `no-use-v-if-with-v-for` and `no-duplicate-attributes`

More details:
- currently, the rules that we're using in the project are `vue/require-default-prop` (3) and `vue/no-required-prop-with-default` (4) — both of which oxlint implements.
- `no-mutating-props` and `no-ref-as-operand` are already enforced in CI (Volar also catches up some extra ones) ✅
- hence the only ESlint-unique rules we're truly missing here are `require-v-for-key` and `no-use-v-if-with-v-for` (easily spottable in a review or in the browser console?)

---

Meanwhile, the benefits are also quite huge if we do switch fully to oxlint because:
- the speed gains could be very valuable (might be useful or not depending on where/how we run our CI. Is Github providing OHF free pipelines or are we paying/bottlenecked by it somehow?). It is after all powered by `ts-go` and is type aware as of today already 👍🏻
- future proof build that hopefully won't be redone in the next 6 weeks of this JS ecosystem 😅 (but VoidZero is kinda backing the whole thing hence might not be just dropped out of thin air)
- oxlint will still cover 100% of the rest aka `script` part for `.vue` SFCs + other files alongside
- usually, I would push back against switching away from ESlint but given the non-legacy codebase + small amount of rules we're using I think it's just fine (moreover we do not use any vue-specific AST tree + could roll our own thanks to `@vue/compiler-sfc` **if the need arises**)

CleanShot 2026-06-24 at 09 32 32@2x

---

Different topic entirely regarding
> Trade off is that we format now and lint and do a pnpm test before every commit but at least when you push everything is clean (still can be bypass by a force push tho)
What is the reason for running everything directly?
- is it for an agentic workflow so that it loops and keeps everything running properly?
- to ensure that a contributor has proper working code? Yet, won't it be caught by the PR pipeline?
- other reason I might not thought about? 😋

I feel like it might be a bit heavy especially as the project grows and we introduce more and more things into the loop. I personally know that quite a few people hate pre-commit hooks especially when they already did their fair share of automated + manual testing (and do not run the beefiest dev machines). 🕊️

PR #1988, opened 2026-06-29
Declutter the player queue list item layout

MarvinSchenkel:
The queue list rows were crowded: the main line combined artist + title, and the subtitle stacked duration and album on top of the album already being visible in the now-playing view. The reorder/menu controls also only appeared on hover.

This simplifies each row so the relevant info is easier to scan and the controls are always reachable.

- Main line shows just the track title (was artist + title)
- Subtitle now shows the artist (album dropped — it's already visible in the now-playing view)
- Duration moved to the right of the row, with figures that line up across rows
- Drag handle and context-menu button are always visible (no longer hover-only)
- Reordering is initiated only from the drag handle; it's shown disabled/grayed on rows that can't be moved (now playing, buffered, played)
- Duration is hidden on mobile to save space
- The item being dragged is highlighted with the accent color

## Before
image

## After
image

Reordering:
image

### DIFF
--- a/src/layouts/default/PlayerOSD/QueueListItem.vue
+++ b/src/layouts/default/PlayerOSD/QueueListItem.vue
@@ -86,11 +86,13 @@
       <!-- Fixed-width slot so the menu stays aligned whether or not a row has a
            grip (only up-next rows are reorderable). -->
       <div class="qitem__actions">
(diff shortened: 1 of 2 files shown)


stvncode (approved):
Good job! LGTM ✅

PR #1990, opened 2026-06-30
Fix browse sort/view settings not persisting per folder

OzGav:
`savedPrefs` captured `props.path` once at setup, but `BrowseView` reuses one `ItemsListing` instance across folder navigation while `props.path` changes. The read key drifted from the write key used by `changeSort`, so selecting a sort in a browse folder was immediately reset to the default by the `restoreSettings` watcher. Re-derive `savedPrefs` from the current `props.path` so each folder reads and writes its own settings.

### DIFF
--- a/src/components/ItemsListing.vue
+++ b/src/components/ItemsListing.vue
@@ -1437,10 +1437,13 @@ const loadData = async function (
   tempHide.value = false;
 };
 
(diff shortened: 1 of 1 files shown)


stvncode (approved):
Nice, thanks @OzGav 🙏

PR #1991, opened 2026-06-30
Fix main listing search stealing focus from other inputs

OzGav:
The global keydown listener in `ItemsListing` opened and focused its own search box on any keypress, even while typing in another input such as the player drawer's search field. Ignore keystrokes that originate from an editable element so they no longer leak into the main listing search.


Sawtaytoes (approved):

stvncode (approved):
Thanks 🙏

PR #1857, opened 2026-06-03
Migrate timeline slider to reka-ui, fix tap not skipping correctly

pierosavi:
Closes https://github.com/music-assistant/backlog/issues/67

Migrated Vuetify slider to Reka's https://reka-ui.com/docs/components/slider
Since Reka's doesn't support ticks for chapters, I extended the component to add the functionality as close as possible

[Screencast_20260603_222910.webm](https://github.com/user-attachments/assets/68778849-b265-442c-b4e7-315515a506eb)

I kept the UI as similar as possible to Vuetify's
Externalized the chapter computer, added tests 

[Screencast_20260603_223148.webm](https://github.com/user-attachments/assets/77a49d00-f590-4237-b52a-f2df0773599e)

PR #1994, opened 2026-06-30
Restore timeline progress bar spacing in the player bar

MarvinSchenkel:
## Problem

The reka-ui slider migration (#1857) set the timeline `SliderRoot` to `h-5` (20px), centering the track ~10px from the top. The old Vuetify slider used a 32px control (~16px above the track), so the compact player bar lost ~6px of breathing room and the progress bar now hugs the play controls. It also added a stray `mb-2` the old slider never had.

## Changes

- Restore the slider height to 32px (`h-8`) so the track re-centers with the original spacing
- Drop the `mb-2` that didn't exist before the migration

## Before
image

## After 
image

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerTimeline.vue
+++ b/src/layouts/default/PlayerOSD/PlayerTimeline.vue
@@ -4,7 +4,7 @@
       <SliderRoot
         v-model="wrappedCurTimeValue"
         data-slot="slider"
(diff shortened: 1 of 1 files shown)


stvncode (approved):
I wanted to work on this but couldn't find the time, thanks !!
Also, i think we had also a padding left and right before that we can add back later 👍

PR #1997, opened 2026-07-01
Restore progress bar fill thickness and time-label spacing

MarvinSchenkel:
## Problem

Follow-up to #1994. The reka-ui slider migration (#1857) left two spacing regressions in the player timeline compared to the old Vuetify slider:

- The played portion (`SliderRange`) renders 6px thick on a 4px track, so the fill bulges beyond the rest of the bar.
- The time labels sit only 4px below the track (was 8px). reka's slider is a 32px box vs Vuetify's 36px, so the existing `margin-top: -10px` pull-up now crowds the labels against the bar.

## Changes

- Set the range height to 4px (`h-1`) so the fill matches the track — uniform bar again.
- Reduce the time-label pull-up (`margin-top` `-10px` → `-6px`) to restore the original 8px gap.

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerTimeline.vue
+++ b/src/layouts/default/PlayerOSD/PlayerTimeline.vue
@@ -23,7 +23,7 @@
         >
           <SliderRange
             data-slot="slider-range"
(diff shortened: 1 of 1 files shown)


stvncode (approved):
🙏

PR #2000, opened 2026-07-01
Add support for authors and narrators

fmunkes:
Update 2026-08-05, is merged now: https://github.com/music-assistant/server/pull/5062

Update 2026-08-01, given the result of the discord poll, we use the audiobooks tabs now.

Author and Narrator tabs if available
2026-08-01-093147-screenshot-region

Collapses to icons if width is not sufficient. Scrollable horizontally if there is even less space.
2026-08-01-093156-screenshot-region

Standard artist view, just for audiobooks.
2026-07-27-221757-screenshot-region


OzGav:
@fmunkes I think the tabbed approach might be better. That is the approach I went with with my Classical idea

image

You could have the Audiobooks label on the left as normal and then Books / Authors / Narrators as tabs?

fmunkes:
@OzGav I integrated it in the audiobooks view. I tried the tabbed idea, but i really struggled with making this look decent, especially on mobile. That's me lacking design skills. I went with a dropdown selection now -  once you click on the title that is shown, but only if these methods are defined, so all the occurences of our ItemsListing remain untouched. Let me know what you think. The user will only have these options if there is a provider supporting them. And the last used view is  saved.
2026-07-10-175600-screenshot-region
2026-07-10-175624-screenshot-region

marcelveldt:
I think it's a bit weird that you need to press the icon/title to make some magical menu appear but I'm also lacking ideas.

Question number 1 I have is if this needs its own dedicated menu item in the (already crowded) navigation menu.
Can't we add dedicated menu buttons to the existing audiobooks view "browse by author" and "browse by narrator" ?

How many authors/narrators does somebody usually have in their library ?
If not many (say < 25) we can even make it a filter

fmunkes:
The magical menu is fair. I could add a down pointing "V" icon to the tile to indicate that there is more. The navigation menu (that is the left hand side, i.e. artists, titles, and so on, correct?) is untouched now - it is all part of the audiobooks view. Within the audiobooks view you can click on the title, and then change the audiobooks views content.

To me that is similar to adding a menu item, however, you are faster in changing that view. Switching between authors/ audiobooks I would assume to be something regular.

The how many authors question is difficult to answer for me - we definitely do have users with an audiobook collection exceeding multiple hundreds of items. No idea how many authors this means.

thomasloven:
Tabs at the bottom?
Ex: Home Assistant:
image

Otherwise just a Group By for Author with sub heading in the list would be a great step forward.
Also Ex Home Assistant:
image

fmunkes:
> Tabs at the bottom? Ex: Home Assistant:

We use tabs at the bottom already when on mobile for the navigation menu. So I think  that doesn't work well together.

> Otherwise just a Group By for Author with sub heading in the list would be a great step forward. Also Ex Home Assistant: 

That's a nice idea indeed - we could just use our regular item's listing and add one per author, i.e. replace 'Albums' here with author name, and list the books. Menu item to switch between regular, author and narrator view:
2026-07-12-103531-screenshot-region


@marcelveldt which way would you prefer me to go?

fmunkes:
@OzGav as you favored the tab idea - I combined your idea, and Marvin's comment on using the artist view. See pictures in PR description. Would be happy to get your opinion on this version - thanks!

OzGav:
Yeah I think that looks good. Personally I probably would have kept them on the Audiobooks page as I think they go together but either way it allows people to look through their collections in ways that make sense.

fmunkes:
> Yeah I think that looks good. Personally I probably would have kept them on the Audiobooks page as I think they go together but either way it allows people to look through their collections in ways that make sense.

Thanks. I'm fine either way, and changing that is simple. Maybe a good opportunity for a quick survey on discord?

OzGav:
You can put up a poll in the #polls channel and ping the feedback-squad. I can do it if you can't or would prefer me to?

fmunkes:
> You can put up a poll in the #polls channel and ping the feedback-squad. I can do it if you can't or would prefer me to?

I do not have the rights to do that, so I'd appreciate it if you could do that for me. Maybe you can either link or add the pictures. Thanks!

fmunkes:
~The requirement of signed commits now made this impossible now. I'll have to create a new PR, and add the changes manually again.~

We are back in business :D

marcelveldt:
@stvncode can you try to tweak the UI a bit for this please ?

stvncode:
Ok so I fixed a couple of things (codewise i mean) and regarding the ui/ux, i'm really not a fan of adding a new behavior on the header. For every page, we have the icon + name and I think it should stay like this for every content page. I tried a couple of things and one solution would be to insert it in the content part like this: 

Screenshot 2026-08-06 at 17 35 01

Wdyt @fmunkes ?

fmunkes:
> Wdyt @fmunkes ?

That looks perfect to me, thanks a lot! 😊

stvncode:
I just push the changes @fmunkes. I'll let you have a look to see if you like the new design and if everything's working has intented! If yes, we can merge this 😃

fmunkes:
@stvncode I tested it, and it is working fine!
Thank you for all the code fixes as well.

What I can't judge in terms of work - can we make it stick to the last selected tab, especially when you use "back" from e.g. an author details view. If not, that is absolutely fine too. Looks quite a bit more polished than what I had initially!

stvncode (approved):
I updated listing file so it preserve data and tab selection on go back 😃 
Thanks for the pr and sorry for the late review 🙏

PR #2012, opened 2026-07-04
Add subtle grey ring to unselected editorial card checkbox

OzGav:
Satisifes: https://github.com/music-assistant/backlog/issues/23#issue-3589082828

On a white background the current select ring is not visible. This adds a subtle grey ring around it so its visible on all backgrounds

BEFORE

image

image

image

AFTER

image

image

image


marcelveldt:
Fine for now but not a real improvement. Longer term we should come up with a better design for this, such as having the checkboxes at a dedicated place or dimming the covers to make the checkboxes more stand out.

marcelveldt (approved):

PR #2014, opened 2026-07-04
Fix DSP parametric EQ layout on narrow and rotated displays

OzGav:
- The three action buttons (Show Multi-Channel Controls / Import APO-REW Preset / Export APO Preset) were staggered at different widths when they wrapped, and their text spilled outside the buttons on narrow screens. I gave them all a uniform width so they line up neatly however they wrap, let long labels wrap instead of overflow, and tuned the width, font size, and letter-spacing so all three sit on one row with the "Show Multi-Channel Controls" label on a single line, while still stacking cleanly on phones.
- iPad rotation: when you rotated to landscape, the main content stayed narrow while the header used the full width. That was Vuetify's container capping its width at wider screen sizes. I made the DSP containers full-width so the view now uses the whole screen.
- The frequency graph stopped short of the right edge, not lining up with the buttons above it. I trimmed the graph's right-side padding (keeping the left padding for the dB labels) so the plot now reaches the same right edge as the button row.
- Satisifed https://github.com/orgs/music-assistant/projects/2/views/1?pane=issue&itemId=101956147 by changing the label and adding a X when there is no individual channel band added.

BEFORE:

image

image

image

AFTER:

image

image

image

image

image

image

### DIFF
--- a/src/components/dsp/DSPParametricEQ.vue
+++ b/src/components/dsp/DSPParametricEQ.vue
@@ -14,17 +14,30 @@
         <v-icon start>mdi-speaker-multiple</v-icon>
         {{ $t("settings.dsp.parametric_eq.show_multichannel_controls") }}
       </v-btn>
-      <v-select
-        v-else
-        v-model="editedChannel"
-        class="pa-2"
-        :items="channelTypes"
-        :label="$t('settings.dsp.parametric_eq.edited_channel')"
-        variant="outlined"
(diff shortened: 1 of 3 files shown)


marcelveldt:
For now this is a short-term fix but longer term this whole view needs to be converted to shadcn.

marcelveldt (approved):
Thanks, @OzGav !

PR #2028, opened 2026-07-06
Fix audio pipeline line spacing

OzGav:
Fixes this 

image

### DIFF
--- a/src/components/StreamDetailsChain.vue
+++ b/src/components/StreamDetailsChain.vue
@@ -103,7 +103,7 @@
       </div>
       <div class="w-full">
         <!-- Input header -->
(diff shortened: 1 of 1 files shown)


marcelveldt:
Also include a screenshot of the change endresult.
Note that you should test both with a syncgroup where all candidates have the same output path as well as multiple output paths as that uses a different route (with a dotted line semi-attached to the block above)

OzGav:
image

<br>
image

image


image

image


I note the line joining the second player no longer links up like it used to but that is not the result of this PR as can be seen by the before image.

stvncode (approved):
Nice thanks 🙏

PR #2030, opened 2026-07-06
Exit selection mode when the browse path changes

OzGav:
Selection mode survived breadcrumb navigation, so landing somewhere that hides the multi-select toggle (e.g. the browse root) left it stuck on with no way to disable it, and item clicks no longer navigated. The selection belongs to the folder it was made in, so clear it and leave selection mode whenever the path changes.


stvncode (approved):
Thanks 🙏

PR #2031, opened 2026-07-06
Fix genre merge cross taxonomy

jozefKruszynski:
Only show  same taxonomy genres in merge dialog
Disable merge option in management table when multiple taxonomy genres are selected

### DIFF
--- a/src/helpers/genreTaxonomy.ts
+++ b/src/helpers/genreTaxonomy.ts
@@ -1,16 +1,9 @@
 import type { MediaType } from "@/plugins/api/interfaces";
 
 // null/undefined content_type both mean the music/general taxonomy.
-export function normalizeGenreTaxonomy(
-  contentType?: MediaType | null,
-): MediaType | null {
-  return contentType ?? null;
-}
-
 export function genresShareTaxonomy(
   contentTypes: Array<MediaType | null | undefined>,
 ): boolean {
-  if (contentTypes.length === 0) return true;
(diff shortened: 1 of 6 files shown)


src/helpers/genreTaxonomy.ts
```diff
@@ -0,0 +1,16 @@
+import type { MediaType } from "@/plugins/api/interfaces";
+
+// null/undefined content_type both mean the music/general taxonomy.
+export function normalizeGenreTaxonomy(
+  contentType?: MediaType | null,
```
stvncode: I don't see the point of this function, it's just returning the value or null but value is already MediaType | null

src/helpers/genreTaxonomy.ts
```diff
@@ -0,0 +1,16 @@
...
+): MediaType | null {
+  return contentType ?? null;
+}
+
+export function genresShareTaxonomy(
+  contentTypes: Array<MediaType | null | undefined>,
+): boolean {
+  if (contentTypes.length === 0) return true;
+  const first = normalizeGenreTaxonomy(contentTypes[0]);
```
stvncode: You access an array with [0] when the array can be null or undefined. You need to check before if the array is not empty

jozefKruszynski:
Sorry, was a bit too hasty pulling the trigger last night before properly checking everything.

marcelveldt (approved):
Thanks, @jozefKruszynski !

PR #2033, opened 2026-07-07
Fix vertical alignment of rail dots in the stream details chain

MarvinSchenkel:
Since the shadcn migration of the quality details popover, the blue node dots on the connector rail no longer lined up with their rows.

The Input/Output header rows sized themselves from the label's line-height (21px) instead of the rail rhythm (17px, half a `--sd-row`), so every header pushed all rows below it out of sync with the rail — 4px drift after the Input header, 8px after Output.

- Add a `.streamdetails-header` class giving the section header rows a fixed `calc(var(--sd-row) / 2)` height
- Simplify `.streamdetails-separator` to a vertically centered dotted line
- Anchor the rail node dots with explicit `top`/`left` instead of relying on their static position

### DIFF
--- a/src/components/StreamDetailsChain.vue
+++ b/src/components/StreamDetailsChain.vue
@@ -103,7 +103,7 @@
       </div>
       <div class="w-full">
         <!-- Input header -->
-        <div class="flex">
+        <div class="streamdetails-header">
           <div
             class="streamdetails-separator"
             style="width: 40px; margin-right: 10px"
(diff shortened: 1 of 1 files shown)


marcelveldt:
Duplicate of https://github.com/music-assistant/frontend/pull/2028 ?

Also include a screenshot of the change endresult.
Note that you should test both with a syncgroup where all candidates have the same output path as well as multiple output paths as that uses a different route (with a dotted line semi-attached to the block above)

MarvinSchenkel:
Haha, Looks like Gav is faster. Let me close this one

PR #2034, opened 2026-07-07
Show track BPM and musical key on the track details page

MarvinSchenkel:
Adds display of the new track audio analysis metadata (BPM + musical key) on the track details page.

Depends on music-assistant/server#4626 (server exposes `Track.audio_metadata` via music-assistant-models 1.1.154).

- Add `AudioMetadata` interface and optional `audio_metadata` field to `Track` (only populated on full track requests, never on listings)
- Show BPM (rounded) and musical key rows in the track details header, hidden when absent
- Add English translation strings for the new labels


stvncode (approved):
Perfect!

PR #2035, opened 2026-07-07
Show a waveform progress bar in the fullscreen player

MarvinSchenkel:
The server now exposes precomputed track waveforms via the `audio_analysis/wave_form` API command (music-assistant/server#4626): 1800 normalized RMS energy bins covering the track duration. This renders that data as a SoundCloud-style waveform progress bar in the fullscreen now-playing view, giving a visual preview of the track's dynamics with the played portion highlighted.

- Add `api.getWaveForm()` wrapper for the `audio_analysis/wave_form` command
- New `WaveformTrack.vue`: two stacked canvases (dim + bright bars); the played portion is revealed via `clip-path: inset()`, so playback never triggers a canvas redraw
- Downsample the 1800 bins to one bar per 3px via max-pooling, re-rendered on resize at device pixel ratio
- Waveform height scales per breakpoint (32/40/48px) while the seek hit area stays ≥44px for touch
- `PlayerTimeline.vue` accepts an optional `waveform` prop that swaps the flat track for the waveform; seek/drag/keyboard interaction is unchanged (same reka-ui slider) and theming follows the existing artwork-derived slider color
- `PlayerFullscreen.vue` fetches the waveform (keyed by the streamdetails provider item id) on track change and falls back to the flat progress bar when no data is available; the footer/homescreen progress bar is untouched
- New per-user "Show waveform" setting (default on) under frontend display settings to opt back into the flat timeline

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
+++ b/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
@@ -1585,9 +1585,10 @@ watchEffect(() => {
   align-items: center;
   justify-content: center;
   max-width: 100%;
-  /* Side padding matches the 5% margins of the timeline/volume rows above
-     and below, so the outermost buttons align with the sliders. */
-  padding: 15px 5%;
+  padding: 15px;
+  /* Breathing room below the timeline, matching the visual gap between the
+     controls and the volume bar underneath. */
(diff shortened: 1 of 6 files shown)


stvncode (approved):
Very nice 😃

PR #2037, opened 2026-07-08
Add drag&drop reordering for sidebar shortcuts

dmoo500:
Adds drag & drop reordering for sidebar shortcuts, following the same pattern used in the player queue.

**What's changed:**

- New `useShortcutDragReorder` composable (based on `useQueueDragReorder`)
- Integrated drag handlers in `NavShortcuts.vue`
- Ghost element with pointer-based dragging (mouse, touch, pen)
- Auto-scroll when dragging near edges
- Only active when sidebar is expanded

The implementation uses pointer events and CSS transforms for smooth animations. Rows slide to open a gap at the drop position, and the reordering happens via the existing `reorderShortcutStandalone()` API.

### DIFF
--- a/src/components/navigation/NavShortcuts.vue
+++ b/src/components/navigation/NavShortcuts.vue
@@ -4,6 +4,7 @@ import {
   BookAudio,
   Disc3,
   EllipsisVertical,
+  GripVertical,
   ListMusic,
   Mic2,
   Music,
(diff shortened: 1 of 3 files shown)


OzGav:
Sorry tests are failing

src/components/navigation/NavShortcuts.vue
```diff
@@ -127,70 +145,127 @@ const { navEl } = useSidebarScrollbarGutter(pinnedItems);
...
-            class="mr-1.5"
-            @click.capture="swallowClickAfterHold"
-            @touchstart.passive="onTouchStart"
-          >
-            <SidebarMenuButton
-              :as="RouterLinkComponent"
-              :to="url"
-              :is-active="isActive(url)"
-              :tooltip="getDisplayName(item)"
+        <div ref="shortcutsScrollEl" style="position: relative">
```
joperafe: Maybe use Tailwind utility class instead of style?

src/components/navigation/NavShortcuts.vue
```diff
@@ -127,70 +145,127 @@ const { navEl } = useSidebarScrollbarGutter(pinnedItems);
...
+                <span v-if="!isCollapsed" class="shortcut-label">
+                  <span class="shortcut-name">{{ getDisplayName(item) }}</span>
+                  <span class="shortcut-type">{{ t(item.media_type) }}</span>
+                </span>
+              </SidebarMenuButton>
+              <div v-if="!isCollapsed" class="shortcut-actions">
+                <Button
+                  variant="ghost"
+                  size="icon"
+                  class="shortcut-drag-handle opacity-0 group-hover/menu-item:opacity-100 h-6 w-6"
```
joperafe, line 217: Just from reading this I'd expect touch devices to not work, since there's no `touch-action` on the handle, I can be wrong 😅
Have you tried it on a touch device, or on chrome devtools? If not, I can give it a try on my phone

src/components/navigation/NavShortcuts.vue
```diff
@@ -127,70 +145,127 @@ const { navEl } = useSidebarScrollbarGutter(pinnedItems);
...
+              </div>
+            </SidebarMenuItem>
+          </SidebarMenu>
+          <!-- Drag ghost -->
+          <div
+            v-if="isDragging && draggedItem"
+            class="shortcut-drag-ghost"
+            :style="{
+              top: `${ghostY}px`,
+              width: '100%',
```
joperafe: As mentioned above, at least `width` could be set as a Tailwind utility class. Not only this style property but maybe also the styles defined in `.shortcut-drag-ghost` could also come from utility classes?

src/components/navigation/NavShortcuts.vue
dmoo500: Has that changed in the fullscreen player in the meantime? I tried to replicate everything exactly as shown there, so I'm a bit surprised.

joperafe: I see what you mean. To be fair the original isn't a utility class either, but this PR already uses Tailwind utilities elsewhere in the template. IMO class="relative" feels like the more consistent choice for this new code

dmoo500: Done - switched to `class="relative"`

src/components/navigation/NavShortcuts.vue
dmoo500: Added `touch-action: none` to the drag handle (works on Chrome now, but the point to drag is more on the right side of the item)

src/components/navigation/NavShortcuts.vue
dmoo500: Converted ghost element to Tailwind utilities (position, layout, dimensions) - kept theme-specific styles (primary color, shadow) in CSS to match the queue drag pattern

marcelveldt:
Sorry, @dmoo500 I was already working on edit mode for the whole menu (including reordering) so Kind of rendered your PR obsolete. Hope you don't mind I close this PR here in favor of #2060 which is the global refactor. Thanks for the efforts anyways!

dmoo500:
No worries - the main thing is that the project is making progress.

PR #2039, opened 2026-07-09
Show diagnostics download button on server 2.9.6+

marcelveldt:
The diagnostics report feature is being backported to the stable 2.9.6 release, but the frontend only showed the "Download diagnostics" button on 2.10.0+ servers.

- Lower the version gate so the button also appears for 2.9.6+ stable users once the backport lands


marcelveldt:
Closing this one. After reconciling: diagnostics is being backported to stable (2.9.6) server-side, but the download button stays a 2.10 frontend feature — stable users can grab the report by browsing to /diagnostics directly. So we keep the gate at 2.10.0 (already on main from #2038).

PR #2042, opened 2026-07-10
Add Timeline view for MusicBrainz artist events

dmoo500:
## Description

Adds a Timeline shelf to the Discover/Home page that displays MusicBrainz artist events (birthdays, memorials, group founding/disbanding dates) falling within the configured scan window.

## Features

- Horizontal scrolling timeline using the standard `EditorialMediaCard` layout — same play button, hover behaviour and sizing as all other artist cards
- Event-specific Lucide icons: Cake (Birthday), Heart (In Memoriam), Users (Founded), UserMinus (Disbanded) shown as subtitle alongside the localized historical date (e.g. `Jan 8, 1947`)
- Date formatted via `Intl.DateTimeFormat` — respects the user's locale
- Events derived client-side from `metadata.life_span` and `metadata.artist_entity_type` — no extra server-side category folders needed
- Events filtered to the server's configured scan window (± days) to avoid showing out-of-window events from stale cache
- Toggle in edit mode to show/hide the timeline row

## Dependencies

- Server PR: music-assistant/server#4687 (provides flat TIMELINE folder with artists + metadata)
- Models PR: music-assistant/models#298 (`ArtistEntityType` + `LifeSpan` on `MediaItemMetadata`)

## Screenshots

musicbrainz_artists_events

## Technical Details

- `EditorialTimeline.vue` — shelf component; derives events from artist metadata
- `EditorialMediaCard.vue` — minor addition: optional `#subtitle` slot so consumers can override the default subtitle
- `interfaces.ts` — adds `ArtistEntityType`, `LifeSpan`, and both fields to `MediaItemMetadata`

## Testing

- Timeline displays artist events with correct icon and date
- Play button matches existing EditorialMediaCard styling
- Toggle works in edit mode
- Events outside the configured window are not shown


dmoo500:
Waiting for models PR music-assistant/models#299 to be merged.

The models PR adds a `RecommendationFolderType` enum that will simplify this implementation. Instead of checking folder URIs to detect MusicBrainz timeline folders, the frontend will check `folder.type === RecommendationFolderType.TIMELINE`.

dmoo500:
The Test / Lint failed, but apparently not because of my changes.

src/components/discover/EditorialTimeline.vue
```diff
@@ -0,0 +1,289 @@
...
+    if (dist < bestDist) {
+      best = c;
+      bestDist = dist;
+    }
+  }
+  return Math.round((best.getTime() - todayMs) / 86_400_000);
+}
+
+function dateLabel(offset: number, mmdd: string): string {
+  if (offset === 0) return "Today";
```
stvncode: This is hardcoded, you should use i18n

src/components/discover/EditorialTimeline.vue
```diff
@@ -0,0 +1,289 @@
...
+const MAX_ART = 280;
+
+const track = ref<HTMLElement | null>(null);
+const tileArt = ref<number | null>(null);
+
+const gap = computed(() => props.gap ?? 14);
+
+const MAX_DAYS = computed(() => props.maxDays ?? 15);
+
+const minDaysLabel = computed(() => {
```
stvncode: minDaysLabel, maxDaysLabel and fullDate() are dead code

src/components/discover/EditorialTimeline.vue
```diff
@@ -0,0 +1,289 @@
...
+const PHONE_CARD_PAD = 8;
+const MIN_ART = 120;
+const MAX_ART = 280;
+
+const track = ref<HTMLElement | null>(null);
+const tileArt = ref<number | null>(null);
+
+const gap = computed(() => props.gap ?? 14);
+
+const MAX_DAYS = computed(() => props.maxDays ?? 15);
```
stvncode, line 54: The 15-day window is hardcoded client-side; nothing passes :max-days, so the server's configured scan window is ignored contrary to the description

src/components/discover/EditorialTimeline.vue
```diff
@@ -0,0 +1,289 @@
...
+  const days = MAX_DAYS.value;
+  return days === 1 ? "Yesterday" : `${days} days ago`;
+});
+
+const maxDaysLabel = computed(() => {
+  const days = MAX_DAYS.value;
+  return days === 1 ? "Tomorrow" : `in ${days} days`;
+});
+
+const EVENT_ICONS: Record<string, unknown> = {
```
stvncode: You can probably replace unknown by either Component from vue or type LucideIcon from @lucide/vue

src/components/discover/EditorialTimeline.vue
```diff
@@ -0,0 +1,289 @@
...
+  title: string;
+  items: MediaItemTypeOrItemMapping[];
+  provider?: string;
+  dimmed?: boolean;
+  tilesPerView?: number;
+  maxDays?: number;
+  gap?: number;
+}>();
+
+const CARD_PAD = 16;
```
stvncode: This is used with same value in 2 different files, EditorialShelf and EditorialTimeline. CARD_PAD/PHONE_GAP/PHONE_CARD_PAD/MIN_ART/MAX_ART and the whole updateTileArt function are identical copies of EditorialShelf.vue:103-107 and :149-160, along with the --ed-gutter/--ed-card-pad vars, the scrollbar-hiding CSS and the 600px media query.

The copy also drops three things the shelf does deliberately:

- Resize handling. The shelf runs a resize listener + ResizeObserver on the track (:203-215); here updateTileArt only fires on mount and on tilesPerView change, so card art stays at its mount-time size when the window resizes.
- touch-action: pan-x pan-y. The shelf sets this with the comment "pan-y too: a vertical swipe starting on a tile must scroll the page". Nothing sets it here, so vertical scrolling over this row on mobile is likely to stick.
- overscroll-behavior-x: contain (plus scroll-snap and the nav chevrons).

EditorialShelf already takes title, provider, dimmed and tilesPerView — exactly what HomeWidgetRows passes to this component. So it can host the row directly. The only genuinely new logic here is the timelineItems derivation; wrapping that in <EditorialShelf> with EditorialMediaCard children would drop ~120 lines and mean future shelf fixes apply to this row too.

dmoo500: Fixed in 1908ac3e - added resize handling (ResizeObserver + window listener) and touch-action CSS. The duplication of constants and updateTileArt remains since EditorialTimeline and EditorialShelf share layout logic but differ in content rendering (chronological events vs. editorial shelf). Extracting to a shared composable could be follow-up work.

src/components/discover/EditorialTimeline.vue
dmoo500: The server already filters artists by scan_window before sending them. The client-side MAX_DAYS filtering determines which events of an artist to display (e.g., show birthday but not death anniversary when both fall within scan_window but only one is within ±15 days of today). This avoids adding scan_window to RecommendationFolder (models change) and passing it from server to client.

If you'd like to modify this on the model and server sides, I can develop it; however, it was simpler to implement it this way.

src/components/discover/EditorialTimeline.vue
dmoo500: Fixed in 1908ac3e - removed minDaysLabel, maxDaysLabel, and fullDate.

src/components/discover/EditorialTimeline.vue
dmoo500: Fixed in 1908ac3e - changed from `Record<string, unknown>]` to `Record<string, Component>`.

src/components/discover/EditorialTimeline.vue
dmoo500: Fixed - simplified to always show localized date via Intl.DateTimeFormat instead of "Today"/"Yesterday"/"Tomorrow". More informative for a historical timeline and no i18n needed.

src/components/discover/EditorialTimeline.vue
marcelveldt: fine as-is and can always be adjusted in a follow-up

src/components/discover/EditorialTimeline.vue
```diff
@@ -0,0 +1,278 @@
...
+const MIN_ART = 120;
+const MAX_ART = 280;
+
+const track = ref<HTMLElement | null>(null);
+const tileArt = ref<number | null>(null);
+
+const gap = computed(() => props.gap ?? 14);
+
+const MAX_DAYS = computed(() => props.maxDays ?? 15);
+
```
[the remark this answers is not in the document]
dmoo500: The server already filters artists by `scan_window` before sending them. The client-side `MAX_DAYS` filtering determines which events of an artist to display (e.g., show birthday but not death anniversary when both fall within `scan_window` but only one is within ±15 days of today). This avoids adding `scan_window` to `RecommendationFolder` (models change) and passing it from server to client.

src/components/discover/EditorialTimeline.vue
```diff
@@ -0,0 +1,278 @@
...
+function mmddOffset(mmdd: string): number {
+  const today = new Date();
+  const year = today.getUTCFullYear();
+  const [mm, dd] = mmdd.split("-").map(Number);
+  // Try this year first; if more than 180 days away, try adjacent year.
+  const candidates = [
+    new Date(Date.UTC(year - 1, mm - 1, dd)),
+    new Date(Date.UTC(year, mm - 1, dd)),
+    new Date(Date.UTC(year + 1, mm - 1, dd)),
+  ];
```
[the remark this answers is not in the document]
dmoo500: Fixed by adding validation before date calculation. The function now returns NaN when mm or dd are out of valid ranges (1-12 for months, 1-31 for days), which handles MusicBrainz entries with 00 values for unknown month/day.

src/components/discover/EditorialTimeline.vue
```diff
@@ -0,0 +1,278 @@
...
+.timeline-scroll {
+  overflow-x: auto;
+  overflow-y: hidden;
+  scroll-behavior: smooth;
+  touch-action: pan-x pan-y;
+
+  /* Hide scrollbar */
+  scrollbar-width: none;
+  -ms-overflow-style: none;
+}
```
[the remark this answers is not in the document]
dmoo500: Added @media (prefers-reduced-motion: reduce) query that sets scroll-behavior: auto to respect user accessibility preferences.

src/components/discover/EditorialTimeline.vue
```diff
@@ -0,0 +1,278 @@
...
+          <template #subtitle>
+            <div style="display: flex; align-items: center; gap: 6px">
+              <component
+                :is="EVENT_ICONS[event.eventType]"
+                :size="12"
+                style="flex-shrink: 0; opacity: 0.75"
+              />
+              <span>{{ event.dateLabel }}</span>
+            </div>
+          </template>
```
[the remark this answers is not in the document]
dmoo500: Changed the wrapper from <div> to <span> with display: inline-flex to preserve the parent container's text-overflow: ellipsis behavior (block elements break ellipsis).

src/components/discover/EditorialTimeline.vue
stvncode: Rather than extracting a shared composable as follow-up work, I'd suggest a simpler resolution that actually shrinks this PR: build `EditorialTimeline` on top of `EditorialShelf` directly.

`EditorialShelf` already renders its cards through a default `<slot>` (that's how `HomeWidgetRows` feeds it `EditorialMediaCard`s), and it already accepts `title`, `provider`, `dimmed`, `tiles-per-view` and an `#actions` slot — everything the timeline shell re-implements. So the component can keep only its real logic (the `life_span` → events derivation and the icon + date subtitle) and delegate the rest:

```vue
<template>
  <EditorialShelf
    :title="title"
    :provider="provider"
    :dimmed="dimmed"
    :tiles-per-view="tilesPerView"
  >
    <template #actions><slot name="actions" /></template>

    <EditorialMediaCard
      v-for="event in timelineItems"
      :key="event.id"
      :item="event.artist"
    >
      <template #subtitle>
        <span class="timeline-event">
          <component :is="EVENT_ICONS[event.eventType]" :size="12" />
          <span>{{ event.dateLabel }}</span>
        </span>
      </template>
    </EditorialMediaCard>
  </EditorialShelf>
</template>
```

The script keeps `EVENT_ICONS`, `GROUP_TYPES`, `mmddOffset`, `dateLabel` and `timelineItems`, and drops `track`/`tileArt`/`updateTileArt`, the five layout constants, the ResizeObserver wiring and nearly all of the stylesheet — the component goes from ~294 lines to roughly 90, and `HomeWidgetRows` doesn't change at all.

Beyond removing the duplication, this restores three shelf behaviours the copy currently drops: mouse-wheel horizontal scrolling, the hover prev/next arrows, and scroll-position tracking. As it stands, the timeline row scrolls differently from every other row on the same page (no arrows, no wheel on desktop). It also means any future improvement to `EditorialShelf` applies to the timeline automatically instead of needing a second edit in the copy.

dmoo500: Sorry - I must have missed that, because I was considering displaying a timeline like 
```
|-----^------|
    today
```
with the dates above the artists and then showing the "normal" MediaCard (without adjusting the subtitles)... But apparently I tried too many things and overlooked this code optimization possibility.

stvncode (approved):

PR #2055, opened 2026-07-10
Make release workflows aware of stable patch builds

marcelveldt:
Stable patch builds (PEP 440 `.post` releases such as `2.17.186.post1`) are published from the `stable` branch. The current release automation assumes every release is a normal main-line `X.Y.Z`, which breaks once a `.post` release exists.

- **auto-release:** ignore `.post`/non `X.Y.Z` tags when picking the latest release, so the nightly still computes the correct next main-line version (a `.post` release would otherwise cause a version collision).
- **release:** for `.post` versions, open the server dependency-bump PR against the server `stable` branch instead of `dev`.

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -29,8 +29,10 @@ jobs:
       - name: Check for new commits
         id: check_commits
         run: |
-          # Get the latest PUBLISHED release (exclude drafts)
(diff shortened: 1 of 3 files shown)


marcelveldt:
Superseded by #2058 — that one has just the two release-workflow changes, without the stale Diagnostics.vue gate edit that snuck in here.

PR #2076, opened 2026-07-11
Fix home widget cleanup

teancom:
I was playing around with the new Safari MCP server, just to see what it could do. Well, one thing it can do is that it found a bug where the Top Pick's horizontal scrollbar could desync in mobile Safari and a problem with cleaning up MEDIA_ITEM_PLAYED when leaving Discover. The lifecycle bug also affects Chrome.

Commit message:

Guard the Top Picks resize observer against stale or missing DOM refs, and rebind it when the grid element changes. Register recommendation event cleanup synchronously so Vue can dispose it reliably on unmount.

### DIFF
--- a/src/components/HomeWidgetRows.vue
+++ b/src/components/HomeWidgetRows.vue
@@ -301,7 +301,6 @@ import {
   GripVertical,
   RefreshCw,
 } from "@lucide/vue";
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):
Thanks, @teancom !

PR #2139, opened 2026-07-14
Fix duplicate-ID console warnings from ListItem on settings pages

teancom:
I've been going through and using playwright to do some automated testing, looking for JavaScript console errors. This is the next thing that I found. 

Commit message:

useListItem passed value (defaulting to false via prop passthrough) to every v-list-item, so all items without an explicit value registered in Vuetify's nested registry under the same ID 'false', flooding the console with 'Multiple nodes with the same ID' warnings on settings/players, system, and tasks.

- Omit value when false/null/undefined so Vuetify falls back to a unique uid per item; falsy-but-valid values (0, '') still pass through
- Only pass link when truthy and drop the link:false default, which matches Vuetify's own default (undefined) with no behavior change for existing consumers
- Add regression tests for the value/link passthrough contract

### DIFF
--- a/src/composables/useListItem.ts
+++ b/src/composables/useListItem.ts
@@ -52,10 +52,17 @@ export interface ListItemEmits {
 
 export const useListItem = (props: ListItemProps) => {
   const listItemProps = computed(() => {
-    const { variant, showMenuBtn, ...vuetifyProps } = props;
+    const { variant, showMenuBtn, link, value, ...vuetifyProps } = props;
 
     const baseProps = {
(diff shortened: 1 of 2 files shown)


marcelveldt:
Should we really pay attention to this now considering we are migrating everything away from vuetify ?
The settings pages are already halfway in the migration - I have been slowly migrating one field type at a time in the past weeks

teancom:
@marcelveldt That's what I get for not paying attention. Happy to close this out, churn in code that's shortly going away is silly.

PR #2140, opened 2026-07-14
Fix guest api errors

teancom:
More playwright / automated testing findings. Split into three commits for reviewability but they're all conceptually tied together. I can open separate stacked PRs if that's preferred though.

Visiting the guest routes (`#/guest`, `#/guest/quiz`) produced console errors and raw server-error toasts for API failures that are entirely expected: the guest-only `can_listen_in` endpoints reject any non-guest session, and `music_quiz/info` isn't even registered when the music_quiz provider isn't loaded. This branch adds a `suppressGlobalError` option to `sendCommand` so best-effort commands can opt out of the global console.error + toast (the promise still rejects for the caller to handle), skips the listen-in availability probe outside guest access sessions, resolves `getMusicQuizInfo` to `null` without a server round-trip when the provider is absent, hardens the guest entry resolver against an info failure stranding it in its loading state, and gates `useMusicQuizPlayer`'s mount-time fetch on API initialization so a hard refresh of the guest quiz page won't briefly show a wrong "no quiz" state while the provider registry is still loading.

### DIFF
--- a/src/composables/useListenIn.ts
+++ b/src/composables/useListenIn.ts
@@ -1,6 +1,7 @@
 import { computed, onBeforeUnmount, onMounted, ref, watch } from "vue";
 import api from "@/plugins/api";
 import { EventType } from "@/plugins/api/interfaces";
+import { authManager } from "@/plugins/auth";
 import { webPlayer } from "@/plugins/web_player";
 
 let nextListenInOperationId = 0;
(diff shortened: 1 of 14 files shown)


src/composables/useListenIn.ts
```diff
@@ -80,10 +81,18 @@ export function useListenIn(options: UseListenInOptions) {
       canListenIn.value = false;
       return;
     }
+    // The server's can_listen_in endpoints are guest-only and reject any
+    // other role, so don't even ask outside a guest access session.
+    if (!authManager.isGuestAccessSession()) {
+      canListenIn.value = false;
+      return;
```
marcelveldt: we should fix this serverside instead because can listen in should also be allowed if we want the host to participate in the quiz

teancom: I can revert that bit and open a separate PR to the server. 👍

teancom: Done!

marcelveldt (approved):
Thanks, @teancom !

PR #2141, opened 2026-07-14
Refactor recommendations to be lazy loaded

chrisuthe:
## What / why

Brings the Discover page's recommendation rows into the unified Discover edit mode from **#2062**, adds the per-row `enabled_by_default` hint (heavy/noisy rows ship hidden-by-default), and makes the page load only the rows it actually shows — now via the server's two-method API:

- One **`getRecommendations()`** call returns the full row catalog **without items** (fast). Row shells render immediately.
- Each displayed row's items load in parallel via the new **`getRecommendationItems(provider, item_id)`** call, with a per-row loading state. Hidden rows are never fetched; toggling a row on fetches it on demand.

> **Note:** an earlier revision sent the server a `wanted` list on a single bulk call. That plumbing (and the localStorage row-catalog cache it needed) is gone — the rows call is cheap and always complete, so edit mode no longer needs a special full refetch either.

This **dovetails into #2062** rather than replacing it — players, Browse-by-genre, drag-reorder, the Top Picks scroller, the `discover.rows` preference, and the legacy-settings migration are all left as-is.

## Changes

- **`HomeWidgetRows.vue`**:
  - Page spinner only gates on the fast catalog call; recommendation shelves then stream in as their items arrive (`rowItemsMap`). A row whose items resolve empty is hidden in normal mode (unchanged end state); edit mode shows every row for toggling.
  - The **Top Picks hero** aggregation is unchanged in behavior, now sourced from the loaded items map; the refresh button re-fetches the rows feeding it.
  - The play-event refresh (debounced 1500 ms) reloads the catalog and re-fetches items for displayed rows. Transient refresh failures keep the previously loaded items instead of blanking rows.
- **`discoverRows.ts`**: unchanged model — `shown` opt-in list + `defaultHidden` input so a backend `enabled_by_default=false` row is hidden by default but openable in edit mode. Row ids (folder URIs) are unchanged, so existing saved `discover.rows` preferences carry over untouched.
- **API** (`plugins/api`): `getRecommendations()` takes no arguments; new `getRecommendationItems(provider, item_id)`.
- New `discover/utils/rowItems.ts` extracts the fetch-set/visibility logic (which rows need items given prefs + edit mode, empty-row visibility) as pure functions with unit tests.

Edit mode, with a default-off row hidden-until-opted-in:
image

## Depends on

- [music-assistant/server PR 4487](https://github.com/music-assistant/server/pull/4487) — **required**: this revision calls the new `music/recommendations/items` command; the page needs the paired server.
- [music-assistant/models PR 302](https://github.com/music-assistant/models/pull/302) (**merged**) — `enabled_by_default` on the shared `RecommendationFolder`.

## Testing

- `tests/components/discover/rowItems.test.ts` covers the fetch-set in normal vs edit mode, toggle-on triggering a fetch, and empty-items visibility; `discoverRows.test.ts` still covers the `shown`/`defaultHidden` model.
- `pnpm build`, `pnpm lint:check`, and `pnpm test:run` (full suite) all green.


src/components/HomeWidgetRows.vue
```diff
@@ -767,15 +797,33 @@ const unsubscribeRecommendations = api.subscribe(
   },
 );
 
+// Entering edit mode needs items for every row (including ones hidden until
+// now) so they render a usable empty/loaded state while toggling.
+watch(
+  () => props.editMode,
+  async (isEdit) => {
+    if (!isEdit) return;
+    await fetchMissingRowItems();
```
marcelveldt: do we really want to fetch items for every row here, including the hidden/default-off ones ? those are off by default exactly because they are heavy - opening edit mode now fans out an items call per row. rendering the hidden shells from the catalog metadata and fetching on unhide (which already works) seems enough

chrisuthe: I think the header/titles are enough to know if you want to add a section...

marcelveldt (approved):
Thanks, @chrisuthe !

PR #2143, opened 2026-07-14
Use the global i18n scope for the missing-players hint

teancom:
The i18n-t component defaults to the parent scope, but Players.vue has no local composer, so vue-i18n logged 'Not found parent scope' and fell back to the global scope anyway. Make the fallback explicit.


marcelveldt (approved):
Thanks, @teancom !

PR #2144, opened 2026-07-14
Use return values instead of deprecated next() in navigation guard

teancom:
vue-router deprecated the next() callback in favor of returning the redirect target (or nothing) from the guard. Convert the global guest/admin guard accordingly and drop an unused variable.

BTW, the same deprecation warning is also emitted by Vuetify's internal useBackButton composable (vuetify/lib/composables/router.js), so a few occurrences will remain in the console until the Vuetify migration replaces that dependency.


marcelveldt (approved):
LGTM, thanks @teancom !

PR #2146, opened 2026-07-15
Improve waveform seek preview with 3-level brightness

dmoo500:
Improves the waveform seek preview in the fullscreen player by differentiating the hover region visually from both the played and unplayed regions.

## Before

The hover position showed at the same brightness as the played (progress) region — hovering forward looked identical to the already-played waveform.

## After

Three distinct brightness levels:

| Region | Alpha | Description |
|---|---|---|
| Unplayed (dim) | 0.3 | Faint — not yet reached |
| Hover preview | 0.5 | Mid — where you will seek to |
| Played (progress) | 1.0 | Full — already played |

**Forward seek** (hover after progress): the region between progress and hover shows at mid-brightness, previewing what will play.

**Backward seek** (hover before progress): the region from 0 to hover dims to mid-brightness, showing the rewind target.

## Screenshots

waveform_jumpforward

waveform_jumpback

## Note

The hover preview currently uses the same color as the rest of the waveform but at a lower alpha. Depending on the album art color palette, the alpha-based difference may be more or less visible. A potential follow-up improvement could be using a distinct hover color (e.g. a mid-tone derived from the palette) to make the preview more clearly distinguishable — left for a future iteration.


MarvinSchenkel:
I honestly don't think this is an improvement. I left out the 'faded colour on hover' on purpose, because I think it gets crowded and confusing quite quickly. Especially when a track has already progressed near the end, going back feels 'clumsy' to me.

@stvncode @marcelveldt  what do you think here?

dmoo500:
Sorry – I wasn't aware that it was implemented this way intentionally.

Personally, I find it a bit confusing without the mouseover function, as I don't know the current position of the track. This is especially true when I want to skip ahead.

This feature could also be designed differently, perhaps by simply displaying the current position with a vertical line and maybe the time.

tbrasser:
In the bottom screenshot we're seeking back right?
In that case I would have expected the full and half dim to be swapped? (full=start to hover, half = hover to current)

then it's consistent:
- full is always whats already played (both on forward/backwars hover and default)
- half is always the jumped part on hover
- dim remains the unplayed part and indeed doesn't change during hover (only after click).

dmoo500:
## Rebased on `main` — adapted to SVG implementation ✅

Upstream changed the waveform rendering from Canvas to SVG. I've adapted the 3-level brightness feature accordingly:

**Canvas → SVG:**
- Replaced three `<canvas>` elements with three `<svg>` layers
- Applied alpha levels via `stroke-opacity` instead of canvas `globalAlpha`
- Same clip path logic for hover/progress regions

**Alpha levels (unchanged):**
- `0.3` — dim (unplayed)
- `0.5` — mid (hover preview)
- `1.0` — full (played/progress)

The visual behavior is identical — forward/backward seek preview works exactly as before.

dmoo500:
Fixed the backward seek consistency as suggested — full brightness now always shows what's played, half brightness always shows the jump region:

* Forward seek: full (0→progress), half (progress→hover)
* Backward seek: full (0→hover), half (hover→progress)

tbrasser:
> Fixed the backward seek consistency as suggested — full brightness now always shows what's played, half brightness always shows the jump region:
> 
> * Forward seek: full (0→progress), half (progress→hover)
> * Backward seek: full (0→hover), half (hover→progress)

Nice! Does it now look less busy when using it?

dmoo500:
Yep - it feels more natural than before. Now we have to wait and see if it gets approved/merged.

PR #2152, opened 2026-07-17
Also send updates when artist or title changes

joostlek:
So when I am listening to a radio stream via Music Assistant, Discord rich presence doesn't update, as the uri, state and player don't change. So by adding the title and artist to the equation, we will trigger more updates and allow rich presence to work correctly again


marcelveldt (approved):
Solid fix, thanks @joostlek !

PR #2180, opened 2026-07-23
Add help text to the DSP filters and gain stages and begin migration to shadcn

OzGav:
This branch does two related things for the DSP settings.

1. Help text for every filter and gain stage. Each DSP filter now explains what it does in a short info panel below its controls. Added for parametric EQ, tone control, gain, balance, and the fixed input and output gain stages. Gain and balance didn't have components of their own, so their sliders moved into a small wrapper to hold the note alongside them. Input and output get their own wording (headroom on the way in, player-volume-still-applies on the way out) rather than reusing the generic gain text.

2. Migrating the pieces this touches from Vuetify to shadcn. As part of the wider Vuetify→shadcn move:
 - The shared DSP slider (the slider + number box used by every filter) is now the shadcn Slider and Input.
 - All the help panels are now the shadcn Alert component.
 - The "DSP is turned off" message is now an amber warning that fades away in place when you enable DSP.

One behaviour note for reviewers: the frequency slider runs on a logarithmic scale. The old code relied on Vuetify treating a zero step size as "continuous"; the shadcn slider can't, so log sliders now use a small fixed step. No visible change, but it's why that math looks different.

The rest of the DSP settings screen (EditPlayerDsp.vue toolbars/menus/dialogs and DSPPipeline.vue) is still Vuetify and will be migrated in a separate PR, this branch only converts what the help-text work directly touched.


marcelveldt:
Just as a heads-up: I think this should be the last thing we change to the DSP views before we migrate it to shadcn.
We have a policy that we dont allow any more vuetify code being added because that means more work to convert.

OzGav:
Eeeppp… I do have all of my DSP filters I have been working on….

OzGav:
I’ll have to get Claude to get me up to speed on shadcn and see if I can switch to that?

OzGav:
OK PR description updated to indicate I have started migration to shadcn

src/views/settings/EditPlayerDsp.vue
```diff
@@ -203,6 +210,18 @@
           </v-card>
         </v-col>
       </v-row>
+
+      <Alert
```
stvncode: Can you use a v-if instead of playing with opacity ? Because the comp is still in the dom even if he's not needed

OzGav: Yes! Done.

stvncode (approved):
Perfect thanks @OzGav 🙏

PR #2184, opened 2026-07-23
Add "Play from beginning" for podcast episodes

chrisuthe:

Adds a **Play from beginning** action to the podcast episode context menu, letting a user replay an episode from position 0 while their saved resume position is preserved (unlike *Mark as unplayed*, which clears progress).

- **API wrapper** (`src/plugins/api/index.ts`): `playMedia` gains a trailing optional `start_from_beginning` boolean, passed through to the `player_queues/play_media` command.
- **Context menu** (`src/layouts/default/ItemContextMenu.vue`): a new single-episode item, shown only when the episode has progress to skip (fully played or a non-zero resume position), placed next to the existing *Mark as unplayed* action.
- **i18n** (`src/translations/en.json`): adds the `play_from_beginning` English string.

image

- Depends on music-assistant/server#4934

## Types of changes

- [x] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] New music/player/metadata/plugin provider — `new-provider`
- [ ] Documentation only — `documentation`

## Checklist

- [x] `pre-commit run --all-files` passes.
- [x] `pytest` passes, and tests have been added/updated under `tests/` where applicable.
- [ ] For changes to shared models, the companion PR in `music-assistant/models` is linked.
- [x] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked.
- [ ] I have raised a PR against the documentation repository targeting the main or beta branch as appropriate.


marcelveldt:
Strictly taken we would probably not have needed the extra boolean because frontend could also have done a "unmark as played" + play to get the same effect

chrisuthe:
> Strictly taken we would probably not have needed the extra boolean because frontend could also have done a "unmark as played" + play to get the same effect

Sure, I actually thought the same thing, and it hinges on me not digging out the rules on when the play log entry gets created. The audiobook path does this too, and I assumed the play log entry was somehow useful being left, but if it's as basic as I assume it is there is no upside and we can clean this up a bit for both podcast and audiobooks.

chrisuthe:
> Strictly taken we would probably not have needed the extra boolean because frontend could also have done a "unmark as played" + play to get the same effect

For the basic cases, including local/rss podcasts the reset + play would be fine, however where it could be an issue is a provider like AudiobookShelf (which is support item that originated this issue). 

mark_item_unplayed awaits only the local playlog delete; the provider reset is fire-and-forget (self.mass.create_task(on_played(..., position=0))). So mark_unplayed() + play() races: the play re-reads ABS live via get_resume_position, and the merge prefers the fresher provider timestamp over the just-deleted (epoch-0) playlog row — so if the background ABS DELETE hasn't landed, it resumes at the old position instead of 0.  Going deeper, awaiting the reset doesn't cleanly fix it either since the provider's get_resume_position falls back to session.current_time and open sessions aren't evicted, so a stale value can still win. start_from_beginning skips get_resume_position in the resolve path.

I could alter the provider code while simplifying to mark_unplayed + play if you want?

src/layouts/default/ItemContextMenu.vue
```diff
@@ -810,6 +811,30 @@ export const getContextMenuItems = async function (
     }
   }
 
+  // play from beginning (podcast episode with saved progress)
+  if (
```
stvncode: This re-derives the "has progress" condition that the block right above already establishes, and needs three as PodcastEpisode casts to do it. Could you do instead something like this: 

```ts
  if (
    items.length === 1 &&
    "fully_played" in items[0] &&
    "resume_position_ms" in items[0]
  ) {
    if (items[0].fully_played || items[0].resume_position_ms) {
      // mark unplayed
      contextMenuItems.push({
        label: "mark_unplayed",
        icon: History,
        action: async () => {
          await api.markItemUnPlayed(items[0]);
          (items[0] as PodcastEpisode).fully_played = false;
        },
      });
      if (items[0].media_type === MediaType.PODCAST_EPISODE) {
        contextMenuItems.push({
          label: "play_from_beginning",
          icon: RotateCcw,
          action: async () => {
            await api.playMedia(
              items[0].uri,
              QueueOption.PLAY,
              undefined,
              undefined,
              undefined,
              true,
            );
          },
          disabled: !store.activePlayer,
        });
      }
    } else {
      // mark played
      ...unchanged...
    }
  }
```

chrisuthe: Agreed, 0604d26 has this changed!

marcelveldt (approved):

PR #2188, opened 2026-07-24
Simplify player card warning styling (#63)

joperafe:
This PR addresses the issue reported in [Imporving warning on settings players
 #63](https://github.com/music-assistant/backlog/issues/63)
 
 Before:
image
image

 
 After:
 **Grid**
image

**List**
image

**What changed:**
- Unified the warning UI to use the existing chip component across both card and list views.
- Switched the player grid to CSS Grid so only the row containing a warning player grows


src/components/SettingsPlayerCard.vue
```diff
@@ -27,17 +26,16 @@
       </div>
 
       <!-- Player needs setup warning -->
-      <div
+      <v-chip
```
stvncode: Can you use the Badge component from Shadcn (components/ui) pls ?
This way you doens't need custom css spread into 2 different place

joperafe: Good point! I'll update this.

Thanks!

joperafe: @stvncode I've, finally, managed to wrap this up. Even though the current status was already simplified in the meantime :)

The description has been updated to the current status

src/views/settings/Players.vue
```diff
@@ -520,28 +499,31 @@ watch(
...
-  opacity: 0.7;
+@media (min-width: 960px) {
+  .players-grid {
+    grid-template-columns: repeat(2, 1fr);
+  }
 }
 
-.player-needs-setup {
-  border-left: 3px solid rgb(var(--v-theme-warning));
+@media (min-width: 1264px) {
```
joperafe, line 515: Ended up leaving these "magical" breakpoints as I haven't found any values defined already, for the css side at least..
Is there a preferred way to handle breakpoints globally, or is this ok?

OzGav:
I am thinking "Start Setup" should be "Reconfigure" as that is used elsewhere in the new UI? And it is blue elsewhere? Amber for the rest of the items looks good.

joperafe:
> I am thinking "Start Setup" should be "Reconfigure" as that is used elsewhere in the new UI? And it is blue elsewhere? Amber for the rest of the items looks good.

Tried both changes locally:
image

What do you think, does the blue button read ok next to the amber badge? To be honest I find the 2 colors there to be a bit much... 

On the wordig topic, I did find the "Reconfigure" button being used here and there but for this case I'm not so sure it would be representative of what the issue is for a given player. As far as I could see "Reconfigure" will be displayed after a player is configured, "Configure" for the first configuration. Would "Configure" make more sense here instead, to stay consistent with that?

Thank you for checking and for the feedback!

OzGav:
With the setup flow now you probably need to think about how you would get into the situation shown. I am thinking it generally would only be in a Reconfigure scenario? Maybe Fully Kiosk or MPD Players would get to that? As for colours I was just thinking about what is shown elsewhere in the UI. If it is a failure mode to be fixed then amber would make sense...?

src/views/settings/Players.vue
pierosavi: The preferred way is using tailwind's breakpoint system, I suggest this instead of the `.players-grid` class:
`<div v-else class="grid grid-cols-1 gap-6 md:grid-cols-2 xl:grid-cols-3">`

pierosavi:
@OzGav In Fully Kiosk you have to click on that button to set the password to finish setup. Even though it's not a failure in this case I prefer keeping "Start setup" (or "Configure" like @joperafe suggested) in amber to catch the attention.

pierosavi (changes requested):
Color/label discussion aside, code looks good and can be merged after update to tailwind's breakpoints

PR #2190, opened 2026-07-24
Add convolution DSP filter with impulse response library

OzGav:
Adds a convolution filter that applies a user-supplied impulse response, with a manager dialog to upload, name and remove IRs and a gain control. IR uploads are sent base64-encoded; removing an IR blanks any convolution filter still referencing it.

Built with shadcn components (Select, Dialog, Input, Alert, Button) and lucide icons. useDSPIRs keeps the list fresh by subscribing to DSP config updates, and the AudioProcessingDetails test mock gains the subscribe and getDSPIRs methods that composable needs.

image

image


src/components/dsp/DSPConvolution.vue
```diff
@@ -0,0 +1,158 @@
...
+const gain = computed({
+  get: () => filter.value.gain,
+  set: (value: number) => {
+    if (Number.isNaN(value)) return;
+    filter.value.gain = Math.min(MAX_GAIN, Math.max(MIN_GAIN, value));
+  },
+});
+
+const onUploaded = (ir: DSPIRMetadata) => {
+  void refresh();
```
stvncode: Since refresh() isn't awaited, ir_id gets set before the list contains the new IR, so the Select briefly labels the just-uploaded IR "Unavailable impulse response". Either await refresh() before assigning, or optimistically insert the returned DSPIRMetadata into the list — the upload response already has everything.

src/composables/useDSPIRs.ts
```diff
@@ -0,0 +1,67 @@
...
+      irs.value = [];
+      available.value = false;
+    }
+  };
+
+  // Removing an IR blanks the ir_id of every convolution filter that used it,
+  // so a config arriving with an unselected convolution filter is the signal
+  // that our list may be stale. Refreshing on every DSP update would refetch on
+  // each slider save instead.
+  const unsubscribe = api.subscribe(
```
stvncode: I think this is missing two cases: an IR uploaded from another session, and removal of an IR nothing referenced (no config update fires at all). Since the backend is still in flight, a dedicated DSP_IRS_UPDATED event would be cleaner. Also note each useDSPIRs() call has its own state, the editor's refresh after upload won't update other instances; module-level shared state would keep them consistent.

src/components/dsp/DSPIRManager.vue
```diff
@@ -0,0 +1,207 @@
...
+        ? error
+        : ((error as Error)?.message ??
+          t("settings.dsp.convolution.upload_failed"));
+  } finally {
+    uploading.value = false;
+  }
+};
+
+const removeIR = async (ir: DSPIRMetadata) => {
+  if (!confirm(t("settings.dsp.convolution.remove_confirm", [ir.name]))) return;
```
stvncode: Native confirm() here. We have the global delete-confirmation dialog for this (eventbus.emit("deleteConfirmationDialog", { message, onConfirm })), which keeps theming/wording consistent. Worth it here since this delete resets filters on every player/preset.

src/components/dsp/DSPConvolution.vue
OzGav: Agreed, optimistic is better — done. The upload response already has everything the list needs, so I add it to the list first, then select it, then refresh in the background. No more flash of the "unavailable" label.

src/composables/useDSPIRs.ts
OzGav: On the IR event I agree. Two questions though. Does it carry the full list like `dsp_presets_updated` does, and does it fire on upload as well as remove? Also, on remove, will it fire alongside the config updates that blank `ir_id`? If it does I'll drop the guesswork in useDSPIRs entirely. Happy either way on whether this PR waits for it or merges with the current heuristic as a stopgap.

OzGav: Actually after looking at the backend I have the answered this myself. It will carry the full list like the presets event does, and fire on both upload and remove. On a remove you get the per-player config updates and the IR event, but they're self-contained so there's nothing to match up between them. So I've dropped the guesswork and subscribed to the new event instead.

src/components/dsp/DSPIRManager.vue
OzGav: Thanks, done!

src/composables/useDSPIRs.ts
stvncode: The event already exists in server#4947 / models#336 (dsp_irs_updated), so: yes it carries the full list (data=self.get_dsp_irs(), same as presets), yes it fires on upload as well as remove, and yes on remove it fires after the PLAYER_DSP_CONFIG_UPDATED blanking events, so it's authoritative when it lands. Only caveat: presets get blanked without any event.

So you can drop the heuristic entirely, make useDSPIRs mirror useDSPPresets (set irs.value = event.data on the event) and add the enum member. No point merging the stopgap since this PR waits on server#4947 anyway, and that's the PR shipping the event.

OzGav: All done and pushed. The guesswork is gone and the handler just takes the list straight off the event, same as the presets one. Yeah, I noticed that gap with `dsp_presets_updated` as well, and I'll submit a separate PR to fix that. Agreed on waiting for server#4947.

stvncode (approved):
Thanks for the changes 🙏

PR #2195, opened 2026-07-25
Add transpose DSP filter

OzGav:
Adds a transpose filter with a semitone slider (min / original key / max scale) and a numeric readout, plus an A=432Hz preset button with a hover hint. Built with shadcn components.

image

### DIFF
--- a/src/components/dsp/DSPTranspose.vue
+++ b/src/components/dsp/DSPTranspose.vue
@@ -1,0 +1,136 @@
+<template>
+  <div class="@container p-4">
+    <div class="flex flex-wrap items-center gap-4">
+      <div class="min-w-0 flex-1 px-1 pb-6">
+        <Slider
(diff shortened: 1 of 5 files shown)


stvncode:
@OzGav, not really a huge fan of the disposition (having two inputs on top of each other with a text that is centered with the first one). Could be a question mark icon with a tooltip at the end if the second input is just an information 👍 
You want me to have look and propose something ?

OzGav:
Yes very happy with your expert take on this!

OzGav:
Oh actually do you mean the A=432? If so, that is actually a preset button.

stvncode:
Hmm, let me try to come up with something and i'll send you a screenshot to see if that's ok with you 👍

stvncode (approved):
🙏

PR #2198, opened 2026-07-25
 Migrate the DSP pipeline, toolbars and dialogs to shadcn

OzGav:
Continues the DSP migration off Vuetify. Rebuilds the filter list, both toolbars and both dialogs on shadcn, leaving `EditPlayerDsp.vue` with six Vuetify usages instead of 55.

> [!NOTE]
> Based on `dsp-filter-help-text`, so marking this as draft until that is merged.

## The filter list

The old list of filters and pill buttons are gone. The list now uses the same style as `AudioProcessingDetails`

A deactivated filter drops both its dot and its tick.

## Filter icons are now shared

`dspFilterIcon()` lives next to `dspFilterText()` in `helpers/audioProcessing.ts` and is used by both the DSP settings view and the audio chain popover. **This changes the Audio Pipeline slightly**: it used to draw `SlidersHorizontal` for every DSP stage and now distinguishes an EQ from a gain filter.

The map is keyed by the filter type's string value rather than the enum, so it already covers the filters waiting on their own branches (safety limiter, compressor, high/low-pass, convolution, transpose, stereo width and crossfeed)

## Chrome

- Page toolbar → flex header with `DropdownMenu`, `Switch` and `Button`s
- Stage toolbar → same, with ghost icon buttons
- Both dialogs → `Dialog` with `Input`, `Label` and `Select`
- Panel `v-card`s → divs on `bg-muted`
- `getButtonClass()` and every `$vuetify.theme` reference deleted

## Not in this PR

- `DSPParametricEQ`
- `v-container`/`v-row`/`v-col` and `useDisplay().mobile` - the layout skeleton. Worth flagging that this view's `mobile` (Vuetify, <960px) disagrees with the rest of the app's `store.mobileLayout` (UA detection, <769px, and the force-mobile setting). Unifying those changes behaviour, so it's deferred.

image


marcelveldt:
the font size of the left options menu is huge and seems completely out of line with the rest of the (shadcn based) UI components ?

OzGav:
I will drop it down a bit but same size was too small when I looked at it. So I was looking at similar overall size to the existing component.

OzGav:
The font size is now the same as the titles in the components on the right

image

src/helpers/audioProcessing.ts
```diff
@@ -31,6 +45,33 @@ export function areDSPConfigsEqual(left: DSPConfig, right: DSPConfig): boolean {
...
+  [DSPFilterType.BALANCE]: ArrowLeftRight,
+  safety_limiter: ArrowDownToLine,
+  compressor: ChevronsDownUp,
+  convolution: Waves,
+  transpose: ArrowUpDown,
+  stereo_width: Expand,
+  crossfeed: Blend,
+};
+
+export function dspFilterIcon(filter: DSPFilter): Component {
```
stvncode, line 63: guesses at the not-yet-existing high_low_pass shape through two casts, and defaults an unknown/renamed mode to "high-pass" rather than to the neutral icon.

src/components/dsp/DSPPipeline.vue
```diff
@@ -188,28 +180,130 @@ const { onHold, onTouchStart, swallowClickAfterHold } = useHoldToOpenMenu(
...
+  display: flex;
+  min-width: 0;
+  align-items: center;
+  gap: 8px;
+  width: 100%;
+  /* Spacing lives on the item, not the row, so the rail stays unbroken. */
+  margin-block: 3px;
+  padding: 6px 8px;
+  border-radius: var(--radius-md);
+  font-size: 1.125rem;
```
stvncode, line 261: font-size still exceeds the view's own h2 (text-base) and is ~1.6× the AudioProcessingDetails text it's said to match. Only partly addresses the existing review comment.

src/helpers/audioProcessing.ts
```diff
@@ -31,6 +45,33 @@ export function areDSPConfigsEqual(left: DSPConfig, right: DSPConfig): boolean {
...
+const filterIcons: Record<string, Component> = {
+  [DSPFilterType.PARAMETRIC_EQ]: SlidersHorizontal,
+  [DSPFilterType.TONE_CONTROL]: AudioWaveform,
+  [DSPFilterType.GAIN]: Gauge,
+  [DSPFilterType.BALANCE]: ArrowLeftRight,
+  safety_limiter: ArrowDownToLine,
+  compressor: ChevronsDownUp,
+  convolution: Waves,
+  transpose: ArrowUpDown,
+  stereo_width: Expand,
```
stvncode: Can you use here STEREO_WIDTH, CROSSFEED and TRANSPOSE instead of raw text please ?

src/components/dsp/DSPPipeline.vue
OzGav: The pipeline font size matches the h3 which is the title of the component right next to the pipeline, both are 1.125rem. The h3 has no text-* class, it picks that up from the legacy heading scale in src/styles/global.css:175. Agreed it doesn't match the view's h2, but that's the pre-existing behaviour, the shadcn migration gave the h2 text-base (1rem) while bare h3 is still 1.125rem, so the panel title is already larger than the view header regardless of what the pipeline does.

src/helpers/audioProcessing.ts
OzGav: Yes. Done.

src/helpers/audioProcessing.ts
OzGav: Ahhh. I did not clearly sort out the stacking of these branches. The high/low-pass filter is actually #2158's and it defines `DSPFilterType.HIGH_LOW_PASS` and a `HighLowPassMode` enum properly. This branch is independent off main so it can't import them yet, hence the casts.

I've fixed the default in the meantime so an unknown mode now falls back to the neutral icon instead of assuming high-pass.

If we merge #2158 first I'll rebase this and drop the casts. `safety_limiter` and `compressor` become enum members in the same pass, so no follow-up PR needed.

Maxim is looking at the backend PRs now so hopefully they will get merged shortly and unblock these.

stvncode (approved):
Perfect @OzGav  thanks 🙏 
Same as the previous, it just need a rebase

OzGav:
Sorry I need to pay more attention. Rebased!

PR #2199, opened 2026-07-25
Migrate the PEQ to shadcn

OzGav:
Third of the DSP shadcn migrations. No Vuetify left in `DSPParametricEQ.vue`.

> [!NOTE]
> Based on `dsp-pipeline-shadcn`, so marking as draft until that is merged.

- The frequency graph now reads `--foreground`, `--border` and `--muted-foreground` off the canvas instead of branching on Vuetify's dark flag with hardcoded hex. The grid is lighter in light mode as a result.

image


stvncode (approved):
Love this! Thanks 🙏

PR #2207, opened 2026-07-25
Show per-provider search status in the search view

jozefKruszynski:
Surfaces per-provider search status in the global search view, using the new
`provider_search_statuses` field from the search API.

- Replaces the timing heuristic for slow providers with the real reported
  status (`complete` / `timeout` / `failed`), falling back to the old heuristic
  when talking to a server that doesn't report it.
- Shows a brief info toast for terminal outcomes only: a provider that failed,
  or that returned no matches while others had results. Transient/retrying
  states aren't toasted.

Companion PRs:
https://github.com/music-assistant/models/pull/331
https://github.com/music-assistant/server/pull/5028

### DIFF
--- a/src/views/Search.vue
+++ b/src/views/Search.vue
@@ -99,6 +99,7 @@ import { MediaType } from "@/plugins/api/interfaces";
 import { $t } from "@/plugins/i18n";
 import { store } from "@/plugins/store";
 import { computed, onBeforeUnmount, onMounted, ref, watch } from "vue";
+import { toast } from "vue-sonner";
 
 // local refs
 const searchHasFocus = ref(false);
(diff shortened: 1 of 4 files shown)


src/views/Search.vue
```diff
@@ -142,11 +143,58 @@ const {
...
+const noteworthyStatusLabel = function (state: "no_matches" | "failed") {
+  if (state === "no_matches") return $t("search_status_no_matches");
+  return $t("search_status_failed");
+};
+
+// Toast targets that reached a terminal outcome that didn't come back clean:
+// failed, or no_matches once other results are on screen. The transient
+// retrying state is not toasted since a toast can't be retracted when the
+// provider recovers. At most one toast per (search-run, target, outcome).
+const toastedOutcomes = new Set<string>();
```
stvncode, line 176: toastedOutcomes never clears between search runs (reset branch unreachable due to same-tick repopulation + pre-flush watcher). Later failures/no-matches are silently swallowed

jozefKruszynski:
This one is a bit stale and broken and needs rethinking. Let's move it to next release ideas.

marcelveldt:
Closed as it needs another approach

PR #2217, opened 2026-07-27
Fix missing cover art in fullscreen player on older browsers

### LINKED ISSUE #5717: The cover is not visible on older Android devices.
### What server version of Music Assistant has the issue?

2.9.3

### How is the MA server installed?

Home Assistant OS Addon

### Mandatory: Carefully read the Troubleshooting FAQ and confirm that

- [x] I have examined the logs and tried to resolve this issue
- [x] I have fixed any errors or warnings in the logs that relate to tags
- [x] I am not running MA across a VPN, VLAN, subnet, behind a firewall, or using local SSL or have any other complex network setup
- [x] I am not using or have disabled tools such as AdGuard, Pi-hole or pfSense and retested
- [x] I have checked my network setup to ensure mDNS/multicast is not being blocked
- [x] I have reviewed the [Open](https://github.com/music-assistant/support/issues) and [Closed](https://github.com/music-assistant/support/issues?q=is%3Aissue%20state%3Aclosed) Issues and [Discussions](https://github.com/orgs/music-assistant/discussions)
- [x] I have reviewed the applicable [player](https://www.music-assistant.io/player-support/) or [music provider](https://www.music-assistant.io/music-providers/) documentation
- [x] I have reviewed the [MA Status Page](https://github.com/orgs/music-assistant/discussions/709)
- [x] I have tried restarting MA and rebooting the host

### As Applicable: Carefully read the Troubleshooting FAQ and confirm that

- [ ] If using HA, I have confirmed the internal URL is set correctly
- [ ] I have tried a wired connection for issues related to interrupted or poor playback quality
- [ ] If the problem relates to a device then I have checked the device settings
- [x] If it is a frontend issue, I have tried a different widely used browser
- [ ] For voice problems, I have sought help elsewhere before returning here
- [ ] For playback problems, I have recycled power to the physical device

### The problem

Image

Problem where the cover is not visible on the track information screen.

### How to reproduce

Older Android 9 devices, homeassistant app browser.

There were no problems when using the Chrome browser on the same device.

### Music Providers

apple music

### Player Providers

google

### Full log output

.

### Additional information

It seems like a problem caused by the low version of the built-in browser on older devices.
Since the issue is that only the cover is not visible, I think it can be sufficiently fixed.

thank you.

### What version of Home Assistant Core (if used) are your running

2026.6

### What type of installation are you running?

Home Assistant OS

### On what type of hardware are you running?

ODROID

### Have you included ALL of the information specified in the Troubleshooting FAQ or explained why you cannot

- [x] Yes

MarvinSchenkel:
The album cover is invisible on the fullscreen player in the Home Assistant companion app on older Android devices, while the rest of the screen renders fine. The same device shows the cover correctly in Chrome.

`.main-media-details-image .v-img` is sized with container query units (`min(100cqi, 100cqh)`), which need Chromium 105+. Older engines discard those declarations at parse time, leaving `width`/`height` at `auto`. Vuetify's `v-img` has no intrinsic width, so it collapses to 0x0 and the cover disappears.

Adds a plain fallback ahead of the container query pair. Modern browsers still use the container query sizing, older ones get a usable image instead of nothing.

- Fall back to `width: 100%` / `height: auto` so the image stays visible when container query units are unsupported; `height: auto` lets Vuetify's 1:1 sizer keep it square

Fixes music-assistant/support#5717

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
+++ b/src/layouts/default/PlayerOSD/PlayerFullscreen.vue
@@ -1487,6 +1487,9 @@ watchEffect(() => {
   container-type: size;
 }
 .main-media-details-image .v-img {
+  /* Fallback for engines without container query units; they drop the pair below and collapse the image to 0x0. */
+  width: 100%;
+  height: auto;
   width: min(100cqi, 100cqh);
   height: min(100cqi, 100cqh);
   flex: 0 0 auto;


stvncode (approved):
🙏

PR #2228, opened 2026-07-29
Reject in-flight API commands when the connection drops

MarvinSchenkel:
A command still awaiting its reply when the websocket goes away is never settled: `commands` is only drained when a matching result message arrives. The caller's promise hangs forever, so its `finally` never runs and any loading flag it set stays set — a view awaiting a command behind a loading flag is stuck there until a full page reload, including its own refresh button if that is disabled while loading.

- reject pending commands on explicit disconnect and on transport `reconnecting` / `failed` / `disconnected`, where the socket is already gone and no reply can arrive
- reject with a dedicated `ConnectionLostError` so callers can tell a dropped connection from a real command failure
- keep `authenticateWithToken` out of `AUTH_REQUIRED` when the failure was a dropped connection, so a blip mid-re-auth no longer bounces the user to the login screen
- clear the matching partial-result buffers, which leaked an entry per dropped command
- move the inline `stateChange` closure into a private method so the behaviour is reachable from tests

Callers now see a rejection where they previously saw silence, so a connection blip can surface an error in views that catch and report; `ConnectionLostError` gives any view that would rather stay quiet a way to filter it.

### DIFF
--- a/src/plugins/api/index.ts
+++ b/src/plugins/api/index.ts
@@ -86,6 +86,14 @@ export enum ConnectionState {
   FAILED = "failed", // Connection failed permanently
 }
 
+/** Rejection for in-flight commands that can never be answered because the connection dropped. */
+export class ConnectionLostError extends Error {
+  constructor() {
+    super("Connection lost");
+    this.name = "ConnectionLostError";
+  }
+}
+
 export class MusicAssistantApi {
   private transport?: ITransport;
   private _throttleId?: ReturnType<typeof setTimeout>;
@@ -337,8 +345,12 @@ export class MusicAssistantApi {
 
       return result;
     } catch (error) {
-      // Token is invalid - require user authentication
-      this.state.value = ConnectionState.AUTH_REQUIRED;
+      // a dropped connection says nothing about the token; leave the
+      // reconnect flow intact so re-auth runs again after reconnect
+      if (!(error instanceof ConnectionLostError)) {
+        // Token is invalid - require user authentication
+        this.state.value = ConnectionState.AUTH_REQUIRED;
+      }
       throw error;
     }
   }
@@ -3267,7 +3279,7 @@ export class MusicAssistantApi {
     this.commands.clear();
     this.partialResult = {};
     for (const command of pending) {
-      command.reject(new Error("Connection lost"));
+      command.reject(new ConnectionLostError());
     }
   }
 
--- a/src/plugins/api/index.ts
+++ b/src/plugins/api/index.ts
@@ -3195,7 +3195,13 @@ export class MusicAssistantApi {
         reject,
         suppressGlobalError: options?.suppressGlobalError,
       });
-      this._sendCommand(command, args, cmdId);
+      try {
+        this._sendCommand(command, args, cmdId);
+      } catch (error) {
+        // a synchronous throw here rejects the promise but leaves the map entry behind
+        this.commands.delete(cmdId);
+        throw error;
+      }
     });
   }
 
@@ -3212,7 +3218,7 @@ export class MusicAssistantApi {
       this.state.value !== ConnectionState.AUTHENTICATED &&
       this.state.value !== ConnectionState.INITIALIZED
     ) {
-      throw new Error("Connection lost");
+      throw new ConnectionLostError();
     }
 
     if (!msgId) {
@@ -3232,10 +3238,15 @@ export class MusicAssistantApi {
     const msgStr = JSON.stringify(msg);
 
     if (!this.transport) {
-      throw new Error("No connection available");
+      throw new ConnectionLostError();
     }
 
-    this.transport.send(msgStr);
+    try {
+      this.transport.send(msgStr);
+    } catch {
+      // transport throws untyped errors; callers filter on ConnectionLostError
+      throw new ConnectionLostError();
+    }
   }
 
   public async fetchState() {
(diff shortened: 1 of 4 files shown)


stvncode (commented):
Ask Claude for a review and this was the critical/problem errors: 

- Login.vue:679-687 wipes the stored token (authManager.clearAuth() / leaveGuestSession()) on a ConnectionLostError, because tryStoredTokenAuth() is always called with no argument. A transient blip during auto-login now forces a manual re-login. Filter ConnectionLostError in that catch.

- sendCommand:3117-3124 never removes the commands entry when _sendCommand throws, so the map grows permanently and _failPendingCommands' notion of "pending" is inaccurate.

- handleResultMessage:2590-2622 can toast an error for a command that already rejected and had opted out via suppressGlobalError, since the flag is lost once the map is cleared.

- The authenticateWithToken guard (337-345) only catches ConnectionLostError, but three other connection-loss errors exist (_sendCommand:3140, :3160, WebSocketTransport.send:100). A drop before the send still bounces the user to AUTH_REQUIRED via App.vue:470's re-auth path, the bug bullet 3 claims to fix. Make all four throw ConnectionLostError

marcelveldt:
Converted to draft because the review comments first need to be addressed

src/plugins/api/index.ts
```diff
@@ -3168,6 +3193,21 @@ export class MusicAssistantApi {
     }
   }
 
+  /** Reject commands still awaiting a reply; a dropped socket can never answer them. */
+  private _failPendingCommands(): void {
```
marcelveldt, line 3283: This drains entries that are already in the map, but `sendCommand` can leak one on the way in: it does `commands.set(cmdId, ...)` and then calls `_sendCommand`, which throws `Connection lost` when the state doesn't allow sending. The promise rejects fine (executor throw), but that entry stays behind - one per attempt while we're down, and no state change follows to clean it up. Worth a try/catch around the `_sendCommand` call that deletes the entry and rethrows.

src/plugins/api/index.ts
```diff
@@ -2562,6 +2538,55 @@ export class MusicAssistantApi {
     this.signalEvent(msg);
   }
 
+  private handleTransportStateChange(state: string): void {
+    console.log("[API] Transport state changed to:", state);
+    this.transportState.value = state;
+
+    // the socket is already gone here, so nothing in flight can be answered
```
marcelveldt, line 2598: This only fires when the transport reports a state change, so a half-open socket still hangs: transport thinks it's connected, replies just stop, nothing rejects. An optional timeout on `sendCommand` would close that one. In #2248 I gave the clock probe its own 2s deadline for exactly this reason - if a timeout lands here, the probe can drop it.

marcelveldt (commented):
Nice, this is the right fix - the hanging loading flag is exactly what bites users. Two gaps I ran into while building the server time sync in #2248, both on the send side rather than the reply side.

src/plugins/api/index.ts
MarvinSchenkel: Fixed in ad3a0b2e — `sendCommand` wraps the `_sendCommand` call, deletes the map entry and rethrows, so a synchronous throw no longer leaves an entry behind. Regression test asserts the map is empty after the rejection.

src/plugins/api/index.ts
MarvinSchenkel: Agreed — a half-open socket never triggers a transport state change, so this PR doesn't cover it. Keeping the scope here to the state-driven rejections and picking up an optional per-command timeout on `sendCommand` as a follow-up; the #2248 clock probe can drop its own 2s deadline once that lands.

src/plugins/api/pendingCommands.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  it.each(["reconnecting", "failed", "disconnected"])(
+    "rejects them when the transport goes %s",
+    async (state) => {
+      const { api } = connectedApi();
+      const inFlight = api.sendCommand("ai_radio/stations/list");
+
+      api["handleTransportStateChange"](state);
+
```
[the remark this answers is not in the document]
MarvinSchenkel: TypeScript deliberately allows element access (`api["private"]`) to private members — it's the standard escape hatch for tests. Verified: `vue-tsc --noEmit` passes on this branch, so `pnpm build` stays green. No change needed.

MarvinSchenkel:
@stvncode addressed in ad3a0b2e:

- **Login.vue token wipe**: fixed — `tryStoredTokenAuth` now skips `clearAuth`/`leaveGuestSession` when the error is a `ConnectionLostError`; a blip says nothing about token validity and the reconnect flow retries.
- **`sendCommand` map leak**: fixed — the entry is deleted when `_sendCommand` throws synchronously (also flagged inline by @marcelveldt), with a regression test.
- **all four throw sites**: fixed — the state guard and no-transport check now throw `ConnectionLostError`, and `transport.send` is wrapped so its untyped throw is rethrown as `ConnectionLostError` (kept the transport itself untouched to avoid an import cycle). Regression test covers a drop before the send no longer landing in `AUTH_REQUIRED`.
- **`suppressGlobalError` lost after map clear**: leaving as is — the WebSocket delivers queued messages before `onclose` fires, so a result arriving after `_failPendingCommands` is practically unreachable, and the toast-for-unknown-id path is pre-existing intentional behavior ("always handle error" branch).

marcelveldt (approved):

PR #2237, opened 2026-07-31
Maintain constant album art size on now playing view

OzGav:
While playing with the Audio AI plugin I noticed that the album art and text row were jumping in size and position.

This appears to be because the album and artist subtitles were removed entirely when empty and the artwork was sized to the remaining space. The subtitle rows now always render, using a non-breaking-space placeholder when the metadata is missing, so the text block height is constant and the artwork keeps a fixed size and position for a given view size


stvncode (approved):
Thanks 🙏

PR #2238, opened 2026-07-31
Fix podcast episode played/unplayed state in the episode list

OzGav:
Three related fixes to marking podcast episodes as played or unplayed. 

**Marking an episode played left it stuck that way.** The mark-played and mark-unplayed menu options only appeared on episodes that carried a resume position, but marking an episode played deleted that value from the item so the "mark unplayed" option disappeared and there was no way to undo it. The played state is now tracked properly and the menu decides what to show based on the item's type instead.

**Partially played episodes could not be marked as played.** The UI only offered "mark unplayed". In-progress episodes now offer both options, matching how a mixed multi-selection already behaved.

**Marking a large selection could drop the server connection.** Selecting every episode of a large podcast sent one command per episode all at once which in my test was around 1100 of them. That overwhelmed the server's message buffer for the client, and it responded by dropping the connection, after which the UI silently stops working until it reconnects (seeking, for example, does nothing). The commands are now sent a few at a time, which keeps things well within limits and is still fast.

Tested with my dev server by running this branch and main side by side against the same backend and the connection drop reproduces on main and does not on this branch.

Update 12 Augst:
While this branch was open, a separate change landed on main that does the same job as one part of this PR: making sure that commands still waiting for a reply are cancelled when the connection drops, rather than being left hanging forever.

Both versions ended up in the file after merging main. They don't clash in an obvious way, so the merge went through quietly, but at runtime main's version always runs first and this branch's version was left with nothing to do. The only visible sign was a test in this PR failing, because it was still checking for the wording this branch used, while the actual failure now comes from main's version.

The fix is to drop this branch's copy and rely on the one in main, which covers more cases: it also fires when the connection is reconnecting or has failed, not just when it closes or is shut down deliberately.


src/helpers/concurrency.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Run `worker` over `items` with at most `concurrency` operations in flight,
+ * returning the results in input order. Bulk actions can span thousands of
+ * items; sending every command at once overruns the server's per-client
+ * message buffer and costs us the connection.
+ *
+ * Unlike Promise.all, a failing item does not cancel the remaining work: every
+ * item is processed and the first error is rethrown afterwards.
+ */
+export async function runWithConcurrency<T, R>(
```
stvncode, line 13: runWithConcurrency stalls permanently and silently if the connection drops mid-run. 
sendCommand never rejects pending commands on disconnect, so hung slots freeze the cursor and the remaining items are never sent.

src/helpers/concurrency.ts
```diff
@@ -0,0 +1,42 @@
...
+          hasError = true;
+          firstError = err;
+        }
+      }
+    }
+  };
+
+  await Promise.all(Array.from({ length: workerCount }, runWorker));
+
+  if (hasError) throw firstError;
```
stvncode, line 40: The rethrow becomes an unhandled rejection. ItemContextMenu.vue:189 calls action() without await or .catch(). Combined with the per-command toast.error, a partial failure over 1100 items is both noisy and unreported as a whole

stvncode:
Also, PodcastDetails.vue:86-116 still holds the pre-fix logic (fully_played || resume_position_ms → else mark_played), so a partially-played episode can't be marked played there. It also appears to be dead code, neither onMenu nor onClick is referenced anywhere in that file's template. Leaving a divergent copy of the logic this PR just fixed is a future-bug trap. You can either wire it up and reuse the fixed path, or delete it. Wdyt ?

src/helpers/concurrency.ts
OzGav: Good catch, and it turned out to be worse than described. A dropped socket goes through `stateChange` to reconnecting and never calls `disconnect()`, so the pending commands map was untouched on exactly the path that matters. `disconnect()` didn't clear it either.

Fixed at the source rather than in the helper. Pending commands are now rejected when the transport closes, and on an explicit disconnect. The server doesn't keep message ids across connections so those results can never arrive. This fixes every caller, not just the bulk path. Two tests added.

src/helpers/concurrency.ts
OzGav: Correct on both counts. `onSelect` called `action()` without awaiting or catching, so the rethrow escaped.

Two changes. `onSelect` now catches, so no menu action can produce an unhandled rejection. And the bulk path no longer depends on the throw at all. Per item failures are collected and reported once.

For the toast noise, `markItemPlayed` and `markItemUnPlayed` now take an options argument, and the bulk path passes `suppressGlobalError`. So 1100 failures give you one summary toast instead of 1100.

OzGav:
Confirmed dead, Deleted both plus the now unused imports.

Went with deleting rather than wiring up because the shared context menu path already covers this view. Wiring it would reinstate the duplicate logic you were warning about.

stvncode (approved):
Nice thanks @OzGav 🙏

stvncode:
Just need a rebase @OzGav

PR #2266, opened 2026-08-04
Keep progress indicators in sync with the play state they follow

marcelveldt:
The progress indicators take their position from the queue, but some of them checked the *player* to decide whether that position was moving. When the two briefly disagree, the position and the play state come from different places and the UI goes wrong: the lyrics freeze while the track keeps playing, the player bar stops moving, and the party progress bar jumps back to zero.

Each indicator now takes its play state from the same source it takes its position from.

- Start and stop the lyrics animation from the active queue's state
- Keep the player bar ticking based on whichever source supplied the position
- Take the party track card play state from the queue
- Add test coverage for the lyrics position

### DIFF
--- a/src/composables/lyrics/useLyricsElapsedTime.ts
+++ b/src/composables/lyrics/useLyricsElapsedTime.ts
@@ -3,7 +3,8 @@
  *
  * Provides a reactive `elapsedTime` ref that updates at ~60fps while playing,
  * suitable for smooth lyrics synchronization. Automatically starts/stops the
- * rAF loop based on playback state and an optional `enabled` guard.
+ * rAF loop based on the active queue's playback state and an optional
+ * `enabled` guard.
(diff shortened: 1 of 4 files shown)


marcelveldt:
Copilot flagged that the gate could stop the loop while the player-timing fallback is still advancing. Not reachable in practice: every write to `api.queues` seeds `api.queueElapsedTime` from the same queue object (index.ts 2536/2545/3259), and the delete at 2578 removes both. So whenever `activePlayerQueue` exists the resolver always takes the queue branch — the player fallback only applies when there's no active queue, and the gate already required `queue.active` before this change.

Deriving the gate from `resolveActiveTiming()` would also pull the ~1Hz `queueElapsedTime` map into the watchEffect deps, making it re-run on every time update for no gain.

PR #2270, opened 2026-08-05
Hide status labels and alerts whose dependency is not met

marcelveldt:
Config entries can declare a `depends_on`, which is meant to keep them out of sight until the setting they depend on is switched on. In practice the frontend only ever *disabled* such an entry. That works for inputs (a greyed-out field), but labels, alerts, dividers and images have nothing to disable, so they rendered at full strength as if the dependency were met.

Visible today as, for example, the Sonic Similarity settings showing "Character engine: disabled" and "Text encoder: disabled" status rows even with both engines switched off, and Smart Playlist warning about a missing AI engine with AI descriptions switched off.

Fixing that surfaced a second problem underneath it. Pressing an action button that does return entries (Verify SSL certificate, Rebuild index, ...) made the form fall back to its default values, because an action responds with entry definitions that carry no values and the settings forms replaced their state with it wholesale. Dependent entries then read the wrong dependency value — so Verify SSL certificate would have hidden the very label holding its result. #2260 covered the empty-response case; this covers the rest.

Changes:
- Added a shared list of entry types that render no interactive control (divider, label, alert, image)
- An unmet dependency now hides those types instead of disabling them, on the settings page and in the setup flow dialog
- Inputs and action buttons keep greying out as before
- Action responses now keep the values already in the form, on the provider, player and core settings forms
- Test coverage for both forms and all three action paths

### DIFF
--- a/src/components/SetupFlowDialog.vue
+++ b/src/components/SetupFlowDialog.vue
@@ -377,7 +377,7 @@ let expiryReconciledFor: string | null = null;
 // reconnect; the double underscores keep it out of reach of server-sent slugs
 const SESSION_ENDED_STEP_ID = "__session_ended__";
 
-// terminal steps: closing them must not abort (the flow already ended server-side)
+// entry types that carry no value, so they take no part in validation or submission
 const PRESENTATIONAL_TYPES = [
   ConfigEntryType.DIVIDER,
   ConfigEntryType.LABEL,
@@ -386,6 +386,7 @@ const PRESENTATIONAL_TYPES = [
   ConfigEntryType.ACTION,
 ];
 
+// terminal steps: closing them must not abort (the flow already ended server-side)
 const isTerminal = computed(
   () =>
     step.value?.type === FlowStepType.FINISH ||
--- a/src/components/SetupFlowDialog.vue
+++ b/src/components/SetupFlowDialog.vue
@@ -312,6 +312,7 @@ import {
 import { Progress } from "@/components/ui/progress";
 import { Spinner } from "@/components/ui/spinner";
 import { serverNow } from "@/composables/useServerTime";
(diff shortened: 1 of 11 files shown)


marcelveldt:
On the two suppressed comments about `mergeActionEntries` using `??` — it's the other way around.

`ConfigEntry.value` on the server defaults to `None` and is serialized (no `serialize="omit"` on that field, unlike e.g. `category_translation_key`), so an action response arrives with `value: null` for every entry it didn't set, never `undefined`. Verified against the model:

```
ConfigEntry(key='account', type=STRING, label='Account').to_dict()
-> value present: True, value: None
```

So treating only `undefined` as "unset" would make every action response overwrite the whole form with nulls again — exactly the bug this PR fixes. `??` is correct here precisely because the server's "unset" is null.

On the clearing-to-null case: no action in the tree does that. The only `handle_config_action` that sets a value at all is the fastmcp connect wizard, and that's a URL entry with a real value (and URL entries are stripped by `openActionUrlEntries` before the merge anyway). Leaving it as is.

marcelveldt:
Good catch on the immediate_apply half of that — fixed in 5fa7f150.

The save loop was reading the raw action response rather than the merged form state, and since the server sends `value: null` for everything the action didn't set, an immediate_apply action wrote those nulls straight over the stored config. It now saves what the form actually ended up with. Added a test that pins it — without the fix it saves `clear_on_start: null` instead of the real value. Two party provider actions use immediate_apply, so this was reachable.

On the `??` itself I'm keeping it, for the reason in my earlier comment: the server always sends `null` for an unset value (it's `ConfigValueType = None` with no `serialize="omit"`), never `undefined`, so treating only `undefined` as unset would put the whole bug back. Nothing in the tree clears a value to null.

PR #2274, opened 2026-08-05
Add a relative mode and a haptics toggle to the volume slider

quadcom:

The volume slider has one fixed behaviour: tap or drag anywhere on the bar and the
volume jumps to that position. That works well with a mouse on a wide slider. On a
phone, where the bar is a thin strip, landing on a specific level is hard and a
mistimed tap can move the volume a long way.

This adds an alternative alongside it, under **Settings → Frontend → Volume control**:

- **Absolute** (default, unchanged) — the tap or drag position sets the volume.
- **Relative** — the volume adjusts by the *distance* dragged, from where the drag
  started, at half pointer speed. Where you first touch does not matter, so a
  mistimed tap cannot jump the volume and small adjustments are practical. This also
  enables click-and-drag with a mouse.

A second setting turns the volume slider's haptic feedback off, for people who do not
want a pulse on every step.

Both defaults reproduce today's behaviour exactly, so nothing changes unless you go
looking for it.

Stacked on #2523, which this needs so that saving these two settings does not reload
the page for no reason.

- n/a

- n/a

## Changes

- Add a volume slider behaviour setting (absolute / relative) as a per-user
  preference, so it follows you across browsers and devices.
- Add a volume haptic feedback toggle, gating every vibration from the volume
  control in one place.
- In relative mode, measure a touch drag from the point it was recognised as a drag,
  so the travel spent recognising it is not itself a volume change.
- Keep a short gesture a tap in relative mode, so the group popout still opens.
- Skip the settings page reload for these two keys, since the slider reads them live.

## Types of changes

## Checklist

- [x] `pnpm build` passes.

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerVolume.vue
+++ b/src/layouts/default/PlayerOSD/PlayerVolume.vue
@@ -67,9 +67,11 @@
         disabled: isDisabled,
         muted: isMuted,
         'not-powered': player.powered == false,
+        'relative-mode': isRelativeMode,
       }"
       :style="{ width: width }"
       @click="onSliderClick"
+      @mousedown="onMouseDown"
       @pointerdown.capture="onPointerDown"
       @pointermove.capture="onPointerMove"
       @pointerup="onPointerUp"
@@ -96,7 +98,7 @@
         :disabled="isSliderDisabled"
         :min="0"
         :max="100"
-        :step="step"
+        :step="effectiveStep"
         class="volume-slider"
         :class="cn('w-full', props.class)"
         @update:model-value="onSliderUpdate"
@@ -119,6 +121,7 @@
 
 <script setup lang="ts">
 import { Slider } from "@/components/ui/slider";
+import { useUserPreferences } from "@/composables/userPreferences";
 import { getVolumeIconComponent, truncateString } from "@/helpers/utils";
 import { cn } from "@/lib/utils";
 import { api } from "@/plugins/api";
@@ -146,6 +149,7 @@ export interface Props {
   /** Ask the parent to expand inline group controls when the slider is tapped */
   requestExpandOnGroupTap?: boolean;
   width?: string;
+  /** Overrides the user's volume step preference when set */
   step?: number;
   allowWheel?: boolean;
   color?: string;
@@ -160,7 +164,9 @@ const props = withDefaults(defineProps<Props>(), {
   enablePopout: true,
   requestExpandOnGroupTap: false,
   width: "100%",
-  step: 2,
+  // No default: falling through to undefined is what lets the user's
+  // volume_step preference apply (see effectiveStep below)
+  step: undefined,
   allowWheel: false,
   color: "secondary",
   class: "",
@@ -170,6 +176,35 @@ const emit = defineEmits<{
   (e: "update:local-value", value: number): void;
   (e: "toggle-group-expansion"): void;
 }>();
+
+// --- User preferences (Settings -> Frontend -> Volume control) ---
+
+const { getPreference } = useUserPreferences();
+
+const DEFAULT_VOLUME_STEP = 2;
+const MIN_VOLUME_STEP = 1;
+const MAX_VOLUME_STEP = 10;
+
+// "absolute": the drag/tap position sets the volume (upstream default).
+// "relative": the drag distance adjusts the volume from where the drag started.
+const volumeSliderMode = getPreference<string>(
+  "volume_slider_mode",
+  "absolute",
+);
+const isRelativeMode = computed(() => volumeSliderMode.value === "relative");
+
+const hapticsEnabled = getPreference<boolean>("volume_haptics", true);
+
+const volumeStep = getPreference<number>("volume_step", DEFAULT_VOLUME_STEP);
+
+// An explicit prop wins; otherwise follow the preference. Sanitised because
+// preferences are free-form JSON on the server side.
+const effectiveStep = computed(() => {
+  if (props.step) return props.step;
+  const pref = Number(volumeStep.value);
+  if (!Number.isFinite(pref)) return DEFAULT_VOLUME_STEP;
+  return Math.min(Math.max(Math.round(pref), MIN_VOLUME_STEP), MAX_VOLUME_STEP);
+});
 
 // --- Player-aware computed properties ---
 
@@ -412,6 +447,11 @@ let pointerStartX: number | null = null;
 let pointerStartValue = 0;
 let pointerMoved = false;
 
+// Mouse drag tracking (relative mode only)
+const isMouseDragging = ref(false);
+const mouseStartX = ref(0);
+const mouseStartValue = ref(0);
+
 let sliderUpdateDebounceTimeout: ReturnType<typeof setTimeout> | null = null;
 const SLIDER_UPDATE_DEBOUNCE_MS = 100;
 const POINTER_DRAG_THRESHOLD = 4;
@@ -419,13 +459,15 @@ const POINTER_DRAG_THRESHOLD = 4;
 onUnmounted(() => {
   if (dragEndTimeout) clearTimeout(dragEndTimeout);
   if (sliderUpdateDebounceTimeout) clearTimeout(sliderUpdateDebounceTimeout);
+  document.removeEventListener("mousemove", onMouseMove);
+  document.removeEventListener("mouseup", onMouseUp);
 });
 
 const clamp = (value: number, min: number, max: number) =>
   Math.min(Math.max(value, min), max);
 
 const roundToStep = (value: number) =>
-  Math.round(value / props.step) * props.step;
+  Math.round(value / effectiveStep.value) * effectiveStep.value;
 
 // --- Drag helpers ---
 
@@ -486,11 +528,13 @@ const onMuteToggle = () => {
 // --- Helpers ---
 
 const vibrate = (duration: number = 10) => {
+  if (!hapticsEnabled.value) return;
   if (store.isTouchscreen && "vibrate" in navigator && navigator.vibrate) {
     navigator.vibrate(duration);
   }
 };
 
+// Absolute mode: the pointer position maps directly onto the 0-100 range.
 const getPercentageFromX = (clientX: number): number => {
   if (!sliderContainerRef.value) return displayValue.value;
 
@@ -500,6 +544,25 @@ const getPercentageFromX = (clientX: number): number => {
 
   return clamp(roundToStep(percentage), 0, 100);
 };
+
+// Relative mode: the drag *distance* is applied to the value the drag started
+// from, at reduced speed so small adjustments are easier to land.
+const RELATIVE_DRAG_SENSITIVITY = 0.5;
+
+const getValueFromDelta = (startValue: number, deltaX: number): number => {
+  if (!sliderContainerRef.value) return displayValue.value;
+
+  const rect = sliderContainerRef.value.getBoundingClientRect();
+  const deltaPercent = (deltaX / rect.width) * 100 * RELATIVE_DRAG_SENSITIVITY;
+
+  return clamp(roundToStep(startValue + deltaPercent), 0, 100);
+};
+
+// Resolves a touch drag position through whichever mode is active.
+const getDragValue = (clientX: number): number =>
+  isRelativeMode.value
+    ? getValueFromDelta(touchStartValue.value, clientX - touchStartX.value)
+    : getPercentageFromX(clientX);
 
 // --- Touch handlers ---
 
@@ -560,7 +623,7 @@ const onTouchMove = (event: TouchEvent) => {
   if (isDrag.value) {
     event.preventDefault();
 
-    const newValue = getPercentageFromX(touch.clientX);
+    const newValue = getDragValue(touch.clientX);
     const valueChanged = newValue !== displayValue.value;
 
     displayValue.value = newValue;
@@ -607,13 +670,9 @@ const onTouchEnd = (event: TouchEvent) => {
       }
     }
   } else if (!isSliderDisabled.value) {
-    // Drag end: send the final absolute value to the server
+    // Drag end: send the final value to the server
     const touch = event.changedTouches[0];
-    const finalValue = clamp(
-      roundToStep(getPercentageFromX(touch.clientX)),
-      0,
-      100,
-    );
+    const finalValue = getDragValue(touch.clientX);
     displayValue.value = finalValue;
     emit("update:local-value", finalValue);
     setVolume(finalValue);
@@ -717,6 +776,73 @@ const onTouchCancel = () => {
   isTouching.value = false;
 };
 
+// --- Mouse drag handlers (relative mode only) ---
+//
+// In absolute mode the Slider owns mouse input and onPointerDown/onSliderClick
+// own the group tap, so every handler here bails out early. In relative mode
+// the slider has pointer-events disabled, which also makes the pointer
+// handlers inert (their targetsSlider check can never match).
+
+const MOUSE_DRAG_THRESHOLD_PX = 5;
+
+const onMouseMove = (event: MouseEvent) => {
+  if (!isMouseDragging.value || isSliderDisabled.value) return;
+
+  const newValue = getValueFromDelta(
+    mouseStartValue.value,
+    event.clientX - mouseStartX.value,
+  );
+  displayValue.value = newValue;
+  emit("update:local-value", newValue);
+};
+
+const onMouseUp = (event: MouseEvent) => {
+  document.removeEventListener("mousemove", onMouseMove);
+  document.removeEventListener("mouseup", onMouseUp);
+
+  if (!isMouseDragging.value) return;
+  isMouseDragging.value = false;
+
+  const deltaX = event.clientX - mouseStartX.value;
+
+  if (Math.abs(deltaX) < MOUSE_DRAG_THRESHOLD_PX) {
+    // Treated as a click rather than a drag: run the group action. This also
+    // stamps lastPopoutToggleTime, so the click event that follows this
+    // mouseup is a no-op in onSliderClick.
+    if (handlesGroupTap.value) {
+      handleGroupTap();
+    }
+  } else {
+    const finalValue = getValueFromDelta(mouseStartValue.value, deltaX);
+    displayValue.value = finalValue;
+    emit("update:local-value", finalValue);
+    setVolume(finalValue);
+  }
+
+  stopDragging();
+};
+
+const onMouseDown = (event: MouseEvent) => {
+  if (!isRelativeMode.value) return;
+  if (isSliderDisabled.value) return;
+  // Leave the mute button and the volume readout to their own handlers
+  if (
+    (event.target as HTMLElement).closest(".volume-prepend, .volume-append")
+  ) {
+    return;
+  }
+
+  isMouseDragging.value = true;
+  mouseStartX.value = event.clientX;
+  mouseStartValue.value = displayValue.value;
+  startDragging();
+  // Suppress text selection while dragging
+  event.preventDefault();
+
+  document.addEventListener("mousemove", onMouseMove);
+  document.addEventListener("mouseup", onMouseUp);
+};
+
 const onWheel = (event: WheelEvent) => {
   if (!props.allowWheel || isSliderDisabled.value) return;
 
@@ -757,6 +883,8 @@ const onSliderUpdate = (values: number[] | undefined) => {
 
 // Desktop clicks use the same group action as touch taps.
 const onSliderClick = () => {
+  // In relative mode the mouseup handler owns this, so don't run it twice
+  if (isRelativeMode.value) return;
   if (!handlesGroupTap.value || isTouching.value || isDragging.value) return;
   if (Date.now() - lastPopoutToggleTime < 500) return;
   handleGroupTap();
@@ -860,11 +988,24 @@ watch(
 
 /* --- Group volume popout styles are in the unscoped style block below --- */
 
+/* Absolute mode on touch devices: the container owns the gesture, so the
+   slider itself must not swallow pointer events. Mouse input still reaches
+   the slider, which keeps click-to-position and keyboard control working. */
 @media (pointer: coarse) {
   .volume-slider,
   .volume-slider :deep(*) {
     pointer-events: none;
   }
+}
+
+/* Relative mode: the container owns pointer interaction on every device. */
+.player-volume-container.relative-mode {
+  cursor: ew-resize;
+}
+
+.player-volume-container.relative-mode .volume-slider,
+.player-volume-container.relative-mode .volume-slider :deep(*) {
+  pointer-events: none;
 }
 </style>
 
--- a/src/layouts/default/PlayerOSD/PlayerVolume.vue
+++ b/src/layouts/default/PlayerOSD/PlayerVolume.vue
@@ -549,6 +549,8 @@ let lastSentValue: number | null = null;
 // Relative mode: the drag *distance* is applied to the value the drag started
 // from, at reduced speed so small adjustments are easier to land.
 const RELATIVE_DRAG_SENSITIVITY = 0.5;
+// Below this a relative-mode mouse gesture counts as a click, not a drag
+const MOUSE_DRAG_THRESHOLD_PX = 5;
 
 onUnmounted(() => {
   if (dragEndTimeout) clearTimeout(dragEndTimeout);
@@ -1059,10 +1061,15 @@ const onTouchCancel = () => {
 // the slider has pointer-events disabled, which also makes the pointer
 // handlers inert (their targetsSlider check can never match).
 
-const MOUSE_DRAG_THRESHOLD_PX = 5;
-
 const onMouseMove = (event: MouseEvent) => {
-  if (!isMouseDragging.value || isSliderDisabled.value) return;
+  if (!isMouseDragging.value) return;
+  // A mouseup swallowed by a context menu or a window switch would otherwise
+  // leave the drag latched, and with it the block on server volume updates.
+  if (event.buttons === 0) {
+    onMouseUp(event);
+    return;
+  }
+  if (isSliderDisabled.value) return;
 
   const newValue = getValueFromDelta(
     mouseStartValue.value,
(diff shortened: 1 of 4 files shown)


marcelveldt:
Some issues with this PR:

- volume step size is defined by the backend (we send volume up/down commands)
- On mobile we use tap on the slider to do volume up/down adjustments so your claim that a tap on mobile immediately moves the volume is incorrect
- Is there a reason to make haptic volume feedback configurable ?

quadcom:
I saw the discussions about people complaining about the volume control. Some people want the buttons back; others don't. The reason those want the buttons back is the difficulty, if not sheer impossibility, of lowering the volume by tapping once the volume slider reaches the extreme ends. The smaller the space is, the harder it is to hit the level you actually want. I myself have a lot of difficulty turning the volume down to, let say, 6. The second you tap the slider and slide ever so slightly, it jumps to 12, and you have to slide it back down. But removing your finger from the slider can cause it to jump back to 10. And when your finger is removed, the volume jumps to that level. If it's late at night and all you want to do is turn it down slightly, and the slider is already at 8, tapping to lower it doesn't work because the hit target is too small.

As for the haptics, personally, I just find it annoying. But making it optional is not forcing anyone to have it or not.  The volume steps config works in the code I put together. I've confirmed it myself.

I came across a fork of the Mushroom cards a long time ago. That added functionality to the sliders, which, once I tried it, I was sold. The ease with which lights could be controlled made the user interface feel way more polished. When I ran into the difficulty controlling MA's volume, I immediately thought of this approach.

I posted comments trying to appease everyone. The slider is still in place as normal, but it also gives the others the fine control they want. All of the effort I put into creating the functionality of the relational sliders was pretty much ignored.

It occurred to me that replacing the volume control outright probably wouldn't be well received, so I considered making this an option people could enable if they want it. That approach might be more agreeable as the disturbance wouldn't be felt by anyone. To make this even easier to incorporate, I went the extra mile and created the changes and confirmed all the testing, so there's no extra work required for the maintainers!

I've tried going down the route of creating my own fork and applying the mods to my own MA build. But from what I have learned, it would be pretty hard for me to maintain. Let me be frank, I'm not a developer at all. If it weren't for Claude, I wouldn't be doing the kinds of things that I now find myself able to do. I wish I were more skilled in development, but at (almost) 54, the new-trick/old-dog thing kicks in.l.

I don't mean to step on anyone's toes or cause disruption. But I truly believe this functionality would be well received by the community and would only enhance MA for its users. I do hope you consider, at the very least, trying it out and seeing for yourself if this makes MA just the slightest bit better.

I might add that the tap to adjust volume has stopped working recently, just for increasing the volume, so far. At least on Android/FF and the HA App, so sliding is the only option for now. I'll write up the bug report.

https://private-user-images.githubusercontent.com/7085931/613291660-43245da9-801e-45fb-ac22-422ea201b64f.mp4?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODU5NTc0MDksIm5iZiI6MTc4NTk1NzEwOSwicGF0aCI6Ii83MDg1OTMxLzYxMzI5MTY2MC00MzI0NWRhOS04MDFlLTQ1ZmItYWMyMi00MjJlYTIwMWI2NGYubXA0P1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI2MDgwNSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNjA4MDVUMTkxMTQ5WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MTcyOGVkMDQ2MmJkZTZiNTljZTBlNTdhMDBkYTRhOWRiNTA3NWQ4ZmZlZWY3ZWQ2Nzc4N2IwOGMwMWZlODc0MyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmcmVzcG9uc2UtY29udGVudC10eXBlPXZpZGVvJTJGbXA0In0.48A20Cyfshtxgm2xOgMZr0M2aKiu-Wa9gxXhIB-UqMI

marcelveldt:
- The haptic feedback never worked on iOS (for me at least) so I never noticed it. If it's indeed annoying, it makes sense to add a user pref
- Step size should be serverside config so it will work for any client consistent - currently this is a bit smart where closer to the edged, it uses step size 1 but further in the middle a bigger step size, but we can make that configurable

I get the point about the relative volume adjustments but I think it would be wise if we can do a bit of a poll on discord what the default should be. This is how it -should- work today:

On mobile:
- Tap at the edges of the volume slider does volume up/down command (left side is down, right side up)
- This tap approach on mobile is there so you never accidentally blast a player to max volume when you mis click.
- Grab the handle explicitly and drag it: adjust the level directly

On desktop:
- Grabbing the handle or clicking at a position within the slider adjusts it immediately, reasoning is that on desktop a mouse control is more precise.

With a small note to both that when playing on a group(ed) player, a tap on the slider will unfold all the volume sliders of that playergroup.


I like to also propose one alternative, one I have seen in the Sonos app recently:
A tap on the volume slider on mobile reveals two large volume up and down buttons.
I'll have to make a screenrecording if it maybe but I thought this could also be potentially very nice.

@MarvinSchenkel @stvncode @Erioldoesdesign what do you think ?
I know there are some complaints heer and there about the volume slider but I also know that the previous approach also did not satisfy everyone. Maybe this is one of the cases where we really can not find the common middle ground and we need to make it configurable ?

Can we put up a poll in the community of what people want to see (by default).
I must say that I do kind of like the relative adjustments approach this PR is proposing, the only thing I'm not sure of is if people would understand this directly (but you can say the same about the tapping on left/right edges).

quadcom:
_> the only thing I'm not sure of is if people would understand this directly (but you can say the same about the tapping on left/right edges)._

This is exactly why I went with the optional aspect. It can be left as is for the default and people can try the relative option. If they don't like it, they can switch back. No harm no foul.

And this PR does include the single tap to show the individual member volume controls still.


On a side note, had MA supported a plugin ecosystem, this could have just been created as one. It would also allow the broader community the ability to tailor their own MA to how they use it.

marcelveldt:
> On a side note, had MA supported a plugin ecosystem, this could have just been created as one.

We have a plugin system, just not for the frontend we ship with it. But anyone is free to create their own client as we have  an extended and fully documented API. In fact there are already multiple (mobile) clients available for MA.

> This is exactly why I went with the optional aspect. It can be left as is for the default and people can try the relative option. If they don't like it, they can switch back.

Sure but in general we try to avoid options (because flexibility = complexity) and just try to build something that works great for the majority and only offer this kind of choices if no clear consensus can be found. In the end every toggle/setting you add, means additional maintenance burden and added complexity for a user.

The volume slider may be one of these cases where a user preference setting is needed, because I really think the opinions are split here. It would be a good way to also fold in those large plus/min buttons as an extra option in the list.

The only thing that I think needs to be dropped is the step size - as explained earlier we should create a backend setting for that (and we can even expose that here for easier access) - unless you argue that the step size in this case is bound to the relative slider you introduced.

quadcom:
I just wanted to note that I did push an edit to this PR just now. The change is a real-time volume adjustment while dragging the slider in relative mode. Rather than setting the volume on release.

If this is a dealbreaker, I do have notes to make a quick reversion.

On another note, I did make another edit to my own fork to accommodate the server-side volume_step configuration. Here are the notes on those premeptive changes. I hope this move make this easier to adopt. If this is acceptable I can merge those changes in to this PR.



_Re the step size — happy to hand it to the backend. Two things I found while wiring it up:

The frontend volume_step and the server's step aren't currently the same knob. Tap-at-edges and the scroll wheel already call volume_up/volume_down, so they've always used the server ladder; the preference only ever affected drag granularity. The settings description overclaimed and I've fixed it.

Proposed contract so the frontend can just adopt it: core config on the players controller, key volume_step, INTEGER, range 1–10, default 0 meaning "adaptive" so today's 1/2/3 ladder in cmd_volume_up stays the behaviour for anyone who never opens the setting. Once it exists it should drive cmd_volume_up/cmd_volume_down and the group equivalents — that's what actually makes it consistent across clients, which a frontend preference never could.

I've built the read side on my fork already: it reads that key if present and falls back otherwise, so it's a no-op against current servers. Happy to fold it into this PR once you're happy with the key and shape._

marcelveldt:
yes that sounds good - feel free to open that server PR and I will have a look

marcelveldt:
> The change is a real-time volume adjustment while dragging the slider in relative mode

Just make sure to debounce this a bit because players can very easily be overloaded if we send too much commands at once. There should be an existing debouncer already in the (volume) code if I remember correctly

quadcom:
> yes that sounds good - feel free to open that server PR and I will have a look

Server PR submitted - https://github.com/music-assistant/server/pull/5571

quadcom:
> > The change is a real-time volume adjustment while dragging the slider in relative mode
> 
> Just make sure to debounce this a bit because players can very easily be overloaded if we send too much commands at once. There should be an existing debouncer already in the (volume) code if I remember correctly

I pushed an update to this PR after I looked into the throttling some more. I tweaked it so that it will have the same effect as the current volume control.

marcelveldt:
Can you update this PR and the description now that the server changes are merged ?

quadcom:
> Can you update this PR and the description now that the server changes are merged ?

Done. I hope the edits are what you were looking for. Still learning my way around things.

Thanks!

quadcom:
I thought I'd test the server DEV branch with #5571 against this front-end PR to ensure the volume_steps preference works as expected. Turns out it exposed a bug that would have caused a headache. The saved preference was not being tracked in the front-end UI under the Settings > User Interface > Volume Control section.

All resolved with the latest commit, and hands-on tests confirmed the fix.

OzGav:
A conflict needs resolving

quadcom:
> A conflict needs resolving

I'll wrap that up later tonight.

marcelveldt (commented):
Thanks for this, the relative drag mode is a nice idea and that part reads well.

Content wise I'm fine with making the slider behaviour configurable. What I do want out of this PR is the whole volume step size part, see the comments below.

Short version: the server side `volume_step` is for the volume up/down commands, so tapping the ends of the slider, the scroll wheel, a button in HA. The step used while *dragging* the slider is something else entirely. Coupling them means someone who sets volume_step to 5 for their remote suddenly gets a slider that only moves in 5% blocks here, which is not what they asked for. And I don't think anyone wants the drag step configurable at all, just pick a sane default (1 or 2) and hardcode it. Two settings that are both called "volume step size" but mean different things is exactly the confusion I want to avoid, it confused me too while reading this PR earlier (and that what got me set you up to create the server side PR).

On top of that, the server setting already shows up automatically under Settings -> Core -> Players. It's a normal core config entry and we render those generically, so there is nothing to build in the frontend for it.

The rest of the comments are things I'd want fixed regardless. Also the branch is quite far behind main by now (~170 commits), can you rebase?

src/composables/useVolumeStep.ts
```diff
@@ -0,0 +1,202 @@
+import { computed, ref, watch, type ComputedRef } from "vue";
+import {
+  setUserPreference,
+  useUserPreferences,
+} from "@/composables/userPreferences";
+import { api, ConnectionState } from "@/plugins/api";
+import { EventType } from "@/plugins/api/interfaces";
+
+/**
```
marcelveldt: Can we drop this whole file? See my main comment: the server step and the drag step are two different things and I don't want them coupled.

Just hardcode the drag step and be done with it. KISS.

src/composables/useVolumeStep.ts
```diff
@@ -0,0 +1,202 @@
...
+};
+
+// There is no CORE_CONFIG_UPDATED event to listen for, so a reconnect and our
+// own writes are the only invalidation points. Another client changing the
+// setting is picked up on the next reload.
+let subscribed = false;
+const subscribeOnce = (): void => {
+  if (subscribed) return;
+  subscribed = true;
+  api.subscribe(EventType.CONNECTED, () => {
```
marcelveldt: Since you went to the trouble of writing the comment above `canReadServerValue`: this handler never does anything. `handleServerInfoMessage` sets state to CONNECTED right before it fires this event, and `canReadServerValue()` only accepts AUTHENTICATED/INITIALIZED, so `ensureLoaded()` here always bails out immediately.

src/composables/useVolumeStep.ts
```diff
@@ -0,0 +1,202 @@
...
+    loaded.value = false;
+    inFlight = null;
+    ensureLoaded();
+  });
+  // The first caller is usually a component mounting while the connection is
+  // still coming up, and a reconnect clears `loaded` before the api is
+  // readable again. Both cases would otherwise leave the value unread until
+  // something happened to call in again, so pick the read up off the state.
+  watch(
```
marcelveldt: And this watch is created from the first PlayerVolume's `setup()`, so vue binds it to that component's effect scope and stops it as soon as that slider unmounts. `subscribed` stays true so it never comes back.

We have `effectScope(true)` in `useActiveTrackWaveform.ts` for exactly this. Not asking you to fix it here since I want the file gone anyway, just noting it so it doesn't reappear somewhere else.

src/composables/useVolumeStep.ts
```diff
@@ -0,0 +1,202 @@
...
+export const serverVolumeStep: ComputedRef<number | undefined> = computed(
+  () => serverValue.value,
+);
+
+/** Whether the server owns this setting, i.e. writes should go there. */
+export const volumeStepIsServerOwned: ComputedRef<boolean> = computed(
+  () => serverValue.value !== undefined,
+);
+
+// Reading the core config needs an authenticated connection. CONNECTED is only
```
marcelveldt: Small thing for future PRs: comments should describe what the code does now, not what an earlier version of the branch did wrong. Same in the new test files. That story belongs in the PR description.

src/layouts/default/PlayerOSD/PlayerVolume.vue
```diff
@@ -160,7 +165,9 @@ const props = withDefaults(defineProps<Props>(), {
   enablePopout: true,
   requestExpandOnGroupTap: false,
   width: "100%",
-  step: 2,
+  // No default: falling through to undefined is what lets the configured
+  // step size apply (see effectiveStep below)
+  step: undefined,
```
marcelveldt: Please keep `step: 2` here.

src/layouts/default/PlayerOSD/PlayerVolume.vue
```diff
@@ -560,14 +692,15 @@ const onTouchMove = (event: TouchEvent) => {
...
-    const newValue = getPercentageFromX(touch.clientX);
+    const newValue = getDragValue(touch.clientX);
     const valueChanged = newValue !== displayValue.value;
 
     displayValue.value = newValue;
     emit("update:local-value", newValue);
 
     if (valueChanged) {
       vibrate(5);
+      sendVolumeLive(newValue);
```
marcelveldt, line 890: This isn't gated on relative mode, so absolute mode on touch goes from one volume command on release to one every 100ms for the whole drag. That's a change for every existing user, not something new users opt into, so "every default reproduces current behaviour exactly" in the description isn't right.

I'm not against live volume while dragging, desktop already does roughly that through the debounce. But then let's be explicit about it. Preferably split it into its own PR, this one is already doing three things.

src/plugins/api/index.ts
```diff
@@ -2095,9 +2095,16 @@ export class MusicAssistantApi {
   public async getPlayerConfigValue(
     player_id: string,
     key: string,
-  ): Promise<PlayerConfig> {
+    defaultValue?: ConfigValueType,
```
marcelveldt: The `defaultValue` params aren't used anywhere. `config/players/get_value` has no callers at all and the two callers of `config/core/get_value` don't pass it. Please leave the api untouched in this PR.

The return type fix (`PlayerConfig` -> `ConfigValueType`) is a real bug though, that one is worth a separate small PR.

src/views/settings/FrontendConfig.vue
```diff
@@ -172,6 +181,46 @@ onMounted(() => {
...
+        { title: "relative", value: "relative" },
+      ],
+      multi_value: false,
+      category: "volume_control",
+      value:
+        (store.currentUser?.preferences?.volume_slider_mode as string) ||
+        "absolute",
+    },
+    {
+      key: VOLUME_STEP_KEY,
```
marcelveldt: This entry can go. The server setting is already reachable under Settings -> Core -> Players, mirroring it here just gives us two places that can drift apart.

src/views/settings/FrontendConfig.vue
```diff
@@ -214,6 +263,21 @@ onMounted(() => {
     }
   }
   config.value = configEntries;
+
+  // The step size is owned by the server when it provides one, and reading it
+  // is a round trip. Render the form first on the locally known value, then
+  // correct it if the server disagrees, so a slow or silent server cannot leave
+  // the settings page blank.
+  await ensureVolumeStepLoaded();
```
marcelveldt: And this block goes with it.

src/views/settings/FrontendConfig.vue
```diff
@@ -230,7 +294,12 @@ const saveValues = async function (values: Record<string, ConfigValueType>) {
       // Action-only entries carry no persistable value
       if (entry.type === ConfigEntryType.ACTION) continue;
 
-      if (DEVICE_SETTING_KEYS.has(key)) {
+      if (key === VOLUME_STEP_KEY) {
```
marcelveldt: Also worth knowing before you remove it: on a stock server `volume_step` is 0, which `sanitizeVolumeStep` rejects, so `volumeStepIsServerOwned` is false and this writes to the user preference instead of the server. It only starts writing server side after someone set a non-zero value on the core players page.

So for pretty much everyone the description string ("Stored on the server so every client agrees on it") isn't true, and the same slider silently changes where it saves to depending on state the user can't see.

src/views/settings/FrontendConfig.vue
```diff
@@ -230,7 +294,12 @@ const saveValues = async function (values: Record<string, ConfigValueType>) {
...
+        // Server-owned when the server provides it, so it routes through its own
+        // setter. Deliberately does not set hasPerUserChanges: consumers read it
+        // reactively, so there is nothing a page reload would fix.
+        await setVolumeStep(values[key]);
+      } else if (DEVICE_SETTING_KEYS.has(key)) {
```
marcelveldt: One thing on the other two settings: they fall through to the generic branch below, which sets `hasPerUserChanges` and triggers a full `window.location.reload()`. Both are read reactively in PlayerVolume, so that reload does nothing useful. Same reasoning as your comment right above, can you make them skip it too?

src/translations/en.json
```diff
@@ -528,6 +529,22 @@
...
         "add_group_player_desc": "Create a permanent player group with {0} players. When you play to this Player Group, the member players will be automatically synchronized.",
+        "volume_slider_mode": {
+            "label": "Volume slider behaviour",
+            "description": "Absolute: tapping or dragging the bar sets the volume to that position. Relative: dragging adjusts up or down from the current volume at half speed, which makes small changes easier to land.",
+            "options": {
+                "absolute": "Absolute (jump to position)",
+                "relative": "Relative (drag to adjust)"
+            }
+        },
+        "volume_step": {
```
marcelveldt: These strings go too.

tests/layouts/PlayerVolume.test.ts
```diff
@@ -15,9 +15,15 @@ const { getVolumeIconComponent } = vi.hoisted(() => ({
 }));
 
 vi.mock("@/plugins/api", async () => {
-  const { reactive } = await vi.importActual<typeof import("vue")>("vue");
+  const { reactive, ref } = await vi.importActual<typeof import("vue")>("vue");
+  const ConnectionState = { CONNECTED: 5 };
```
marcelveldt: `ConnectionState = { CONNECTED: 5 }` means `canReadServerValue()` compares 5 against undefined for both AUTHENTICATED and INITIALIZED, so `getCoreConfig` is never called here. The test passes, but not for the reason the comment below claims. Same in PlayerVolumeRelative.test.ts.

Moot once useVolumeStep is gone, but a mock that quietly disables the thing under test is worth avoiding in general.

tests/layouts/PlayerVolumeRelative.test.ts
```diff
@@ -0,0 +1,293 @@
...
+import { api } from "@/plugins/api";
+import {
+  IdentifierType,
+  PlaybackState,
+  type Player,
+  PlayerFeature,
+  PlayerType,
+} from "@/plugins/api/interfaces";
+import { store } from "@/plugins/store";
+import { mount } from "@vue/test-utils";
```
marcelveldt: Please add `enableAutoUnmount(afterEach)` here. Nothing unmounts and the mouse tests attach document level listeners, so leaked instances keep reacting to later tests' events. It passes today, but by luck.

Would also be good to cover the two behaviour changes to the existing absolute path (live sends during a touch drag, and the new commit-on-cancel in onTouchCancel).

src/layouts/default/PlayerOSD/PlayerVolume.vue
marcelveldt: Also if I remember correctly we have had issues on the past with mobile and enabling the live dragging so that alone is worth a separate PR so it can first be tested for side effects.

quadcom:
Thanks @marcelveldt , that all made sense — the PR is now narrowed to the two settings and rebased onto current `main`. Description updated to match.

**Volume step size is gone entirely**, along with `useVolumeStep.ts` and its tests. Your framing was the part I had wrong: the server's `volume_step` drives up/down commands and the drag step is a different quantity, so coupling them was the bug, not a missing feature. Drag step is back to the hardcoded `step: 2` default and `step: 2` is restored on the prop. Nothing in the frontend for the server setting — confirmed it renders generically under Settings → Core → Players against 2.10.0b14, with no duplicate under Frontend.

Also confirmed your point about it writing to the wrong place: on a stock server `volume_step` is `0`, `sanitizeVolumeStep` rejected it, so `volumeStepIsServerOwned` was false and the field silently wrote to the user preference while its description claimed the server. That was true for essentially everyone.

**Live volume while dragging is out of this PR** and now lives on its own branch, so "every default reproduces current behaviour exactly" is true again. You were right to push back on it — it wasn't gated on relative mode, so absolute-mode touch drags changed for every existing user. I've since tested it against 2.10.0b14 from a phone: a 655ms drag sends 7 commands at ~100ms where `main` sends 1 on release, and I found that on a group it needed a window that scales with member count, since `set_group_volume` fans out in parallel and every member's state change is broadcast to every client. Given your note about past mobile issues with live dragging, I'd rather it land separately anyway. It's now #2470, based on the same `main` and independent of this one, with the measurements in the description — review it on whatever timeline suits, this PR doesn't depend on it.

**`api/index.ts` is untouched.** You were right that the `default` arguments had no callers. The return type fix turned out to have already landed on `main` in the meantime, so there's nothing left for me to send as a follow-up.

**The reload skip needed more than exempting the two keys.** `EditConfig` submits every entry on save, not just edited ones, so `theme`, `language` and `show_waveform` were re-saved every time and always set `hasPerUserChanges` — the reload fired regardless. Saving now compares each value against a snapshot of what the form was built from. Verified both directions: saving a volume setting doesn't reload, changing the theme still does.

Two smaller ones: `enableAutoUnmount(afterEach)` added to `PlayerVolumeRelative.test.ts`, and the `PlayerVolume.test.ts` mock that quietly disabled the path its comment claimed to exercise is gone — that file is now identical to `main`. Point about comments describing current behaviour rather than the branch's history taken; that story is in the description now.

Both notes on the deleted composable are noted for future reference rather than fixed here: the `CONNECTED` handler that could never fire, and the module-level `watch` bound to the first mounting component's effect scope. `effectScope(true)` in `useActiveTrackWaveform.ts` is the pattern if that shape comes up again.

Testing: `vue-tsc`, `oxlint`, `eslint` and `prettier` clean, 1953 tests across 207 files. Relative mode exercised on a desktop browser and on an Android phone against 2.10.0b14 — details in the description.

src/layouts/default/PlayerOSD/PlayerVolume.vue
```diff
@@ -913,6 +964,86 @@ const onTouchCancel = () => {
...
+  }
+
+  stopDragging();
+};
+
+const onMouseDown = (event: MouseEvent) => {
+  if (!isRelativeMode.value) return;
+  // a muted slider still answers a tap by opening the group's own controls,
+  // which is how you get to a child slider to unmute it
+  if (isSliderDisabled.value && !handlesGroupTap.value) return;
```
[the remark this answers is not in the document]
marcelveldt: Agreed, fixed - onMouseDown now bails on anything but the primary button. Worth noting it also interacted badly with the lost-release handling: a right click latched the drag and blocked server updates until the mouse happened to move again.

marcelveldt:
Picked this up and finished it off, rather than going another round.

All the earlier points are addressed - `useVolumeStep` and the `volume_step` entry are gone along with their strings, `api/index.ts` is back to untouched, `step: 2` is kept, the api mock that disabled its own test path is gone, `enableAutoUnmount` is in, and the two settings skip the reload. Live volume while dragging was split out to #2470, which is where your "worth a separate PR so it can be tested for side effects" note landed.

Rebased on current main and stacked on #2523, which lifts the settings reload fix out of here so it can be reviewed on its own. This PR is now just the two volume settings.

Also fixed four things in the mouse path while going through it:

- a tap on a touchscreen ran the group action twice, because the browser replays it as compatibility mouse events after touchend had already handled it
- a muted group could not be opened with the mouse, unlike every other path
- a lost mouseup latched the drag and left that slider cut off from server updates for the rest of the session
- a right click started a drag

Each one has a test that fails without the fix. Resolving the outdated threads.

marcelveldt:
Rebased onto current main now that #2470 and #2523 are in. The reload fix that used to live here is in #2523, so this PR is just the two volume settings plus the exemption that stops them triggering a pointless page reload.

Worth noting the two volume PRs compose the way you would want: the live send from #2470 now takes the mode-aware value, so dragging updates as you go in relative mode too, without either change knowing about the other.

I pushed some final tweaks to the code so you don't have to wait longer going back and forth to get this merged.

marcelveldt (approved):
Nice work @quadcom !

PR #2279, opened 2026-08-05
Fix the player progress bar standing still or starting at zero

marcelveldt:
Two problems with the position shown by the player's progress bar and time counter:

- Some players report their playback position at player level only, without a per-track position — a Home Assistant media_player that exposes a `media_position` but no track title, for example. For those the progress bar sat still and only jumped forward whenever the server sent an update, instead of running smoothly.
- Opening the fullscreen player or the Now Playing dashboard on a paused track showed 0:00 instead of the position it was paused at, until playback resumed.

- Animate the progress bar whenever the position it shows is playing, whatever reports it
- Show that position as soon as the progress bar appears, instead of waiting for it to change
- Add tests for the progress bar, covering all three position sources


marcelveldt:
@copilot review

marcelveldt:
@copilot review

marcelveldt:
@copilot review

PR #2293, opened 2026-08-06
Player card no longer uses its own copy of the queue lookup

marcelveldt:
The player card worked out which queue it belongs to with its own copy of logic that already lives in a shared helper. The copy was missing one check, so during the moment a player switches between Music Assistant and an external source (line-in, Spotify Connect) the card could briefly act on the wrong queue.

The "queue ended" and "queue empty" hints in the full screen player already use the shared helper, so the card now matches them.

- Use the shared `resolvePlayerQueue` helper in `PlayerCard.vue` instead of a local copy.

### DIFF
--- a/src/components/PlayerCard.vue
+++ b/src/components/PlayerCard.vue
@@ -218,6 +218,7 @@ import {
   isBuiltinPlayer,
 } from "@/helpers/utils";
 import api from "@/plugins/api";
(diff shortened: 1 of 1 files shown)


marcelveldt:
@copilot review

marcelveldt:
Reworked: instead of documenting the difference, the card now just uses the shared `resolvePlayerQueue` helper. Please re-review.

@copilot review

PR #2298, opened 2026-08-06
Keep the app initialization tests independent of their order

marcelveldt:
`tests/App.test.ts` only passed in the order its tests happen to be declared in, so its result could flip whenever a test was moved, inserted or filtered. `vi.clearAllMocks()` clears call history but leaves mock implementations in place, so a return value or rejection installed by one test kept applying to every test after it — once the failed-locale-push test had made `api.setLocale` reject, the test that expects a state refresh after a locale change no longer got one. Run in a shuffled order, the file failed 7 out of 10 times.

- Reset mock implementations between tests, not just call history
- Install every mock default from `beforeEach`, so that is the only place a mock's behavior comes from
- No app code changes


marcelveldt:
@copilot resolve the merge conflicts in this pull request

marcelveldt:
@copilot resolve the lint errors in this pr

PR #2301, opened 2026-08-06
Avoid duplicate waveform requests after a hot reload during development

marcelveldt:
The shared waveform watcher lives in a module-level effect scope that is not tied to any component. During development, editing the composable leaves the watcher from the replaced module copy running next to the newly loaded one, so every track change fires the waveform request twice until the page is fully reloaded.

- Stop the shared watcher scope when Vite replaces the module
- Accept the update in this module, so it becomes the boundary the disposal is tied to, and pass the update on to the components afterwards

Development-only cleanup, no change in behaviour for users.


marcelveldt:
Closing this one. The single `dispose()` call doesn't actually do anything: Vite keys the disposal by the hot update boundary, which for this composable is the consuming component, so the callback never runs. Making it work needs the composable to accept the update itself and then invalidate, which is more machinery than a dev-only double fetch is worth.

PR #2302, opened 2026-08-06
Use the shared queue resolution in the player card

marcelveldt:
The player card had its own copy of the logic that works out which queue a player is playing, and that copy missed the check that the player's own queue is actually active. While a queue hand-over is still settling the card could briefly act on the wrong queue, which the shared helper already guards against.

- Player card now uses the shared queue resolution helper instead of its own duplicate.


marcelveldt:
Closing as a duplicate of #2293, which is the same change and got there first.

PR #2307, opened 2026-08-07
Clean up unreachable error handling in the logout path

marcelveldt:
The logout path in the auth plugin had a `try/catch` that no longer guarded anything. It used to wrap a dynamic `import()` of the API client, which could fail on a bad chunk load. Since #2294 the client is imported normally, so the only remaining statement inside the `try` is the `api.logout()` call — an async method that can only reject, never throw synchronously, and its rejection is already handled by the attached `.catch()`.

No functional change.

- Removed the redundant outer `try/catch` (and its unused error binding) around the logout command
- Kept the fire-and-forget `.catch()` and the comment explaining why we don't await the response

### DIFF
--- a/src/plugins/auth.ts
+++ b/src/plugins/auth.ts
@@ -304,14 +304,10 @@ export class AuthManager {
 
     // Send logout command to server first (best effort)
     if (this.token) {
-      try {
-        // Send logout command but don't wait for response to avoid race condition
-        api.logout().catch(() => {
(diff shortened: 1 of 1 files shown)


marcelveldt:
Heads-up: the failing Lint check is not from this PR. `main` has been red for several commits on a `vue/html-self-closing` error in `EditorialTimeline.vue` (came in with #2042). Being fixed separately — I'll rebase once it lands.

PR #2308, opened 2026-08-07
Tidy up leftover dynamic imports of the auth plugin

marcelveldt:
The build logged three warnings about dynamic imports that don't actually split anything off into their own chunk. Traced the import graph to see which ones still serve a purpose.

Two were leftovers: `App.vue` was dynamically importing the auth plugin in three places even though it already imports it normally at the top of the same file, and the companion plugin did the same. Neither avoids a circular import, so they're now plain imports.

The other two are kept on purpose — each one is the single edge that keeps two plugins from importing each other directly (auth ↔ companion, and the API client ↔ the router). Both now carry a note saying so, so they don't get "cleaned up" next time.

- Use the already-imported auth plugin in `App.vue` instead of re-importing it three times
- Import the auth plugin normally in the companion plugin
- Document why the auth → companion and API → router imports stay dynamic

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -141,7 +141,6 @@ const handleRemoteAuthenticated = async (credentials: {
   user?: User;
 }) => {
   try {
-    const { authManager } = await import("@/plugins/auth");
     let user = credentials.user;
 
     if (credentials.user && !credentials.token && !credentials.username) {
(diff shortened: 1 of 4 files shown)


marcelveldt:
@copilot review

PR #2322, opened 2026-08-07
Move device type detection into the device helper

marcelveldt:
Follow-up to #2321. Device detection was spread across two places: the store and the breakpoint plugin each parsed the browser's user agent themselves, so the same detection ran twice at startup.

- Moved the device type into `src/helpers/device.ts`, which is now the single place device detection lives
- The breakpoint plugin reuses those results instead of doing its own detection
- The store imports the device type instead of working it out itself
- Added tests for the breakpoint plugin's phone/tablet/mobile handling, which had none

No change in behaviour.

### DIFF
--- a/src/helpers/device.ts
+++ b/src/helpers/device.ts
@@ -4,10 +4,16 @@ export type DeviceType = "desktop" | "phone" | "tablet";
 
 const md = new MobileDetect(window.navigator.userAgent);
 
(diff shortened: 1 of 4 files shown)


marcelveldt:
Copilot flagged that the test mutates `window.innerWidth` without restoring it. Vitest isolates per file here so it can't actually leak, but it's cheap to be tidy - now restored in `afterEach`.

PR #2324, opened 2026-08-07
Add provider filter to Discover recommendation rows

bferd:
Adds a filter to recommendation rows on Discover so you can hide items from specific music providers (e.g. hide Audiobookshelf plays, show only Spotify and local library history).

- Filter icon in a row's header opens a checklist of your configured music providers; hiding one removes its items from that row
- Choice persists per-user, per row
- Filtering happens on the server, so it works without extra per-item lookups and can be enabled for any recommendation row, not just Recently Played

2026-08-07 15_37_29-Filter Recently Played shelf

Requires the matching server-side support, released in music-assistant/models#365 and music-assistant/server#5768.


marcelveldt:
Should we add a provider filter to the recently_added command maybe to make this a bit easier to handle ?

bferd:
> Should we add a provider filter to the recently_added command maybe to make this a bit easier to handle ?

Did you mean recently_added or recently_played. Also, wouldn't that be server side. Have to admit I got into vehicle repair and can't look into it right now. Can you clarify?

bferd:
@marcelveldt

Vehicle repair on hold till parts arrive so I had a chance to do some research into this. I am going to assume that you meant `recently_played` as this is a recently played shelf and not a recently added one - let me know if I'm off there. 

In the DB the play history gets written to the provider column (stored per play) - this runs into the same underlying problem I found on the frontend, where `item.provider` collapses into simply `provider: library`. So in short, a server side provider filter added to `recently_played` would just run into the same wall I did, tracing back to `media_item.provider` being "library" - it would need the same kind of real provider mapping resolution on the back end, not just a filter parameter. Being that this PR intentionally deals with frontend only (especially being a first PR) I don't want to over complicate things. The only reason for server side management would be for reuse somewhere else beyond the scope. 

Do let me know if that hits right.

marcelveldt:
We can add the filter in the backend (which in turn then can resolve library items into providers in an efficient way.
I will add that so this PR can get unblocked.

marcelveldt (approved):
Thanks, @bferd !

bferd:
> Thanks, @bferd !

Wow! Thanks @marcelveldt for getting this to the finish line. Awesome!

PR #2327, opened 2026-08-07
Drop null checks on data the server always sends

marcelveldt:
Follow-up to #2326. A number of reads still checked for null on fields the server always fills. Those checks never do anything, and they make the data look less reliable than it is — which is how the mismatch in #2326 crept in. Each one was cross-checked against the server model before removing it.

- Remove the redundant null checks on always-present fields: item metadata, artists, provider mappings, search results, user filters and setup form entries
- An output protocol always carries its provider domain, including the native one, so drop the null handling and the nine assertions that worked around it (the type comment claiming otherwise was wrong)
- Remove the image cache-buster: it silently stopped working when the server dropped the field it read, so images were never actually cache-busted
- Remove a leftover property passed to the radio "other versions" list that the list stopped accepting in 2023
- Search tests now return the full result set the server sends, instead of a partial stand-in

### DIFF
--- a/src/components/AddManualLink.vue
+++ b/src/components/AddManualLink.vue
@@ -200,7 +200,7 @@ watch(
     if (active != null) store.dialogActive = active;
     if (active && compProps.editItem) {
       // Pre-populate fields with current values
-      const thumbImage = compProps.editItem.metadata?.images?.find(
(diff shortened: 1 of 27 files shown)


marcelveldt:
@copilot review

PR #2334, opened 2026-08-08
Match player types to the null values the server sends

marcelveldt:
The server always sends every field of a player, using `null` for the ones it has no value for — it never leaves a key out. Our types said the opposite (`powered?: boolean`), so code was written to expect `undefined` and `null` slipped past unnoticed.

This corrects the player types to match what actually arrives, so the compiler now points at the places that need a null check.

- `Player`, `PlayerMedia`, `PlayerOption`, `DeviceInfo`, `OutputProtocol` and `MediaItemPalette`: optional fields are now spelled as nullable
- Widened `getMediaImageUrl`, `computeChapterTicks` and the Discord now-playing signal to accept null
- Updated the player test fixtures to build the same shape the server sends

Verified against the `music_assistant_models` dataclasses field by field. Queue, media item, config and provider types have the same problem and follow in separate PRs.

A few fields here (`playback_state`, `has_setup_flow`, `PlayerOption.translation_key`) are marked optional but are never null either — those need to become plain required fields, which is a separate change and is tracked as a follow-up.

### DIFF
--- a/src/helpers/chapters.test.ts
+++ b/src/helpers/chapters.test.ts
@@ -10,6 +10,7 @@ const chapters: MediaItemChapter[] = [
 
 describe("computeChapterTicks", () => {
   it("returns [] when duration is missing or zero", () => {
+    expect(computeChapterTicks(chapters, null)).toEqual([]);
     expect(computeChapterTicks(chapters, undefined)).toEqual([]);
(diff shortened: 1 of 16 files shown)


marcelveldt:
@copilot review

stvncode (approved):

PR #2340, opened 2026-08-08
Add Milkdrop visualizer plugin

jozefKruszynski:
MilkDrop visualizer, powered by the new `milkdrop_visualizer` server plugin
(companion PR: https://github.com/music-assistant/server/pull/5511). Renders Butterchurn behind the fullscreen player, the
now-playing dashboard, and the party screen, replacing the gradient background.

- Full butterchurn preset library (lazy-loaded), favourites, random/fixed preset
  modes, and preset auto-switch on downbeats via MA's beat analysis
- Per-user toggle, blur and opacity controls in the player menu; quality tiers
  and a settings page at `/visualizer`, surfaced as a sidebar entry when the
  plugin is enabled
- Frames arrive over the plugin's WebSocket relay with clock sync, so the
  visuals track what is audibly playing (pause lands on a flatline)

Requires WebGL2; sessions that lack it (including Google Cast receivers, which
have no WebGL support) keep the existing background. Feature is fully gated on
the plugin being installed and enabled.


package.json
```diff
@@ -34,6 +34,8 @@
         "@tanstack/vue-virtual": "3.13.30",
         "@types/qrcode": "^1.5.6",
         "@vueuse/core": "^14.3.0",
+        "butterchurn": "2.6.7",
+        "butterchurn-presets": "2.4.7",
```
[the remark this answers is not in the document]
jozefKruszynski: Inherent to butterchurn-presets, already neutered via core-js: false in pnpm-workspace.yaml

src/composables/visualizer/useVisualizerEngine.ts
```diff
@@ -0,0 +1,169 @@
...
+export interface VisualizerEngine {
+  loadPresetByName(name: string, blendSec?: number): Promise<void>;
+  loadRandomPreset(): Promise<string>;
+  destroy(): void;
+}
+
+/**
+ * Whether the browser can run butterchurn (WebGL2 required).
+ */
+export function isVisualizerSupported(): boolean {
```
stvncode, line 36: `isVisualizerSupported()` creates a fresh `<canvas>` + WebGL2 context on every call, and it runs (via `visualizerCanRender()`) inside the `visualizerActive` computed of every hosting view. Since `setUserPreference` replaces the whole `preferences` object, every preference write re-evaluates those computeds, so dragging the blur slider spawns a burst of throwaway WebGL2 contexts while the visualizer's own context is live. Chrome caps live WebGL contexts and drops the *oldest* on overflow which is the running visualizer, so the feature can kill its own rendering with a `webglcontextlost` the engine never handles.

WebGL2 support can't change within a session, so probe once and cache:

```suggestion
let webgl2Supported: boolean | null = null;

export function isVisualizerSupported(): boolean {
  if (webgl2Supported === null) {
    try {
      webgl2Supported = !!document
        .createElement("canvas")
        .getContext("webgl2");
    } catch {
      webgl2Supported = false;
    }
  }
  return webgl2Supported;
}
```

src/layouts/default/PlayerOSD/VisualizerMenuControl.vue
```diff
@@ -0,0 +1,277 @@
...
+  }
+  if (value === RANDOM_FAVORITES_VALUE) {
+    void setPreference("visualizer_preset_mode", "random_favorites");
+    return;
+  }
+  void setPreference("visualizer_preset", String(value ?? ""));
+  void setPreference("visualizer_preset_mode", "fixed");
+};
+
+const onBlurChange = (value: number[] | undefined) => {
```
stvncode, line 211: These fire on every `update:model-value` tick while dragging, and each `setPreference` call is a full `api.updateUser` round-trip plus a `preferences`-object replacement that re-evaluates every preference-derived computed app-wide. A single 0→30 blur drag is up to 30 API calls (and multiplies the WebGL context churn flagged in `useVisualizerEngine.ts`)

The reka-ui Slider (already forwarded by our shadcn wrapper) emits `@value-commit` at drag end. It keep `update:model-value` for the live preview if you like, but only persist on commit:

```vue
<Slider
  :model-value="[blurPref]"
  ...
  @value-commit="(v: number[]) => setPreference('visualizer_blur', v[0] ?? 0)"
/>
```

Alternatively debounce, matching the pattern `PlayerVolume.vue` uses for volume writes. Same applies to the opacity slider below and the three sliders in `VisualizerConfig.vue`

src/views/settings/VisualizerConfig.vue
```diff
@@ -0,0 +1,374 @@
...
+              {{ $t("settings.visualizer_opacity.description") }}
+            </div>
+          </div>
+          <div class="flex w-64 shrink-0 items-center gap-3">
+            <Slider
+              :model-value="[opacityPref]"
+              :min="10"
+              :max="100"
+              :step="5"
+              @update:model-value="
```
stvncode, line 102: Same issue as the sliders in `VisualizerMenuControl.vue`: this persists to the server on every drag tick. Please switch this one, the blur slider, and the beat-dwell slider to `@value-commit` (or debounce).

src/components/VisualizerCanvas.vue
```diff
@@ -0,0 +1,310 @@
...
+  if (!canvas || initialized) return;
+  // Remote (WebRTC) sessions cannot reach the relay route; starting up would
+  // only produce an endless connect/retry loop.
+  if (api.isRemoteConnection.value) return;
+  if (canvas.clientWidth > 0 && canvas.clientHeight > 0) {
+    void initialize();
+    return;
+  }
+  if (sizeObserver) return;
+  sizeObserver = new ResizeObserver(() => {
```
stvncode, line 217: Narrow race that defeats the single-engine guarantee: this observer is attached when the canvas is 0×0, but the callback never re-checks `covered`, and the `watch(covered)` teardown early-returns when `!initialized` without disconnecting a pending observer. So: dashboard mounts uncovered with an unlaid-out canvas → fullscreen player opens (`covered` = true) → canvas gains layout → this fires and starts relay + engine *behind* the fullscreen player → two engines render at once.

```suggestion
  sizeObserver = new ResizeObserver(() => {
    if (covered.value) return;
    if (canvas.clientWidth > 0 && canvas.clientHeight > 0) {
      sizeObserver?.disconnect();
      sizeObserver = null;
      void initialize();
    }
  });
  sizeObserver.observe(canvas);
```

(The later uncover path already calls `initializeWhenSized()`, which checks size before the `if (sizeObserver) return` guard, so a held-back observer still gets its init. Disconnecting the observer in the covered branch of the watch would work too.)

stvncode (approved):
Perfect, thanks 🙏

howels:
This plugin doesn't appear in the plugin list for 2.9.13 despite looking for all stages. Is this fully implemented?

jozefKruszynski:
It's complete, but it has not even been released in the latest beta yet. Only so far in nightlies.

howels:
> It's complete, but it has not even been released in the latest beta yet. Only so far in nightlies.

Ok, it appeared in the release note for 2.9.13 when I looked but the notes appear to have been revised now.

PR #2367, opened 2026-08-10
Handle 'stop' player command from companion apps

teancom:
The desktop companion app forwards OS media-control stop events (MPRIS Stop, macOS stopCommand, Windows SMTC Stop) as a 'stop' command to __COMPANION_PLAYER_COMMAND__, but handlePlayerCommand only knew play/pause/next/previous, so stop was logged as unknown and dropped. Route it to api.playerCommandStop like the others.


stvncode (approved):
Thanks 🙏

PR #2373, opened 2026-08-10
Match the mobile status bar colour to the app background

MarvinSchenkel:
On iPhone the status bar was always black instead of matching the app. iOS Safari tints that area from the page background, and ours was transparent because the colour is painted by `.v-application` rather than by `html` or `body`. The `theme-color` meta tags we had were keyed on `prefers-color-scheme`, which is the system setting rather than the app's own theme preference, and iOS ignores them anyway.

- Set `html` background to the Vuetify theme background, so the bar follows light, dark and auto
- Add `useStatusBarColor`, keeping the `html` background and the `theme-color` meta in sync, since iOS reads the first and Android Chrome the second
- Let the now playing screen override the bar with the top colour of its artwork gradient, cleared on close
- Replace the two `prefers-color-scheme` `theme-color` metas with one that tracks the active theme
- Point the manifest `theme_color` and `background_color` at the dark theme background instead of `#424242` and `#ffffff`

Verified on iPhone: normal screens match the theme background, the now playing screen matches the artwork, and going back no longer leaves the bar black.

### DIFF
--- a/index.html
+++ b/index.html
@@ -8,8 +8,8 @@
     <link rel="apple-touch-icon" href="apple-touch-icon.png" sizes="180x180">
     <link rel="mask-icon" href="favicon.svg" color="#FFFFFF">
     <link rel="manifest" href="manifest.webmanifest">
-    <meta name="theme-color" media="(prefers-color-scheme: light)" content="#ffffff">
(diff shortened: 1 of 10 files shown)


stvncode (approved):
Nothing to add, nice 🙏

PR #2380, opened 2026-08-10
Fix shortcut drag-reorder and duplicate prevention

dmoo500:
## Summary
Fixes two shortcut-related bugs:
1. **Drag-reorder failure**: Reordering shortcuts failed when the stored preference URI differed from the item's current URI
2. **Duplicate shortcuts**: Same item could appear multiple times in sidebar due to identity mismatches

## Changes
- **NavShortcuts.vue**: Use `findPinnedUriForItem()` to look up actual stored URIs for drag-reorder (instead of constructing new ones)
- **useShortcuts.ts**:
  - Introduce `isUriMatchingItem()` helper for provider-mappings-aware URI-to-item matching
  - Export `findPinnedUriForItem()` for components that need to resolve stored URIs
  - Update `pinShortcutStandalone()` to check identity match (not just exact URI)
  - Add Set-based deduplication in `pinnedItems` computed
  - Restore original `findPinnedUriByIdentities()` implementation to minimize diff

## Testing
- ✅ Drag-reorder shortcuts works correctly (even when stored URI differs from item URI)
- ✅ No duplicate shortcuts appear in sidebar
- ✅ Context menu shows correct pinned state
- ✅ All useShortcuts tests pass (9 tests)

## Related Issues
Fixes bugs identified by @STVNCode in review.

### DIFF
--- a/src/components/navigation/NavShortcuts.vue
+++ b/src/components/navigation/NavShortcuts.vue
@@ -32,7 +32,7 @@ import {
 } from "@/composables/useHoldToOpenMenu";
 import { useListDragReorder } from "@/composables/useListDragReorder";
 import {
-  getShortcutUri,
+  findPinnedUriForItem,
   reorderShortcutStandalone,
   useShortcuts,
   type ShortcutItem,
@@ -164,7 +164,10 @@ const {
     const source = pinnedItems.value[from];
     const target = pinnedItems.value[to];
     if (!source || !target) return;
-    reorderShortcutStandalone(getShortcutUri(source), getShortcutUri(target));
+    const sourceUri = findPinnedUriForItem(source);
+    const targetUri = findPinnedUriForItem(target);
+    if (!sourceUri || !targetUri) return;
+    reorderShortcutStandalone(sourceUri, targetUri);
   },
 });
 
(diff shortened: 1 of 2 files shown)


stvncode (commented):
- Drag-reorder in NavShortcuts.vue (line 163–168) silently fails for shortcuts whose stored URI differs from the resolved item's URI 
- Duplicate-identity preference entries (legacy state from this very bug, or newly created via pinShortcutStandalone) render the same item twice with duplicate :keys via the new pinnedItems matching

dmoo500:
Fixed both issues:

1. **Drag-reorder**: Now exports `findPinnedUriForItem()` so `NavShortcuts` can look up the actual stored preference URIs (instead of constructing new ones from the resolved item). This ensures reordering works even when the stored URI differs from the item's current URI.

2. **Duplicate shortcuts**: 
   - `pinShortcutStandalone()` now checks identity match (provider_mappings aware) instead of just exact URI match
   - `pinnedItems` computed uses a Set to deduplicate before rendering

Also restored the original `findPinnedUriByIdentities()` implementation from main to minimize the diff.

stvncode (commented):
Thanks @dmoo500 🙏

PR #2396, opened 2026-08-11
Keep the mobile menu bar clear of the phone's navigation buttons

marcelveldt:

The bottom menu bar did not account for the space phones reserve for their own
controls. On Android with 3-button navigation the system buttons landed on top
of Menu / Discover / Search, and on iOS the home indicator covered part of them.

Inside the Home Assistant panel the opposite happened on iOS: Home Assistant
already keeps the panel clear of the home indicator, and we reserved that same
space a second time, leaving a large empty band under the menu.

- The bar now keeps its buttons a fixed, small distance above whatever the
  system occupies, instead of padding that stacked on top of it
- The device inset is ignored when Music Assistant runs embedded in a frame,
  where the host sizes our viewport and already reserves that space
- The buttons sit closer to the player bar, and the bar is 12px shorter overall,
  so there is a bit more room for content
- Everything anchored to the bar (player bar, sheets, popovers, page scroll
  padding, settings save button, party view) now follows one shared
  `--mobile-navigation-height` value instead of ten copies of the same number

- user report, no issue filed

## Types of changes

## Checklist

- [x] `pnpm build` passes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,9 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
-    <!-- viewport-fit=cover is what makes env(safe-area-inset-*) resolve to the
-         real device insets instead of 0 -->
-    <meta name="viewport" content="width=device-width,initial-scale=1.0,viewport-fit=cover" />
+    <meta name="viewport" content="width=device-width,initial-scale=1.0" />
     <link rel="alternate icon" href="favicon.ico" type="image/png" sizes="16x16">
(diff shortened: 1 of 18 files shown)


marcelveldt:
Note for reviewers: no `viewport-fit=cover` here on purpose. Without it the browser keeps the page inside the safe area by itself, so nothing is covered. Opting into edge-to-edge would put the sidebar drawer, the bottom sheets and the landscape edges under the system bars - a much bigger job for no gain here.

The iOS case in the HA panel was the opposite problem: Safari does report the parent page's insets inside an iframe (Chrome reports none), so we reserved space HA had already reserved. That's what the iframe check in main.ts fixes.

PR #2400, opened 2026-08-11
Fix smart playlist last played validation and rule loading

dmoo500:
## Summary
Fixes validation error when using the "not played within" filter in smart playlists: `last_played_before_value and last_played_before_unit must both be set or both be None`

Also fixes rules not loading when editing existing smart playlists with duration or last_played filters.

## Changes
- **SmartPlaylistRuleRow.vue**: Ensure both `value` and `unit` are cleared together when either becomes undefined
- **useSmartPlaylistRulesForm.ts**: 
  - Load duration and last_played rules from `initialRules`
  - Fix null checks: use `!= null` instead of `!== undefined` (backend sends `null` for unset values)
  - Refactor 200+ line watch method into maintainable helper functions:
    - `buildRulesFromInitial()`: Constructs all rule rows from initial rules
    - `hydrateSeedItems()`: Loads and enriches seed items with API data
    - `watch()`: Now ~20 lines, coordinates initialization

## Testing
- ✅ Create smart playlist with "not played within X days" filter
- ✅ Clear the value field → no validation error, both value and unit cleared
- ✅ Edit existing smart playlist with duration filter → rules load correctly
- ✅ Edit existing smart playlist with last_played filter → rules load correctly

## Related Issues
Fixes user-reported validation error when clearing last played filter values.


src/composables/smart-playlist/useSmartPlaylistRulesForm.ts
```diff
@@ -209,6 +209,174 @@ export function useSmartPlaylistRulesForm(
...
+
+    if (
+      initial.last_played_before_value != null &&
+      initial.last_played_before_unit != null
+    ) {
+      const row = newRule("last_played", "is");
+      row.lastPlayedBeforeValue = initial.last_played_before_value;
+      row.lastPlayedBeforeUnit = initial.last_played_before_unit;
+      fresh.push(row);
+    }
```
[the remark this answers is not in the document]
dmoo500: fixed

src/components/smart_playlist/SmartPlaylistRuleRow.vue
```diff
@@ -376,17 +376,21 @@ function handleDurationBlur(e: Event, field: "min" | "max") {
...
     value,
-    unit: props.rule.lastPlayedBeforeUnit || "days",
+    unit:
+      value !== undefined
+        ? props.rule.lastPlayedBeforeUnit || "days"
+        : undefined,
   });
 }
 
 function handleLastPlayedUnitChange(
```
stvncode, line 383: selecting a unit before entering a value discards the selection (dropdown snaps back to "days") and the subsequent save silently uses "days" instead of the chosen unit. Enforce the both-or-neither invariant once in buildFinalRules() and revert the row-handler/newRule changes to their simpler forms

dmoo500: Great observation! You're right — the row-handler clearing logic broke the "select unit first, then enter value" workflow.

Fixed in 1c08760c:
- Simplified row handlers to pass through values as-is without clearing logic
- Moved the both-or-neither invariant enforcement to buildFinalRules() where it runs once at save time
- This allows selecting a unit before entering a value, and the backend validation now happens in the right place

stvncode (approved):
good thanks 🙏

PR #2444, opened 2026-08-12
Keep the blur behind the mobile player bar sized to the dock

marcelveldt:

The blur behind the mobile dock was a fixed height, but since #2441 the dock's
top is not: it follows the player card, which is measured at runtime. The two had
drifted apart, leaving only 16px of slack — a card taller than ~104px pushes the
see-through dock onto unblurred content, with the blur's edge cutting across it.
That is already reachable today at larger text sizes (a 1.5× text scale puts the
card at ~108px), and any extra row in the card would do the same.

The blur now measures the dock the same way the dock does, and clears it by a
fixed 16px. With a player active that works out to the same height as before, so
nothing looks different there. With no player selected the card drops its volume
row, and the blur now shrinks with it rather than staying full height — the halo
above the dock goes from 44px to the same 16px as everywhere else.

Builds on the `--mobile-dock-rim` token from #2443.

- n/a

- n/a

## Types of changes

## Changes

- Sized `--mobile-player-scrim-height` from the navigation, the measured player
  card and the dock rim, plus a fixed fade margin, replacing the hardcoded
  `180px`.
- Added `tests/styles/playerScrim.test.ts` covering the blur's clearance over the
  dock, the frame before the card has been measured, and that the navigation
  still lifts its surface to the same dock top.

## Checklist

- [x] `pnpm build` passes.

### DIFF
--- a/src/styles/global.css
+++ b/src/styles/global.css
@@ -47,10 +47,14 @@
   --mobile-navigation-height: calc(
     var(--mobile-navigation-item-height) + var(--mobile-navigation-inset-bottom)
   );
-  /* Height of the blur behind the floating player bar. It softens what scrolls
-     under it without hiding it, and fades out towards the top. */
+  /* Height of the blur behind the dock. It softens what scrolls under it without
+     hiding it, and fades out towards the top, so it clears the dock's own top by
+     a further 16px for the fade to finish above it rather than across it. The
+     card is measured at runtime, so this follows it rather than assuming a height
+     the card is free to outgrow. */
   --mobile-player-scrim-height: calc(
-    180px + var(--mobile-navigation-inset-bottom)
+    var(--mobile-navigation-height) + var(--player-bar-overlay-height, 0px) +
+      var(--mobile-dock-rim) + 16px
   );
   /* Space the player bar occupies in the desktop layout, where it sits flush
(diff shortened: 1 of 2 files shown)


marcelveldt:
Checked both of Copilot's suppressed notes on `dockSurfaceTop()`; neither holds, so
leaving the helper as is.

**On the `TypeError`:** not reachable. Optional chaining short-circuits the whole
rest of the chain, so `.replace()` is never evaluated when the selector or the
`top:` match comes back nullish — it returns `undefined` and the assertion fails
normally. I did add a message to that `expect` so a break says what to look at
instead of just "expected undefined".

**On using the CSSOM instead:** doesn't work here. The limitation isn't computed
styles, it's happy-dom's CSS parser — it drops a `calc()` containing a `var()` at
parse time, for stylesheet rules as much as for computed values. Probed it against
this exact declaration: the rule is found and its siblings read fine
(`style.zIndex` → `"-1"`), but `style.top` → `""` and `top` is absent from
`style.cssText` altogether. So that version of the test would assert `"" === ""`
and keep passing even if the `top` declaration were deleted outright, which is the
one thing it exists to catch.

The regex normalises whitespace for that reason, and it's already proven against a
declaration prettier wraps over three lines.

PR #2456, opened 2026-08-12
Keep the volume button's screen reader state in one place

marcelveldt:

The volume button in the player bar told screen readers itself whether its panel was open (`aria-expanded`) and what kind of panel it opens (`aria-haspopup`). The popover it lives in already announces both, so the same information was being written twice. Both copies happen to agree today, so nothing is wrong on screen or in a screen reader — but the hand-written pair can silently drift away from the real panel state, which is exactly the kind of bug that only shows up for the people relying on it.

- n/a

- n/a

## Types of changes

## Changes

- Removed the duplicated `aria-expanded` and `aria-haspopup` from the volume button; the popover trigger supplies both
- Kept the button's own `aria-label`, which spells out the current volume — the popover has no notion of that
- Added a test that mounts the real popover and pins the attributes to it, including the flip to open

Same fix as #2415 did for the group button, and verified the same way.

## Checklist

- [x] `pnpm build` passes.

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerBarVolumeControl.vue
+++ b/src/layouts/default/PlayerOSD/PlayerBarVolumeControl.vue
@@ -28,8 +28,6 @@
         :data-suppress-hover="suppressHover"
         :disabled="disabled"
         :aria-label="`${$t('audio_overlay_volume')}: ${displayVolume}%`"
-        :aria-expanded="open"
-        aria-haspopup="dialog"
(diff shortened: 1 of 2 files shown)


marcelveldt:
Copilot's point was fair: the real-reka test on its own can't tell a re-added duplicate from reka's own attributes. Added a second check on the stubbed harness (where the trigger supplies nothing), and confirmed it fails when the attribute comes back.

PR #2457, opened 2026-08-12
Keep the smart playlist picker's list steady while it closes

marcelveldt:

The smart playlist rule value picker cleared its search when the popover
closed. The popover stays on screen for its 150ms fade-out, so the list the
user had just filtered down visibly jumped back to the full list while it
faded away — the same glitch already fixed in the filter dropdown (#2433) and
the AI radio playlist picker (#2435).

The search is now cleared when the picker opens instead, so the closing
animation keeps showing the list the user was looking at.

- Clear the search term, results and spinner on open instead of on close
- Drop the duplicate clearing when a value is picked, so picking only removes
  the picked row from the closing list instead of re-expanding the whole list
- Add tests for the picker, including one that holds the closing animation
  open to check what stays on screen

- n/a

- n/a

## Types of changes

## Checklist

- [x] `pnpm build` passes.

### DIFF
--- a/src/components/smart_playlist/SmartPlaylistRuleValuePicker.vue
+++ b/src/components/smart_playlist/SmartPlaylistRuleValuePicker.vue
@@ -156,9 +156,10 @@ watch(query, (q) => {
   runSearch(q);
 });
 
-// Reset transient state when the popover closes.
+// Closing only unmounts the popover content, so the search has to be dropped by
+// hand - on open, to keep the restored list out of the closing animation.
 watch(open, (isOpen) => {
-  if (!isOpen) {
+  if (isOpen) {
     query.value = "";
     remoteResults.value = [];
     isSearching.value = false;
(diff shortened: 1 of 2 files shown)


marcelveldt:
Note on the pick path, since it came up in review: dropping the clearing in `onPick` means the closing popover keeps the filtered list minus the picked row, instead of re-expanding to the full list. When the picked row was the only match that leaves the "no matches" empty state on screen for the fade, which looks like a downgrade — but measuring the popover's height during the close says otherwise:

| filter → pick | before pick | this PR | previous |
| --- | --- | --- | --- |
| "jazz" → Jazz | 141px | 148px | 260px |
| "e" → Metal | 186px | 164px | 228px |

Keeping the old clearing grows the popover by up to 119px while it fades out, and refills it with rows that never matched what was typed. At 150ms fading to zero the shape change is what registers, not the sentence.

PR #2458, opened 2026-08-12
Play an announcement from the player menu

marcelveldt:

There was no way to make an announcement from the Music Assistant interface. Now that the server can speak a text message on a player, the player menus offer it directly: type what should be said, choose whether the chime plays first, and send it.

- New "Play announcement" entry in the player menu and the fullscreen player menu
- Dialog to type the message and toggle the pre-announce chime
- The entry only appears for players that support announcements and when a text-to-speech engine is set up

- n/a

- https://github.com/music-assistant/server/pull/5621

## Types of changes

## Checklist

- [x] `pnpm build` passes.


marcelveldt:
Both fixed. The catalog now shares one in-flight request, and the composable is only touched when the entry is actually a candidate.

Separately, the independent review caught something bigger: the entry was gated on PlayerFeature.PLAY_ANNOUNCEMENT, but that flag means the player announces *natively* - the server falls back to its own implementation for everything else. So it was hidden on Chromecast, DLNA, squeezelite, universal groups etc. while the command works fine there. Gate is now on non-protocol players. Also seeded the chime toggle from the player's own setting, it was silently overriding it.

marcelveldt:
Also took the suppressed one: the catalog is only refetched while there's nothing to speak with, so once engines are loaded a menu open costs nothing.

PR #2465, opened 2026-08-13
Keep the focus on the player menu when it opens on touch

marcelveldt:

Opening the player menu on a touch device left the keyboard closed, but it also left the focus nowhere. The handler focused the element reka reports the opening focus on, which is the wrapper it positions the panel with — that wrapper carries no tabindex and cannot hold focus, so focus stayed outside the panel: no screen reader announcement, and the tab order started outside the menu.

It also keyed on the mobile layout instead of on a touchscreen. Those are not the same thing: a phone held sideways counts as desktop and got no protection at all, while a narrow desktop window with a mouse ran the handler it never needed.

Every other picker in the app already routes this through the shared `preventOnScreenKeyboardOnOpen` helper, which aims at the panel nested inside that wrapper and gates on the touchscreen. The player menu now uses it too.

- Use the shared focus helper in the player menu instead of its own copy
- Focus now lands on the panel itself, so it is announced and keeps the tab order
- The keyboard stays down on any touchscreen, whichever way the phone is held
- Cover it with tests that mount the real popover, since the previous ones drove a stub that was focusable when the real panel wrapper is not

## Types of changes

## Checklist

- [x] `pnpm build` passes.


marcelveldt:
@copilot resolve the merge conflicts in this pull request

PR #2493, opened 2026-08-13
Reload the app when a page fails to load after an update

marcelveldt:

After an update, opening a page whose code hadn't been loaded yet could leave you
stuck. The app is supposed to notice this, reload itself and continue to the page
you clicked, but that reload never happened: because the app uses `#` based
addresses, the address it sent the browser to only differed after the `#`, so the
browser just moved to that spot on the same outdated page. You had to refresh by
hand to get going again.

- Reload the page explicitly after pointing it at the page you clicked
- Keep the rest of the address intact, so Home Assistant ingress sessions no longer lose their parameters on such a reload
- Let the recovery run again after a later update instead of only once per browser session
- Cover the recovery with tests

- n/a — spotted while reviewing #2483

- n/a

## Types of changes

## Checklist

- [x] `pnpm build` passes.

### DIFF
--- a/src/plugins/router.ts
+++ b/src/plugins/router.ts
@@ -655,6 +655,8 @@ const router = createRouter({
 // When a dynamic import fails with a 404, it means the chunk no longer exists
 // on the server (likely due to a new deployment with different hashes).
 // In this case, we reload the page to get the fresh assets.
+const CHUNK_RELOAD_STORAGE_KEY = "chunkReloadAttempted";
+
 router.onError((error, to) => {
   // Check if this is a chunk loading error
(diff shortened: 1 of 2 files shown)


marcelveldt:
On the two suppressed Copilot comments:

- Modelled \`href\` in the location stub. An href-based recovery already failed the tests, but now it says so: \`expected [ 'href' ] to deeply equal [ 'hash', 'reload' ]\`.
- Skipped the query param assertion. Nothing in the recovery writes to \`location.search\`, so asserting it survives can never fail - the step order above is what actually catches an implementation that would drop it.

PR #2502, opened 2026-08-13
Search command for desktop

stvncode:

https://github.com/user-attachments/assets/d2009e18-ab60-42fb-a747-25f5c6190f87

## Types of changes

## Checklist

- [ ] `pnpm lint:check` passes.
- [ ] `pnpm test:run` passes, and tests have been added/updated where applicable.
- [ ] `pnpm build` passes.
- [ ] I have read and complied with the project's [AI Policy](https://github.com/music-assistant/.github/blob/main/AI_POLICY.md) for any AI-assisted contributions.


stvncode:
@marcelveldt, before this is going to the abyss because the position of the search and also how to handle the internal search will take a bit of time, i basically remove my button in the toolbar and i put back the search item in the navbar. What this pr is adding is the command center that is only accessible by cmd + k for now. So at least we have this for 2.10 and we can think about the rest for 2.11

marcelveldt:
> @marcelveldt, before this is going to the abyss because the position of the search and also how to handle the internal search will take a bit of time, i basically remove my button in the toolbar and i put back the search item in the navbar. What this pr is adding is the command center that is only accessible by cmd + k for now. So at least we have this for 2.10 and we can think about the rest for 2.11

The problem I have with the searchbar in the sidebar is that there are now 2 searchbars visible - clearly visible in your screen recording as well. If we want to do this, we do it right.

But you can ofcourse prepare it and already do the CMD-K approach ?

Why don't you replace the current global search page with what you built now ? It is miles better than what we have today so it's a win from day one.

stvncode:
So the video is not really up to date. You don't have the button to open the search with cmd + k now. 
I kept the other search page for mobile because you don't have command there and also to when you click on the command on "see all items" you go to the search page, that's why i kept it

marcelveldt:
But can you remove the searchbar from the navigation pane ?

PR #2542, opened 2026-08-17
Highlight active audiobook chapters

teancom:
# What does this implement/fix?

We've had a 'faux-queue' entry for audiobooks and podcasts with chapters for a while, but we don't show which is the current chapter. This adds a 'highlight' to the current chapter, along with various tests.

Screenshot 2026-08-16 at 9 21 58 PM

Screenshot 2026-08-16 at 9 22 07 PM

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [X] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [X] The code change is tested and works locally.
- [X] `pnpm lint:check` passes.
- [X] `pnpm test:run` passes, and tests have been added/updated where applicable.
- [X] `pnpm build` passes.
- [X] I have read and complied with the project's [AI Policy](https://github.com/music-assistant/.github/blob/main/AI_POLICY.md) for any AI-assisted contributions.

PR #1063, opened 2025-08-03
Feature: audiobook/podcast skip controls

andykelk:
Adds configurable skip controls for audiobooks and podcasts with user-selectable skip amounts of 10, 15, 30, or 60 seconds.

 **Features**
-  Forward and backward skip buttons appear only when playing audiobooks or podcasts
-  Users can choose skip duration (10/15/30/60 seconds) in Settings > Audiobooks & Podcasts
-  Rapid clicking accumulates skip time instead of jumping to the same position
-  Works across mini player, fullscreen player, and browser media session controls
-  Automatically detects audiobooks, podcasts, and podcast episodes

 **Implementation Details**

-  Skip buttons positioned between previous/next and play controls
-  Generic fast-forward/rewind icons
-  Shared SkipControlManager handles throttling logic similar to existing media session controls
-  Browser media session uses configured amounts for keyboard/headphone controls
-  Falls back to 10 seconds for music content to maintain existing behaviour

 **Test Plan**
-  Skip buttons appear only for audiobooks and podcasts
-  Setting configurable in frontend settings under new "Audiobooks & Podcasts" category
-  Skip amount changes are applied immediately without page refresh
-  Rapid clicking accumulates skip time correctly
-  Browser media session (keyboard/headphone controls) uses configured amounts
-  No skip buttons shown for regular music content

 Fixes https://github.com/orgs/music-assistant/discussions/3745

PR #2548, opened 2026-08-17
Add audiobook and podcast chapter-based progress

teancom:

When playing back an audiobook (or podcast with defined chapters) treat each chapter as an individual track. Opt-in, so the default doesn't change. And I'm not 100% happy with the name of the option ('Show audiobook chapter progress') though the description does a decent job of describing the actual functionality, so I'd love to get feedback on that in particular. Also update it so that next/prev changes chapters vs. 'real' queue items, when playing back long-form content, with handling for first and last chapters doing the right thing (actually changing tracks).

Screenshot 2026-08-17 at 1 23 16 PM

Inspired by https://github.com/music-assistant/frontend/pull/1203 but it's been open so long it was much easier to reimplement from scratch, given how much has changed since last November.

## Types of changes

- [X] New feature (non-breaking change which adds functionality) — `new-feature`

## Checklist

- [ ] `pnpm build` passes.
- [ ] I have read and complied with the project's [AI Policy](https://github.com/music-assistant/.github/blob/main/AI_POLICY.md) for any AI-assisted contributions.


teancom:
BTW, this PR pairs nicely with https://github.com/music-assistant/frontend/pull/2542 but they are not dependent on each other.

marcelveldt:
I might be mistaken but don't most podcast/audiobooks apps do this a bit different and show the chapters in a dedicated view/popout and have the skip buttons to skip ahead and back e.g. 15 seconds ?

I would instead make the most convenient experience for podcast and audiobooks listening the default instead of opt-in (it would be rather an opt-out if you dont want that)

teancom:
So we already have what I call the faux-queue when listening to an audiobook, https://github.com/music-assistant/frontend/pull/2542 is just adding the highlight to the current chapter. As for the skip ahead / skip back buttons, you're racing ahead of my re-implementation/modernization of https://github.com/music-assistant/frontend/pull/1063 which will be separate. 😄

And I can definitely switch this to on-by-default. I'm just generally nervous about that, feels braggadocious.

tests/helpers/chapters.test.ts
marcelveldt, line 1: why is the test here and not in tests folder ?

teancom: There are some other tests in the same folder so I was pattern matching. But easy enough to move it, done.

teancom:
Sorted out the merge conflicts as well. 👍

marcelveldt (approved):
Thanks, @teancom !

PR #2574, opened 2026-08-19
Show provider logos in the Browse list

OzGav:

Feature request from here: https://github.com/orgs/music-assistant/discussions/5970

The top level of the Browse view lists one entry per music provider, but every entry showed the same generic folder icon, so you had to read the names to tell them apart. Those entries now show the provider's own logo, the same one you see on the settings pages.

Folders inside a provider, and the ".." entry, keep the folder icon. A provider that has no logo falls back to it as well.

Some before and after pics

image

image

- n/a

- n/a

## Types of changes

Changes:

- Provider entries in Browse show the provider logo in both the list and the grid view
- Moved the provider icon loading out of `ProviderIcon.vue` into a shared `useProviderIcon` composable, so the badge, the list row and the grid tile all use the same theme and variant rules
- Added `getProviderRootDomain` to spot the browse entries that stand for a provider, with tests

## Checklist

- [X] `pnpm build` passes.


src/composables/useProviderIcon.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  const applyInvert = computed(
+    () =>
+      variant.value === ProviderIconVariant.MONOCHROME &&
+      !theme.current.value.dark,
+  );
+
+  const iconDataUri = ref<string | null>(null);
+  watchEffect(async () => {
```
stvncode: Async watchEffect with no staleness guard can write a stale icon; derive from the reactive api.providerIcons cache instead of assigning

OzGav: Good catch. It now reads the icon from the cache using the provider showing at that moment, so a slow response can't land on top of a newer one. Added a test that finishes the requests in the wrong order.

marcelveldt (approved):
Nice QoL improvement @OzGav !

PR #2577, opened 2026-08-19
Tint the MilkDrop visualizer with the track color

jozefKruszynski:

Tints the MilkDrop visualizer toward the current track's artwork color.

The visualizer relay client now parses the server's new `{"type": "color", ...}` messages (color@v1 fields, pushed by the milkdrop_visualizer tap when its `color_tint` provider setting is on, added in https://github.com/music-assistant/server/pull/5839) and accumulates them into a merged palette. `VisualizerCanvas` renders a tint layer between the canvas and the legibility scrim using `mix-blend-mode: color`, so the preset keeps its own luminance and only shifts hue toward the album color. The palette side follows the theme through `paletteFromServer` itself, the same helper the OSD gradient uses, so the two cannot pick differently: `on_light` in the dark theme, `on_dark` in the light theme, and no tint at all when that side is missing, since the OSD has no fallback either. The tint element is always mounted while streaming (transparent until a color arrives), so the color transition never animates from an unset just-inserted state.

The fullscreen player's text treatment is unchanged: above 50% opacity the view keeps the forced dark look with light text, tint or no tint. Because `mix-blend-mode: color` preserves the preset's luminance, a tint cannot guarantee a surface light enough for dark text, so white text over the darkened visualizer remains the only arrangement with a contrast floor. The relay does clear a tint kept across a reconnect when `stream/start` no longer advertises the `color` type (`color_tint` disabled server-side), so a stale tint cannot outlive the setting.

Related PR: https://github.com/music-assistant/frontend/pull/2588, which drops the Sendspin gate now that the waveform comes from playback.

- n/a

- related backend PR https://github.com/music-assistant/server/pull/5864

## Types of changes

## Checklist

- [x] `pnpm build` passes.


src/components/VisualizerCanvas.vue
```diff
@@ -118,6 +140,50 @@ const playbackPaused = computed(() => {
...
+
+// mix-blend-mode: color keeps the preset's own luminance, just shifts hue.
+const tintStyle = computed(() => ({
+  backgroundColor: tintColor.value ?? "transparent",
+}));
+
+// Read by other views (e.g. the fullscreen OSD). Ownership-checked: the
+// canvases overlapping in a fullscreen handoff must not clear each other.
+const tintFlagOwner = Symbol("visualizer-tint");
+watch(
```
stvncode: The last-writer ownership still breaks in the handoff it was added for, just in the other direction. With two canvases both tinting, whichever asserts second owns the flag. When it stops, it clears the flag while the first is still painting, and the first never re-asserts because its watcher source hasn't changed:

Dashboard canvas B tints → tintFlagWriter = B, flag true
Fullscreen opens, canvas A tints → tintFlagWriter = A, flag stays true
Fullscreen closes, A unmounts (line 444) → tintFlagWriter === A → flag false
B is still streaming and tinting, but nothing re-fires — the flag stays wrong for the rest of the session

This is reachable: the canvas is mounted at three sites (PlayerFullscreen.vue:18, DashboardNowPlayingView.vue:3, PartyDashboardView.vue:4), and the fullscreen one is v-if'd on store.showFullscreenPlayer, so it mounts and unmounts over a long-lived dashboard canvas.

A module-scoped Set<symbol> of currently-tinting owners with visualizerTintActive.value = owners.size > 0 on every transition and on teardown gets both directions right for about the same amount of code.

src/layouts/default/PlayerOSD/PlayerFullscreen.vue
```diff
@@ -1248,9 +1249,11 @@ watchEffect(() => {
   // theme/palette treatment instead of forcing the dark look. This component is
   // permanently mounted via the OSD footer, so gate on the fullscreen player
   // actually being open, or --text-color would stay forced app-wide.
+  // A live tint plus the scrim is already legible, so it keeps that treatment.
   if (
     store.showFullscreenPlayer &&
     visualizerActive.value &&
+    !visualizerTintActive.value &&
```
stvncode: This drops the legibility floor in the light theme. Previously opacity > 50 always forced white text over a dark base; now a live tint falls through to the normal theme treatment, which in the light theme is #000000 text over the visualizer, an arbitrary album tint, and the rgba(0, 0, 0, 0.25) scrim.

The problem is that nothing in that chain can lighten the composite. mix-blend-mode: color preserves the canvas's luminance by design, so the tint only shifts hue and the scrim is a darkening layer, which protects light text, not dark text. A dark preset plus a dark scrim under black text has no contrast guarantee.

Could this be gated on the dark theme only, or the text color derived from the tinted composite's luminance rather than from visualizerTintActive alone?

src/plugins/visualizer-relay.ts
```diff
@@ -178,8 +197,17 @@ export class VisualizerRelayClient {
       }
       if (message.type === "auth_ok") {
         if (this.ws) this.startTimeSync(this.ws);
+      } else if (message.type === "color" && message.payload) {
```
stvncode: Payload is only checked for truthiness, but line 205 then runs field in payload. A primitive payload like ```{"type":"color","payload":1}``` throws TypeError: Cannot use 'in' operator to search for 'background_dark' in 1, and it's uncaught: the try/catch above only wraps JSON.parse.

This is unvalidated wire data, and the server/time branch just below already guards its fields carefully for the same reason. A ```typeof payload === "object" && !Array.isArray(payload)``` check before the loop closes it. (Arrays are harmless "primary" in [] just returns false so only primitives are the risk.)

MarvinSchenkel (approved):

PR #2653, opened 2026-08-26
Scrolling rows on the artist page

OzGav:

The albums, top albums and similar artists lists on the artist page were shown as tall grids. Each one took up a lot of vertical space, so getting to the sections further down meant a long scroll past rows of covers you were not looking for.

Those three lists are now single rows that scroll sideways, the same way the rows on the home and discover pages work.

- Show albums, top albums and similar artists as one scrolling row each
- Scroll a row by dragging it with the mouse, by the arrows that appear when you hover, or by swiping on a phone or tablet
- Load the next batch of items when you reach the end of a row
- Move the shared row behaviour into its own component so the artist page and the discover shelves use the same code
- Fix the mouse wheel over these rows in Firefox, which moved the page a few pixels at a time instead of a normal step. The rows on the home and discover pages get that fix too

image

- Discord discussion https://discord.com/channels/753947050995089438/1004642314968367124/1524351413457518704

- N/A

## Types of changes

## Checklist

- [x] `pnpm build` passes.

### DIFF
--- a/src/components/ItemsListing.vue
+++ b/src/components/ItemsListing.vue
@@ -112,14 +112,53 @@
             <PanelViewSkeleton />
           </v-col>
         </v-row>
+        <ShelfTrack
+          v-else-if="viewMode === 'row'"
+          class="items-row"
+          :tiles-per-view="
+            panelViewItemResponsive($vuetify.display.width) + 0.5
+          "
+        >
+          <PanelViewSkeleton
+            v-for="n in 8"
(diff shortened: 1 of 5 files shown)


marcelveldt:
I don't think this is the way to fix this. This needs a better design first.
What we discussed is that we want widget rows for the first content and then detail pages to see the full content.
Also the design should re completely revisited.

Let's plan a design/vision session first before we tackle this.

PR #2671, opened 2026-08-28
Keep album art when its metadata provider is turned off

OzGav:

Images that are already a full web address were being thrown away when the provider that supplied them was disabled, because the check only looked at whether that provider was still loaded. Such an image carries everything needed to fetch it, so it now goes through regardless.

Detail pages hid the problem by falling back to another stored image. Tiles carry a single image, so they dropped to the initials placeholder instead.

- Gate on the image rather than the provider that supplied it
- Keep rejecting images that only the provider can resolve
- Cover the helper with tests

- related issue https://github.com/music-assistant/support/issues/6216

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue) — `bugfix`

## Checklist

- [X] `pnpm build` passes.

### DIFF
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -353,11 +353,13 @@ export const getMediaImageUrl = function (
 };
 
 /**
- * Check if an image provider is available.
+ * Check if an image can still be fetched.
  */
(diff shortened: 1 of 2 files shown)


OzGav:
closing as a duplicate

PR #2678, opened 2026-08-31
Update volume control behavior in PlayerVolume.vue

modernman1:

Mouse scroll wheel does not work to change the volume on individual speakers or speaker groups in the volume pop-up on the page that slides up when going to the current playing song. 

This change makes the scroll wheel work on these volume controls as it does elsewhere.

## Types of changes

- [X] Enhancement to an existing feature — `enhancement`

## Checklist

- [X] `pnpm build` passes.

### DIFF
--- a/src/layouts/default/PlayerOSD/PlayerVolume.vue
+++ b/src/layouts/default/PlayerOSD/PlayerVolume.vue
@@ -55,7 +55,11 @@
             <div class="group-popout-label">
               {{ truncateString(child.name, 20) }}
             </div>
-            <PlayerVolume :player="child" width="100%" />
+            <PlayerVolume
+              :player="child"
+              width="100%"
+              :allow-wheel="allowWheel"
+            />
           </div>
           <!-- Group volume at bottom with divider -->
           <div class="group-popout-divider"></div>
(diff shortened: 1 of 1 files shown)


stvncode:
@marcelveldt, wdyt about that ? I'm thinking about miss scroll that up the volume if not wanted and could be an issue

modernman1:
@stvncode - I understand concern of making change that might cause unintended volume change, but I don't understand 'miss scroll that up the volume'...  This change allows mouse wheel to adjust volumes by hovering over the control (slider) and moving the wheel. This is the current (and desired) behavior on the primary page for master and group volume - where a change was also made that affects _mouse _clicks__ .  These slide-up pages with volume did not get that change (which changes volume by small step instead of to the clicked position) - I don't think there is a situation where this wheel change would result in unintended volume change - you have to hover over the control for the wheel to function.

The wheel action is important to me: multi-monitor setup allows moving between monitors/windows and affecting windows that _do not have _focus__ and make an adjustment without changing Windows focus, so can continue to type and do other keyboard work without having to click in and out of wiindows.  The Master volume does work with wheel on the slideup (see screenshots), but when click, the popup volume (2nd screenshot) appears and the behavior is bad - wheel does not work, and clicking/tapping on the volume moves to the setting that was tapped.

Here are screenshots:
image

image

PR #2679, opened 2026-08-31
Add the item menu to search results

OzGav:

Search results in the search popup now have the usual item menu. Right click a result on desktop or hold it on touch to play the item, manage favourites and library membership, add it to a playlist, and so on. This was reported as an issue as you couldn't easily build a playlist from the search screen as you had to close it each time you wanted to add an item and you had to then reopen and re-search.

Getting this working also surfaced a few problems around the menu itself, fixed here as well.

- clicking away from the menu no longer closes the search popup underneath it
- remove from library and remove from favourites now work for results that came from a streaming provider, since those actions need the library copy of the item
- a track a provider marks as liked is no longer treated as a favourite unless it really is one in your library
- the remove confirmation now appears on top of the search popup instead of hidden behind it
- the menu checks the library when it opens, so rows left over from an earlier search no longer offer removals that fail, and a removed item can be added back from the same row

## Types of changes

## Checklist

- [x] `pnpm build` passes.

### DIFF
--- a/src/components/CommandCenter.vue
+++ b/src/components/CommandCenter.vue
@@ -108,9 +108,13 @@
         <CommandItem
           v-for="item in section.items"
           :key="item.uri"
+          v-hold="(evt: Event) => onHold(evt, item)"
           :value="item.uri"
           class="gap-3 py-2"
           @select="onMediaSelect(item)"
+          @contextmenu.prevent="onMediaMenu($event, item)"
+          @click.capture="swallowClickAfterHold"
+          @touchstart.passive="onTouchStart"
         >
           <div class="relative size-10 shrink-0 overflow-hidden rounded-md">
             <MediaItemThumb :item="item" :size="40" />
@@ -239,13 +243,20 @@ import {
   commandCenterHotkeyLabel,
   useCommandCenter,
 } from "@/composables/useCommandCenter";
+import {
+  getEventPosition,
+  useHoldToOpenMenu,
+} from "@/composables/useHoldToOpenMenu";
 import { useOrderedPlayers } from "@/composables/useOrderedPlayers";
 import {
   SEARCHABLE_MEDIA_TYPES,
   useProgressiveSearch,
 } from "@/composables/useProgressiveSearch";
 import { useUserPreferences } from "@/composables/userPreferences";
-import { handlePlayBtnClick } from "@/helpers/media_item_actions";
+import {
+  handleMenuBtnClick,
+  handlePlayBtnClick,
+} from "@/helpers/media_item_actions";
 import { getArtistsString, getPlayerName } from "@/helpers/utils";
 import { getListItemProviderIconDomain } from "@/plugins/api/helpers";
 import {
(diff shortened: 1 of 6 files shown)


marcelveldt:
Does this also add the 3 dots contextmenu to each line ?

We should btw consider backporting this because this is a real miss in the current stable - I was about to look into it myself as I also found this flaw while using search

OzGav:
No I wasn't sure how much change to make to the actual look so I just sent with right click / long press to bring up the menu

stvncode:
That's something i wanted to add at some point. We just have to be careful with the dropdown. Since you are in a portal (the search modal), opening another portal might break it. I think you need to play with the zindex and also the prevent default event 👍

OzGav:
The z-index turned out to be fine already. The menu sits well above the search popup so it comes up on top without any changes. The one thing that did need sorting was the delete confirmation. It shares a z-index with the other dialogs so it was appearing behind. I moved it to render last so it appears on top.

The portal nesting broke in two ways. One press was closing both the menu and the search popup, because they both watch for a click outside and both reacted to the same press. The menu now swallows it so only the menu closes. The other was our dialogActive flag getting cleared when the menu closed, which left the search popup in a half open state. That now gets remembered and put back.

Prevent default is in there for the right click, and long press on touch has its own handling so the row underneath doesn't open as well.

Tested by hand on desktop Chrome and Android Chrome. Escape closes just the menu, clicking outside closes just the menu, the search box still takes typing afterwards, and long press opens the menu with nothing else firing. Also went search, menu, remove from library, confirmation on the phone. The confirmation came up on top and cancelling left the search working normally.

modernman1:
Sorry to have created Issue about this - I did not see this change mentioned in any of the recent release notes.
Personally, I do not like the change at all - I don't understand why the design choice was made in the first place - a modal small window that needs scrollbars to display content while the entire background of the large screen behind it is blurred makes zero sense to me.  

I have updated to latest MA release and this is so problematic: there is still no way to work with the search results aside from clicking the PLAY button to interrupt and immediately play a selection, or click on the selection to move to a particular search result. The tragedy is that when go back to the search popup (CTRL+K) it goes back not to the other search results, but back to the ALL tab with recent searches!   

Suggestions:  Give option to have the old full frame search - that would be my preference - if there is use case for this revision, support both and let user choose.

If not going to give full screen back, give some option to control the size of this popup so can eliminate horizontal scroll bar for the type of search selection.  Add the hamburger menu option to give the queue options, add to favorites, etc; AND, make the last search type (All, Tracks, Songs) etc persistent so it when re-opening it defaults back to the last search type used - but this still leaves the inability to do decent research into a search results list as can't get back to the last results once anything is done that makes the popup disappear.

image

marcelveldt:
The reason for global search in a modal was that you can quickly search for something anywhere in the app without leaving the page where you were. But I do agree that the modal is small and without the contextmenu a bit annoying even.

modernman1:
> The reason for global search in a modal was that you can quickly search for something anywhere in the app without leaving the page where you were. But I do agree that the modal is small and without the contextmenu a bit annoying even.

It is way more than being small:  the functionality is so bad I'm trying now to fork the frontend and forget freeze with something that works for me:  the problem is the way I used search, I would get a list, and with the hamburger menu be able to add to cue, etc; I could also click on a search result and get more info, and be able to GO BACK to the original search results to explore other results - this is no longer possible - if I search for 'Let It Be'  there will be lots of results from variety of artists. The process now of looking at more than one search result is so cumbersome as to be worth my time to spend days to fix this (or go back to an old version of MA and refrain from updating until it no longer works...)

PR #2687, opened 2026-09-02
Improve player and group controls in playback bar

trisweb:
This change improves the player & group controls to be more logical and easy to understand.

- Reorder playback-bar controls so the current player appears before its group control.
- Show “Group” with a more appropriate icon when no group is active, to afford "grouping."
- Show the member count with a reasonable 'group' icon (I went with Lucide "copy" but open to ideas) when a group is active.
- Apply the state-aware icons to the compact mobile control as well.
- Tests added appropriately, and test suite passes.

### Before

image

#### No group yet active:

image

#### Group active:

image

#### Mobile:

Screenshot From 2026-09-01 21-44-28
Screenshot From 2026-09-01 21-44-41
Screenshot From 2026-09-01 21-44-52

(Note: I don't think the mobile approach is great here; ideally the speaker remains tied to the grouping functionality in proximity and function; but this keeps the status quo for now, and I think is still an improvement on the former approach on mobile)

Fixes UI confusion between Speaker and Group controls, and improves their clarity and logical order.

## Types of changes

## Checklist

- [x] `pnpm build` passes.


marcelveldt:
Not sure about this entirely;

- the current implementation came after a design phase which was tested and shared with the community before implementing
- I think the active speaker button should always be at the right
- I'm not sure people will understand these icons ?

Can you make the change of the active speaker always full right and then first share the design on discord ?

trisweb:
(Apologies, I replied to the wrong PR)

Thanks @marcelveldt. I can definitely take this back a step and discuss on Discord first. However, the ordering of the buttons _is_ the important change here, so that will be the primary thing I'd like to discuss.

I can understand why the conclusions of "primary button far right" might have been reached, but I think there are ways to make this clearer while keeping the principles that lead to that decision. For example: maybe this should be one button.

PR #2688, opened 2026-09-02
Improve player card hierarchy and playback status

trisweb:

This improves the visual hierarchy and clarity of the Players panel and player cards, and reduces usability issues with quickly and easily finding and selecting Players, primarily due to a few key inconsistencies causing unpredictable behavior.

### Changes

tl;dr: redesigns the Player panel and functionality within it holistically.

- Make the layout of the player cards consistent, so the title is always in the same place. This resolves scannability issues.
- Reduce visual noise around the player cards
- Add a "Now playing" indicator to now playing speakers
- **Make the order of Players stable: the selected player no longer moves to the top.** This seems like a big change, but in usage the impact is only positive IMO: the position of players quickly become predictable, and the user can build muscle memory to make switching fast and requiring little thought.
- Even spacing around the panel
- Reduced grab-handle height, more proportional
- Minimal but clear view when no song is active, still clear when selected and still has all actions available
- Also use the same playback indicator on Discover player cards for consistency

## Screenshots

Screenshot From 2026-09-02 21-58-38 Screenshot From 2026-09-02 21-58-29

## Types of changes

## Checklist

- [x] `pnpm build` passes.

### DIFF
--- a/src/layouts/default/PlayerSelect.vue
+++ b/src/layouts/default/PlayerSelect.vue
@@ -106,7 +106,7 @@
         ref="playerList"
         class="player-volume-scroller min-h-0 flex-1 overflow-y-auto pb-6"
       >
-        <!-- outranks the selected player badge on the cards scrolling underneath -->
+        <!-- keeps search controls above the cards scrolling underneath -->
         <div
           v-if="showSearch"
           class="bg-background sticky top-0 z-20 px-3 pt-3 pb-2"
--- a/src/layouts/default/PlayerSelect.vue
+++ b/src/layouts/default/PlayerSelect.vue
@@ -74,30 +74,6 @@
             </DropdownMenuLabel>
             <DropdownMenuSeparator />
             <DropdownMenuCheckboxItem
-              :model-value="showSelectedPlayerFirst"
-              @select.prevent
-              @update:model-value="
-                setBooleanPreference(
-                  PLAYER_SELECT_PREFERENCES.showSelectedPlayerFirst,
-                  $event,
-                )
-              "
-            >
-              {{ $t("player_select.show_selected_player_first") }}
-            </DropdownMenuCheckboxItem>
-            <DropdownMenuCheckboxItem
-              :model-value="showActivePlayersFirst"
-              @select.prevent
-              @update:model-value="
-                setBooleanPreference(
-                  PLAYER_SELECT_PREFERENCES.showActivePlayersFirst,
-                  $event,
-                )
-              "
-            >
-              {{ $t("player_select.show_active_players_first") }}
-            </DropdownMenuCheckboxItem>
-            <DropdownMenuCheckboxItem
               :model-value="showGroupMemberNames"
               @select.prevent
               @update:model-value="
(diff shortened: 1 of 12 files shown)


marcelveldt:
Should we add some placeholder text for a player that is idle and has nothing loaded ?
I agree with the consistency fix for always having the player names in the same place but at the same time these empty player cards now take a lot of precious space. Any ideas for that ?

trisweb:
Thanks @marcelveldt. I can definitely take this back a step and discuss on Discord first. However, the ordering of the buttons _is_ the important change here, so that will be the primary thing I'd like to discuss.

I can understand why the conclusions of "primary button far right" might have been reached, but I think there are ways to make this clearer while keeping the principles that lead to that decision. For example: maybe this should be one button.

trisweb:
Yes, good idea, I think it can be made more minimal in that state still. I will try some things. Thanks!

trisweb:
@marcelveldt what do you think of this? I really like it.

* Even spacing around the panel
* Reduced grab-handle height, more proportional
* Minimal but clear view when no song is active, still clear when selected and still has all actions available

And one that's likely to be controversial: the ordering of this list is stable. The list does not reorder the selected player to the top. To me, this is a massive improvement; the players are always in a predictable location, so the user can build muscle memory on switching, which is a frequent and core action IMO. It feels very good to me, natural.

Let me know what you think.

image

Screenshot From 2026-09-02 21-58-29

src/layouts/default/PlayerSelect.vue
```diff
@@ -250,14 +224,6 @@ const showVolumeForActivePlayersOnly = getPreference<boolean>(
   PLAYER_SELECT_PREFERENCES.showVolumeForActivePlayersOnly,
   true,
 );
-const showSelectedPlayerFirst = getPreference<boolean>(
```
trisweb, line 253: This preference likely needs cleaning up. I can add that.

I recommend this not being a user preference; we can be confident in the stable order being the appropriate UI decision.

marcelveldt:
> The list does not reorder the selected player to the top. To me, this is a massive improvement

You are aware that this was a setting in the top right context menu ?
What if I have 20 players ?
There has been a whole design discussion about this, presented to people, tweaks and now you want to change it again haha.

You want to quickly find the relevant player(s) - the actively selected one, the ones that are currently playing.

I don't think we should remove the ordering of players (there is already a setting for that) but we should maybe not yet do the jump while you are still on the list.

trisweb:
Working my way through understanding the history, thanks for your patience as always. Discord is hard for me as I get easily distracted; it's like walking into a casino, and I try to avoid it. But I will discuss more there on bigger decisions (I knew that one was not a slam dunk 😅).

Happy to go back on the sorting, but I like the idea of not immediately having the UI jump on selecting a player, that would help.

I think the remainder of the visual improvements here are pretty straightforward improvements. 

I'll take a look at the sorting preference and make sure it's respected and not change it here.

src/layouts/default/PlayerSelect.vue
marcelveldt: I dont agree, sorry but I do agree its really annoying that the speaker jumps.
if we can follow sonos style by default (and offer the toggles for those that dont agree) that would be great:
- selected/active player at the top
- playing players next
- the rest after that alphabetically
- dont live jump the list when selecting a player (only grab the sort order on open of the dialog)

marcelveldt:
Can you also think about my comment about the icons ?
Icons are great to uniquely identify your players and in fact we have started our own catalog of icons for this reason.
maybe we should care less about the now playing thumbnail in this player selector and care more about the player icon ?

so make it consistent having the player icon + name at the same location, regardless the status

PR #2695, opened 2026-09-07
Improve Player Group UX clarity

trisweb:
### What does this implement/fix?

Improves clarity of the Group selector, with a clearer starting state that indicates the function of the control. Keeps the order and general approach the same.

Adds a subtle visual cue to indicate the two controls are linked, that changes slightly when a group is active to indicate they're linked more.

Changes icons to be more intuitive on the group creation action when not yet grouped, and group state when grouped.

#### Before / After

image Screenshot From 2026-09-06 22-30-22

#### After, when a group is active:
Screenshot From 2026-09-06 22-30-37

---

#### Mobile:
image

---

Design note: this does not fix all of the issues with this experience, as there are still pretty significant problems with how the concepts work. However, more substantial changes will require more time, and this is a solid improvement to the current approach for now.

---

**Method**: OpenAI Codex / GPT 5.6 Sol, with my complete review and approval.

## Types of changes

## Checklist

- [x] `pnpm build` passes.


marcelveldt:
Hey @trisweb,

sorry for the late reply! As I already told you on discord this needs a bit more discussion since some of the changes were made on purpose after careful consideration and some testing in both the Figma design and the beta cycle of 2.10

We do agree that the grouping icon being the same for the default speaker icon is weird and needs to be adjusted.
The whole idea is/was that each player just has its own unique icon so we have been focusing on that but ofc reality is that a lot of speakers cary the default speaker icon. And that being the same as the grouping icon is not good.

We do think that changing the icon for grouping is a need but we don't think that the plus button to form a new group is good. Also the blue indicator is something we deliberately changed as it looked too much like a notification. Hence teh less prominent grey-ish count bade.

What we want to propose to change on this PR is to just always use your proposal of the icon with the 2 squares to indicate grouping controls, with the count badge as how it currently is (so not blue). No separate icon to form a new grouped playback.

We should then also consider changing the icon in the playerselect menu where you pick a player.

marcelveldt:
@trisweb I went ahead and pushed the above change to your PR. Let's first test how that works in the beta channel before we make more changes.

marcelveldt (approved):
Thanks, @trisweb !

trisweb:
@marcelveldt thanks! Sounds great.

PR #2712, opened 2026-09-10
Show radio stream artwork in the Party dashboard

### LINKED ISSUE #6281: Player drawer image art (radio stream) not carried through to Party plugin
### Music Assistant version

2.10.1

### Browser and operating system

Chrome

## Summary

The Party dashboard's blurred album-art background and album-art shows a placeholder when a radio stream is playing, even though the player drawer on the same screen shows the correct (now-playing) artwork. Spotify/track playback is unaffected.

### How to reproduce

Listen to a radio station that is displaying album art - then click party

### Screenshot or recording

Image

pierosavi:

`getMediaItemImage` never used a radio queue item's live `streamdetails.stream_metadata.image_url`, `streamdetails in mediaItem.media_item` check tested the wrong object, so it fell back to the station's static artwork.

`getMediaItemImageUrl` still routed the resulting image through the legacy `/imageproxy?path=&provider=` form when it had no `proxy_id`. Servers on schema 31+ reject that legacy form with a 400, so the image failed to load and fell back to a placeholder. Now falls back to the image's own URL when there's no proxy_id.

#### Before 

image

#### After

image

**Related issue:**

Closes https://github.com/music-assistant/support/issues/6281

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue) — `bugfix`

## Checklist

- [X] `pnpm build` passes.

### DIFF
--- a/src/helpers/utils.test.ts
+++ b/src/helpers/utils.test.ts
@@ -4,7 +4,7 @@ import type {
   MediaItemType,
   QueueItem,
 } from "@/plugins/api/interfaces";
-import { describe, expect, it, vi } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 
 // `getProvider` decides whether an image is considered fetchable: an
 // unloaded/disabled provider is absent from the map entirely. `schema_version`
@@ -37,6 +37,8 @@ const albumWith = (images: MediaItemImage[]) =>
   ({ name: "Black to the Blind", metadata: { images } }) as MediaItemType;
 
 describe("getMediaItemImage", () => {
+  afterEach(() => vi.restoreAllMocks());
+
   it("keeps a remote image whose provider is no longer loaded", () => {
     // artwork written by a metadata provider that has since been disabled: the
     // url is self-contained, so the server can still resolve and resize it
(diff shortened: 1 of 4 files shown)


MarvinSchenkel:
There is still an edge case in [the new fallback](https://github.com/music-assistant/frontend/blob/106e3e940d9220a30daba13adfc84cb1cd2b72d0/src/helpers/utils.ts#L468-L471): when the frontend uses HTTPS and the live artwork URL uses HTTP, `getMediaImageUrl()` converts it back into `/imageproxy?path=...`. That is the unsupported legacy endpoint this change is meant to avoid, so the artwork still fails to load.

Can we skip the legacy fallback on schema 31+ and retain usable station artwork when the live URL cannot load safely? Showing the HTTP live artwork over HTTPS would require a server-issued proxy URL. Please cover the HTTPS frontend + HTTP artwork case as well.

pierosavi:
@MarvinSchenkel Good catch, I removed the fallback to query param image proxy and added logic to fallback to station instead, added test case for HTTPS frontend with HTTP artwork

MarvinSchenkel (approved):
Nice work, thanks @pierosavi

PR #2717, opened 2026-09-11
Allow selecting text in read-only config label fields

MarvinSchenkel:
> Auto-triaged from music-assistant/support#6394 and opened as a draft for review.

Config label fields (`ConfigEntryType.LABEL`) show read-only informational text that users frequently need to copy, such as the OAuth callback URL in the Spotify setup wizard. `global.css` disables text selection app-wide (`user-select: none`), so this text could not be selected or copied.

This opts label fields back in to text selection, the same way `Diagnostics.vue` already does for log output. The rule also covers the rendered markdown descendants, since the global rule targets each element individually.

- related issue https://github.com/music-assistant/support/issues/6394

## Types of changes

## Checklist

- [ ] `pnpm lint:check` passes.
- [ ] `pnpm test:run` passes, and tests have been added/updated where applicable.
- [ ] `pnpm build` passes.

### DIFF
--- a/src/views/settings/fields/LabelField.vue
+++ b/src/views/settings/fields/LabelField.vue
@@ -7,6 +7,22 @@ defineProps<{ text: string }>();
 <template>
   <MarkdownText
     :text="text"
-    class="text-muted-foreground bg-muted/50 mt-2 mb-4 rounded-md px-3 py-2 text-sm leading-relaxed"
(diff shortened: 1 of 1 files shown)


pierosavi (approved):
LGTM, I suggest making `selectable-label` global here or the next time we need it, and updating `Diagnostic.vue`

PR #2737, opened 2026-09-12
Remove unused provider translation keys

marcelveldt:

A batch of `settings.*` strings about providers is no longer used anywhere in the app, but they are still shipped in `en.json`. This removes them.

Each key was checked individually against the whole codebase, including the dynamically built lookups (`settings.${...}`), before being removed. Nothing rendered changes.

Note: this only cleans up the English file. The same keys still sit in the other locale files and in the Lokalise project, and our Lokalise workflow only downloads (it never uploads `en.json`), so they need removing on the Lokalise side too before the stale entries disappear from the other languages.

**Removed (17 keys, all under `settings.`):**

- `musicproviders`, `pluginproviders`, `metadataprovider`, `musicprovider`, `pluginprovider`
- `no_providers`, `no_providers_detail`, `no_music_providers`, `no_music_providers_detail`, `no_player_providers`, `no_player_providers_detail`
- `need_help_setup_provider`, `not_loaded`, `add_new_provider_button`, `edit_provider`, `providers_description`, `provider_type`

**Kept on purpose:** `settings.playerprovider` — still used by the add player group dialog, even though its siblings are gone.

- n/a

- n/a

## Types of changes

## Checklist

- [x] `pnpm build` passes.

### DIFF
--- a/src/translations/en.json
+++ b/src/translations/en.json
@@ -113,11 +113,6 @@
         "enabled": "Enable",
         "invalid_input": "The input is invalid",
         "metadataproviders": "Metadata providers",
-        "musicproviders": "Music providers",
-        "need_help_setup_provider": "Need help setting up this provider ?",
-        "no_providers": "No Music Providers configured",
-        "no_providers_detail": "Start your Music Assistant experience by adding your Music providers below.",
-        "not_loaded": "The provider is not (yet) loaded",
         "player_address": "Address",
         "player_disabled": "This player is disabled",
         "player_id": "Player ID",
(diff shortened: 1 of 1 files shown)


marcelveldt:
Holding off on merging: `Analyze (actions)` and `Analyze (python)` are red, but not because of this PR.

Both jobs finish their analysis fine and fail on the "Uploading results" step; `Analyze (javascript-typescript)` passes. The same two jobs were green on the earlier heads of this PR with the identical diff, and the CodeQL run on `main` itself failed at the same time (run 34749290214 on b14749e9), while every earlier run on main was green. GitHub's credential service was also handing out 503s in that window.

This PR deletes 17 lines from `en.json` and nothing else, so it can't affect the actions or python extractors. Re-running the failed jobs returns 403 ("cannot be retried"), so there's nothing to do here but wait for the upload service to recover — it should clear on the next run once main is green again.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2757, opened 2026-09-14
Resolve a music source's name through one helper

marcelveldt:

The provider detail page worked out its title with its own copy of the name
fallback chain, rather than the shared `providerDisplayName` helper the settings
list, the access dialog and the onboarding wizard use. The two copies disagreed
about what wins: the custom name you set, or the name the running instance
reports.

They now agree on the custom name. That is what the server itself prefers when
it names an instance, so nothing looks different — the detail page already
worked this way, and this keeps it that way while bringing the other three in
line. Doing it the other way round would have cost the detail page its instant
update when you rename a source.

- Point the provider detail page at the shared helper, dropping its private copy
- Let a custom name win over the running instance's name in the helper too, so
  all four callers agree, and match what `getPlayerName` already does for players
- Update the tests that pinned the old order

No user-visible change, hence `maintenance`.

- related issue `n/a`

- related backend PR `n/a`

## Types of changes

## Checklist

- [x] `pnpm build` passes.


marcelveldt:
On the two points Copilot raised on the first commit (the optimistic rename in `saveRename`, and the breadcrumb in `Settings.vue` not reading the config): both are real, but neither is caused by this PR. The detail page already resolved its title config-first before this change, so the rename rollback gap and the breadcrumb mismatch were both there already — this PR just keeps that behaviour while bringing the other three callers in line.

Leaving them out to keep this one a straight dedupe. Raising them separately.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2777, opened 2026-09-15
Show Party, AI Radio and the visualizer to guests

marcelveldt:

A signed-in guest never saw Party, AI Radio or the visualizer in the sidebar. The app checked each plugin through its provider settings, which a guest may not read, so every app start sent four refused commands and dropped all four entries. Since guests may play AI Radio stations (https://github.com/music-assistant/server/pull/6301), that hid a feature they are allowed to use.

- take the enabled plugins from the loaded providers, which every role may list
- a plugin that is still loading or failed to load gets no entry, since its page can't work until it has loaded
- guests now get Party, AI Radio and the visualizer; hosting a Music Quiz still takes the invite permission
- four fewer requests on every app start
- tests for a guest, for unavailable plugins and for the plugin list the app starts with

- n/a

- none: uses the provider list the server already sends to every role

## Types of changes

## Checklist

- [x] `pnpm build` passes.


marcelveldt:
Re Copilot's notes: the `else` branch is for party and quiz join sessions, see `isGuestAccessSession()`. Those stay on /guest and never show the sidebar. A signed-in account with the guest role goes through the regular branch and gets `refreshEnabledPlugins()`, which is what the new guest test covers, and it passes.

Skipping screenshots, since guests only get sidebar entries members already have and nothing new is drawn.

